PR #93, opened 2021-03-05
fix webgl shader compiler warnings 

toloudis:
The warnings were of the form "output of vertex shader not read by fragment shader"

Vertex shaders were declaring a full set of output variables and the fragment shaders were only using the ones they needed.
This change uses "ifdefs" to selectively output only the needed data from each vertex shader instance.

There are also a few whitespace changes, getting rid of some tabs and extra spaces.


blairlyons (approved):

schoinh (approved):
Nice, thanks!

PR #96, opened 2021-03-25
allow for a file name to be passed via url query string

toloudis:
This allows us to test known files on the server without having to add them to the dropdown list.
This is just for the "npm start" viewer testbed app.
You will be able to load localhost:8080/public/index.html?file=myfavoritetesttrajectory.simularium and the file will be set up as first in the dropdown list ready to be loaded.


schoinh (approved):
Nice!

meganrm (approved):

PR #97, opened 2021-03-31
Feature/convert TrajectoryFileInfo to latest version

schoinh:
Resolves: [Convert TrajectoryInfoData to latest version in simularium-viewer](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1372)

Equivalent conversion in simulariumio: https://github.com/allen-cell-animated/simulariumio/blob/master/simulariumio/file_converter.py

I ended up calling the conversion function in viewport/index.ts instead of NetConnection.ts because of 2 reasons:
1. I wasn't able to resolve some type errors, detailed [here](https://allencellscience.slack.com/archives/CMQFRJPPF/p1617220984102800?thread_ts=1616714552.041800&cid=CMQFRJPPF).
2. If we want to prettify the v1 scale bar label (otherwise the scale bar will display in meters for v1 trajectories which will impact currently posted models), I need the original TrajectoryFileInfo intact in viewport/index.ts so that I can use it along with the `tickIntervalLength` that visGeometry gives me in viewport. (See the TODO comment I put in viewport/index.ts for further explanation)

**Pull request recommendations:**

- [x] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [x] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_
- [x] Provide description and context of changes.
- [x] Provide relevant tests for your feature or bug fix.
- [ ] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

### DIFF
--- a/src/simularium/types.ts
+++ b/src/simularium/types.ts
@@ -49,6 +49,8 @@ export interface EncodedTypeMapping {
 }
 
 interface TrajectoryFileInfoGeneric {
+    connId: string;
+    msgType: number;
     version: number;
     timeStepSize: number;
     totalSteps: number;
--- a/src/simularium/types.ts
+++ b/src/simularium/types.ts
@@ -48,7 +48,7 @@ export interface EncodedTypeMapping {
     [key: number]: AgentDisplayData;
 }
 
-interface TrajectoryFileInfoGeneric {
+interface TrajectoryFileInfoBase {
     connId: string;
     msgType: number;
     version: number;
@@ -62,11 +62,11 @@ interface TrajectoryFileInfoGeneric {
     typeMapping: EncodedTypeMapping;
 }
 
-export interface TrajectoryFileInfoV1 extends TrajectoryFileInfoGeneric {
+export interface TrajectoryFileInfoV1 extends TrajectoryFileInfoBase {
     spatialUnitFactorMeters: number;
 }
 
-export interface TrajectoryFileInfoV2 extends TrajectoryFileInfoGeneric {
+export interface TrajectoryFileInfoV2 extends TrajectoryFileInfoBase {
     spatialUnits: {
         magnitude: number;
         name: string;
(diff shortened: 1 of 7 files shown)


src/test/versionHandlers.test.ts
```diff
@@ -0,0 +1,96 @@
+import { updateTrajectoryFileInfoFormat } from "../simularium/versionHandlers";
+
+const invalidVersionData = {
+    connId: "7496831076a233f0-2c337fed-4493-ad92-79f194744174ba05635426fd",
```
meganrm, line 4: couldn't this just have `version: 999.9` and no other data? seems like it would be a clearer test case that way,

meganrm (approved):

src/test/versionHandlers.test.ts
EricJIsaac: I may be mis-remembering my TS, but wouldn't any type-checks related to the `updateTrajectoryFileInfoFormat` function called fail without the other data?

EricJIsaac (approved):

src/simularium/types.ts
```diff
@@ -49,6 +49,8 @@ export interface EncodedTypeMapping {
 }
 
 interface TrajectoryFileInfoGeneric {
```
toloudis, line 51: What does "generic" mean here?

schoinh: Oh it's probably not an appropriate use of that word... What I meant is more like `TrajectoryFileInfoBase` or `TrajectoryFileInfoRoot` or `TrajectoryFileInfoCommon`? It just contains all the stuff that v1 and v2 versions have in common. Do you have a preference or suggestions?

src/test/versionHandlers.test.ts
schoinh: Yes @EricJIsaac is correct -- the test works with just the version number, but TS needs everything else 😛

PR #98, opened 2021-04-08
bugfix/yellow-agents

EricJIsaac:
**Pull request recommendations:**

- [x] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_
- [x] Provide description and context of changes.
- [ ] Provide relevant tests for your feature or bug fix.
- [ ] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

Check for a color mapping before rendering, return early from `updateScene(...)` if none has arrived yet. With the arraybuffer changes, the data for the first frame was processing and arriving before the trajectory file info.

We will need to increment the back-end version to completely fix this in staging and production. They array-buffer changes removed an unused net message type. The old versions of the back-end will be sending an outdated message type for trajectory file update (type 13 vs type 12).

After these changes, no more (unintended) yellow agents.

### DIFF
--- a/src/simularium/VisGeometry.ts
+++ b/src/simularium/VisGeometry.ts
@@ -8,6 +8,7 @@ import { USE_INSTANCE_ENDCAPS } from "./VisTypes";
 import PDBModel from "./PDBModel";
 import TaskQueue from "./worker/TaskQueue";
 import { REASON_CANCELLED } from "./worker/TaskQueue";
+import FrontEndError from "./FrontEndError";
 
 import {
     Box3,
@@ -148,6 +149,7 @@ class VisGeometry {
     public agentPathGroup: Group;
     public instancedMeshGroup: Group;
     public idColorMapping: Map<number, number>;
+    private isIdColorMappingSet: boolean;
     private raycaster: Raycaster;
     private supportsMoleculeRendering: boolean;
     private membraneAgent?: VisAgent;
@@ -173,6 +175,7 @@ class VisGeometry {
         this.pdbLoadAttempted = new Map<string, boolean>();
         this.scaleMapping = new Map<number, number>();
         this.idColorMapping = new Map<number, number>();
+        this.isIdColorMappingSet = false;
         this.followObjectId = NO_AGENT;
         this.visAgents = [];
         this.visAgentInstances = new Map<number, VisAgent>();
@@ -934,6 +937,7 @@ class VisGeometry {
 
     public clearColorMapping(): void {
         this.idColorMapping.clear();
+        this.isIdColorMappingSet = false;
     }
 
     private getColorIndexForTypeId(typeId: number): number {
@@ -951,6 +955,12 @@ class VisGeometry {
     }
 
     public setColorForIds(ids: number[], colorId: number): void {
+        if (this.isIdColorMappingSet) {
+            throw new FrontEndError(
+                "Attempted to set agent-color after color mapping was finalized"
+            );
+        }
+
         ids.forEach((id) => {
             this.idColorMapping.set(id, colorId);
         });
@@ -962,6 +972,10 @@ class VisGeometry {
             this.colorsData[index * 4 + 1],
             this.colorsData[index * 4 + 2]
         );
+    }
+
+    public finalizeIdColorMapping(): void {
+        this.isIdColorMappingSet = true;
     }
 
     /**
@@ -1338,6 +1352,10 @@ class VisGeometry {
      *   Update Scene
      **/
     public updateScene(agents: AgentData[]): void {
+        if (!this.isIdColorMappingSet) {
+            return;
+        }
+
         this.currentSceneAgents = agents;
 
         let dx = 0,
--- a/src/viewport/index.tsx
+++ b/src/viewport/index.tsx
@@ -245,6 +245,7 @@ class Viewport extends React.Component<ViewportProps, ViewportState> {
                         .getHexString();
                 colorIndex = colorIndex + 1;
             });
+	    this.visGeometry.finalizeIdColorMapping();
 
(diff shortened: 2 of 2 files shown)


EricJIsaac:
This also changes the `setColorForTypeId` function; it will throw an error if called after `finalizeIdColorMapping()`. Given the color issues we've encountered, I think it is a good idea to be stricter w/ this interface (add an explicit `updateColorForId(...)` if/when needed).

src/viewport/index.tsx
```diff
@@ -245,6 +245,7 @@ class Viewport extends React.Component<ViewportProps, ViewportState> {
                         .getHexString();
                 colorIndex = colorIndex + 1;
             });
+	    this.visGeometry.finalizeIdColorMapping();
```
toloudis, line 248: this looks like it's got the wrong amount of indent

toloudis (approved):

toloudis:
oh, one question:  if updateScene does the early exit, does the color mapping trigger another updateScene call?

EricJIsaac:
Currently updateScene is attempted whenever there is agent-data to draw. updateScene will be triggered on the next animation frame (and continually return early) until the color mapping is set.

src/viewport/index.tsx
EricJIsaac: It looks a bit funky but is in line w/ line 209 above (and passes the linter). The preceeding ~30 lines are a callback definition

toloudis: so is line 250 wrongly indented?

toloudis:
> Currently updateScene is attempted whenever there is agent-data to draw. updateScene will be triggered on the next animation frame (and continually return early until the color mapping is set).
> 
> Edit: just double checked, a new animation frame is requested in the Viewport's `animate()` function regardless of if updateScene succeeds.

Related to this, my confusion is that I thought we only call updateScene if we need to FORCE a refresh of all the agents, or if there is really new agent data.  The normal animation loop doesn't call updateScene over and over unless it has new agent data.  So I thought :)

EricJIsaac:
> 
> 
> > Currently updateScene is attempted whenever there is agent-data to draw. updateScene will be triggered on the next animation frame (and continually return early until the color mapping is set).
> > Edit: just double checked, a new animation frame is requested in the Viewport's `animate()` function regardless of if updateScene succeeds.
> 
> Related to this, my confusion is that I thought we only call updateScene if we need to FORCE a refresh of all the agents, or if there is really new agent data. The normal animation loop doesn't call updateScene over and over unless it has new agent data. So I thought :)

You are right on this, my understanding was outdated. I'm actually not sure how this is working currently (but can observe that it does)... the viewer still updates correctly even if initial update attempts are refused.

schoinh (approved):

schoinh:
@EricJIsaac Can/should we merge this? 😛

src/simularium/VisGeometry.ts
```diff
@@ -951,6 +955,12 @@ class VisGeometry {
     }
 
     public setColorForIds(ids: number[], colorId: number): void {
+        if (this.isIdColorMappingSet) {
+            throw new FrontEndError(
+                "Attempted to set agent-color after color mapping was finalized"
+            );
```
meganrm, line 961: Is this being thrown as often as we were seeing all yellow agents before? This doesn't seem like a helpful/actionable message to the front end user. If I understand this correctly, I think this should just silently return out.

toloudis:
I wonder if things would be cleaner if we just forced the client to wait for the TrajectoryFileInfo before it is allowed to receive the first visdata.  It seems like it should be "ok" from a design/architecture standpoint to just say that TrajectoryFileInfo is required because it provides critical bootstrapping info.

src/simularium/VisGeometry.ts
EricJIsaac: No, this is thrown if a dev calls finalize, then attempts to set the color. There shouldn't be a way for a user to cause this through the simularium-ui, this error was to let us know quickly if we use the interface in a way that isn't intended.

On second look, it would make more sense to print this out to the console than throw a FE error. I'll update the PR.

EricJIsaac: This function is currently only called when the trajectory file info arrives (and the render loop references the results)

EricJIsaac:
> I wonder if things would be cleaner if we just forced the client to wait for the TrajectoryFileInfo before it is allowed to receive the first visdata. It seems like it should be "ok" from a design/architecture standpoint to just say that TrajectoryFileInfo is required because it provides critical bootstrapping info.

It seems reasonable to me; in a way, that is what the current PR attempts to do indirectly

EricJIsaac:
> @EricJIsaac Can/should we merge this?

I don't think it is a good idea to merge this until the deployment issues are resolved. It will add another layer of confusion (since these changes could conceivably prevent rendering). It would be difficult to distinguish between a deployment issue and an issue due to this update.

Currently dev is inaccessible (no web-socket traffic reaching), production is down (w/ the only change being a cleared cache), and staging is encountering an error. Still debugging.

EricJIsaac:
> > @EricJIsaac Can/should we merge this?
> 
> I don't think it is a good idea to merge this until the deployment issues are resolved. It will add another layer of confusion (since these changes could conceivably prevent rendering). It would be difficult to distinguish between a deployment issue and an issue due to this update.
> 
> Currently dev is inaccessible (no web-socket traffic reaching), production is down (w/ the only change being a cleared cache), and staging is encountering an error. Still debugging.

Major issues are resolved in the staging site; we'll need to update production to at least 2.7.0 (so the message types are in sync again), but afterwards the production site should be back to functional.

PR #99, opened 2021-04-09
Feature/faster fibers

toloudis:
Optimized fiber rendering.
Previously, everything that was a fiber was constructing a TubeGeometry from a 3d curve, and adding sphere end caps.  Since the curve points can(and are expected to) change every frame, every curve is reconstructed.  This was being done entirely on the cpu. 

These changes move all the curve reconstruction work onto the GPU.  The only data sent to the gpu is the curve control points, and a set of vertices to form the tube triangles.  The geometry is still in the form of a tube.  Endcaps are built in to the gpu representation, so the fiber ends are closed properly.

For fallback webgl1 rendering, we still create TubeGeometries on the CPU in the former slow way.  However, I have removed all the end cap geometry so the tubes have no ends.

In order to render most efficiently, there is a separate fiber setup for each possible length of fiber (in terms of number of curve control points).  The InstancedFiberGroup contains one InstancedFiber per number of control points.  (For example, if there are 7 fibers of length 3, and 6 fibers of length 11, the InstancedFiberGroup will draw two instanced objects. The first object will be a tube for a curve of 3 control points, and will be instanced 7 times.  The second object will be a tube for a curve of 11 control points, and will be instanced 6 times. )   The per-fiber curve data is put into a texture for each of the fiber objects, and read out in the shader in order to position the vertices in 3d.

This might be sufficiently complicated to discuss in one of our dev meetings.


toloudis:
The core implementation for the curve rendering is in InstancedFiberShader.ts, mostly in the vertexShader at the beginning of that file.

src/simularium/VisGeometry.ts
```diff
@@ -312,27 +320,28 @@ class VisGeometry {
...
+        this.fibers.clear();
+        removeByName(this.instancedMeshGroup, InstancedFiberGroup.GROUP_NAME);
+
         // tell instanced geometry what representation to use.
         if (this.renderStyle === RenderStyle.GENERIC) {
-            this.fiberEndcaps = new InstancedFiberEndcapsFallback();
         } else {
-            this.fiberEndcaps = new InstancedFiberEndcaps();
+            this.fibers = new InstancedFiberGroup();
         }
```
schoinh, line 342: Looks like nothing happens after checking `if (this.renderStyle === RenderStyle.GENERIC)` anymore.

src/simularium/VisAgent.ts
```diff
@@ -414,42 +403,22 @@ export default class VisAgent {
...
-            runtimeFiberEncapMesh1.position.y =
-                curvePoints[curvePoints.length - 1].y;
-            runtimeFiberEncapMesh1.position.z =
-                curvePoints[curvePoints.length - 1].z;
-            runtimeFiberEncapMesh1.scale.x = collisionRadius * scale * 0.5;
-            runtimeFiberEncapMesh1.scale.y = collisionRadius * scale * 0.5;
-            runtimeFiberEncapMesh1.scale.z = collisionRadius * scale * 0.5;
+        if (regenerateMesh) {
+            // expensive
+            const fibergeometry = new TubeBufferGeometry(
```
schoinh, line 409: [nit] `fiberGeometry` (camelcase)?

src/simularium/VisAgent.ts
```diff
@@ -414,42 +403,22 @@ export default class VisAgent {
...
-            runtimeFiberEncapMesh1.position.x =
-                curvePoints[curvePoints.length - 1].x;
-            runtimeFiberEncapMesh1.position.y =
-                curvePoints[curvePoints.length - 1].y;
-            runtimeFiberEncapMesh1.position.z =
-                curvePoints[curvePoints.length - 1].z;
-            runtimeFiberEncapMesh1.scale.x = collisionRadius * scale * 0.5;
-            runtimeFiberEncapMesh1.scale.y = collisionRadius * scale * 0.5;
-            runtimeFiberEncapMesh1.scale.z = collisionRadius * scale * 0.5;
+        if (regenerateMesh) {
```
schoinh, line 407: I'm a little confused on where the fallback rendering vs non-fallback rendering are happening. This is where the fallback creation of TubeGeometries the slow way is being triggered right? I keep getting `GENERIC` vs `MOLECULAR` mixed up... I think some code comments about those would be helpful like @blairlyons said :)

toloudis: I changed the names GENERIC-->WEBGL1_FALLBACK and MOLECULAR-->WEBGL2_PREFERRED

schoinh (approved):

meganrm (approved):

PR #101, opened 2021-04-20
bugfix/correct-binary-pre-parse-caching

EricJIsaac:
Problem
=======
Binary broadcasts were being parsed incorrectly. This is an urgently needed fix to get the deployed BE, FE, and data back in sync.

Solution
========
Copy the previous contents of the 'pre-parse' binary cache when appending new cache data.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/simularium/VisData.ts
+++ b/src/simularium/VisData.ts
@@ -548,8 +548,7 @@ class VisData {
                 this.netBuffer.byteLength
             );
 
-            this.netBuffer = new ArrayBuffer(tmp.byteLength);
-            new Uint8Array(this.netBuffer).set(new Uint8Array(tmp));
+            this.netBuffer = tmp;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/simularium/VisData.ts
```diff
@@ -538,8 +538,18 @@ class VisData {
...
+            const tmp = new ArrayBuffer(
+                this.netBuffer.byteLength + frame.byteLength
+            );
+            new Uint8Array(tmp).set(new Uint8Array(this.netBuffer));
+            new Uint8Array(tmp).set(
+                new Uint8Array(frame),
+                this.netBuffer.byteLength
+            );
+
+            this.netBuffer = new ArrayBuffer(tmp.byteLength);
```
toloudis, line 551: could just set this to tmp now:  `this.netBuffer = tmp;` ?

toloudis (approved):

schoinh (approved):

PR #102, opened 2021-04-22
Feature/procedural js simulation

toloudis:
Problem
=======
1. Needed a way to stress test the rendering code with arbitrary numbers of agents, quickly and easily
2. Local file handling was mixed in with network communication to back end
3. NetConnection is only one mode available to run simulations.  Drag and drop files, and javascript cpu simulation are other types of "connection" that the controller could be handling polymorphically.

Solution
========
Controller now has an instance of the ISimulator interface, which can be a NetConnection, a LocalFileConnection, or a ClientSimulatorConnection.   ClientSimulator is further split to support multiple different types of client simulator classes. See PointSimulator and CurveSimulator, both of which are used by ClientSimuatorConnection.

Drag n drop now does not cache the whole file up front.  It runs the trajectory frame by frame and caches as it goes.

There are several decisions, proposals, and changes still to be made to improve this code, most of which are commented as TODOs in this change set.

Key breaking api change:  the Controller's changeFile function now takes different arguments. This will require a small change to simularium-website.

I have not been able to determine that simularium-website still works the same because of ongoing back end deployment issues.  This will be tested further once that stabilizes.

## Type of change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This change requires updated or new tests

Steps to Verify:
----------------
For now, can run the Viewer example app from this repo.  It supports drag n drop, loading files from server, and the hardcoded custom simulators.

-----------------------
1. Controller/index.ts
2. simularium/ISimulator.ts
3. simularium/ClientSimulatorConnection.ts
4. simularium/LocalFileConnection.ts
5. simularium/NetConnection.ts

Thanks for contributing!


schoinh:
> Drag n drop now does not cache the whole file up front. It runs the trajectory frame by frame and caches as it goes.

Does this make the drag and drop playback any slower? I think an advantage of drag-and-drop files vs. networked files so far has been that there is far less lag on the drag-and-drop files once loaded, for example when scrubbing on the slider or jumping ahead to a point in time.

toloudis:
> > Drag n drop now does not cache the whole file up front. It runs the trajectory frame by frame and caches as it goes.
> 
> Does this make the drag and drop playback any slower? I think an advantage of drag-and-drop files vs. networked files so far has been that there is far less lag on the drag-and-drop files once loaded, for example when scrubbing on the slider or jumping ahead to a point in time.

Your observation is correct:  it makes drag and drop play slower the first time through, but for large files you can start playing sooner because it's not converting up-front.   I think it would be fairly easy to change the code to have it convert/cache the entire file if that's too much of a behavior change.  After the first playthrough everything is as smooth as before.

toloudis:
Pre-Review notes from the TODOs, for our in-person meeting: 
* SimulariumControllerParams needs to be refined
* SimulariumConnectionParams needs to be refined
* SimulariumController.changeFile api needs to be refined/renamed
* ISimulator interface
* NetMessage vs actual message payload data, factor out the connectionid and msgType;  see line 167 of NetConnection vs line 10 and 54 of types.ts
* Can we force sync of the TrajectoryFileInfo so we guarantee it arrives before the first frame data?

src/controller/index.ts
```diff
@@ -131,32 +137,48 @@ export default class SimulariumController {
...
+            this.netConnection = new NetConnection(netConnectionConfig);
+        } else {
+            throw new Error(
+                "Insufficient data to determine and configure simulator connection"
+            );
+        }
+
+        this.netConnection.setTrajectoryFileInfoHandler(
+            (trajFileInfo: TrajectoryFileInfo) => {
+                this.handleTrajectoryInfo(trajFileInfo);
```
schoinh, line 162: Able to update TrajectoryFileInfo to latest version here (and just do it once here)??

src/simularium/ISimulator.ts
```diff
@@ -0,0 +1,68 @@
+//import jsLogger from "js-logger";
+//import { ILogger } from "js-logger";
+
+import { VisDataMessage, TrajectoryFileInfoV2 } from "./types";
+
+// TODO we need to examine this whole interface and document it with comments here
+// what is each function expected to do, from the caller's perspective?
+// are they named appropriately?
+// and are some of them completely specific to websockets only?
+export interface ISimulator {
```
schoinh, line 10: I think this interface should have some sort of a doc block

src/simularium/ClientSimulatorConnection.ts
```diff
@@ -0,0 +1,289 @@
...
+import {
+    createSimulator,
+    ClientSimulatorParams,
+} from "./localSimulators/ClientSimulatorFactory";
+import { ISimulator } from "./ISimulator";
+
+// setInterval is the playback engine for now
+let simulatorIntervalId = 0;
+
+export class SimulatorConnection implements ISimulator {
```
schoinh, line 19: Bit confusing that the file name doesn't match this class name

src/simularium/localSimulators/IClientSimulator.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// TODO these should not be needed anywhere except ClientSimulatorConnection
+export const enum ClientPlayBackType {
+    ID_LIVE_SIMULATION = 0,
+    ID_PRE_RUN_SIMULATION = 1,
+    ID_TRAJECTORY_FILE_PLAYBACK = 2,
+    // insert new values here before LENGTH
+    LENGTH,
+}
```
schoinh, line 36: Is it possible to avoid having these enums in 2 places (in this file and in NetConnection.ts)?

src/controller/index.ts
```diff
@@ -131,49 +137,67 @@ export default class SimulariumController {
...
+    ): void {
+        if (clientSimulatorParams) {
+            this.simulator = new SimulatorConnection(clientSimulatorParams);
+        } else if (localFile) {
+            this.simulator = new LocalFileConnection(
+                this.playBackFile,
+                localFile
+            );
+        } else if (netConnectionConfig) {
+            this.simulator = new NetConnection(netConnectionConfig);
```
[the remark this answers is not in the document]
toloudis: I agree about using Simulator instead of Connection.  I'll make this change.  And how about RemoteSimulator instead of NetworkSimulator?

src/simularium/VisGeometry.ts
```diff
@@ -86,12 +86,6 @@ interface MeshLoadRequest {
     cancelled: boolean;
 }
 
-interface HSL {
-    h: number;
-    s: number;
-    l: number;
-}
```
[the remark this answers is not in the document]
toloudis: hue, saturation, luminance.  it's a way to define a color.

src/simularium/NetConnection.ts
```diff
@@ -75,6 +88,17 @@ export class NetConnection {
         window.addEventListener("beforeunload", this.onClose.bind(this));
     }
 
+    public setTrajectoryFileInfoHandler(
+        handler: (msg: TrajectoryFileInfoV2) => void
```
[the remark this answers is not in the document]
toloudis: agreed

src/simularium/LocalFileConnecton.ts
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+    }
+
+    public requestTrajectoryFileInfo(_fileName: string): void {
+        this.onTrajectoryFileInfoArrive(this.simulariumFile.trajectoryInfo);
+    }
+
+    private getFrame(theFrameNumber: number): VisDataMessage {
+        //        return this.getAllFrames();
```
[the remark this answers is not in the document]
toloudis: I left that in to show how we could pre-cache the whole trajectory in this way. I'll remove it or change it to a more verbose comment about that.

src/simularium/localSimulators/IClientSimulator.ts
toloudis: the only one that matters is the one in netconnection.  the ones in here are totally artificial.  I'll think about removing them but they don't affect the ISimulator / SimulariumController operations.

src/controller/index.ts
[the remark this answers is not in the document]
schoinh: That sounds good to me!

src/simularium/LocalFileConnecton.ts
```diff
@@ -0,0 +1,194 @@
...
+        for (
+            let frame = 0,
+                numFrames = this.simulariumFile.spatialData.bundleData.length;
+            frame < numFrames;
+            frame++
+        ) {
+            const frameTime = this.simulariumFile.spatialData.bundleData[frame]
+                .time;
+            if (timeNanoSeconds < frameTime) {
+                const theFrameNumber = Math.max(frame - 1, 0);
```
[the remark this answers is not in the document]
toloudis: Yes it was copied from VisData.  In the case of LocalFileSimulator, we have to be able to access a time before it was cached in VisData, because I no longer guarantee that every frame is precached.  I have updated this code to copy the algorithm from VisData now

src/controller/index.ts
```diff
@@ -1,7 +1,7 @@
 import jsLogger from "js-logger";
 import { noop } from "lodash";
 import {
-    NetConnection,
+    RemoteSimulator,
```
EricJIsaac, line 4: This is a much better name for it; it also seems like this may be a good way to split the interface (remote vs. frontend simulators) in future refactors

src/simularium/RemoteSimulator.ts
```diff
@@ -376,6 +402,9 @@ export class NetConnection {
             "file-name": fileName,
         };
 
+        //  TODO is this comment correct?
+        // begins a stream which will include a TrajectoryFileInfo and a series of VisDataMessages
```
EricJIsaac, line 406: This comment is correct

src/viewport/index.tsx
```diff
@@ -249,12 +251,13 @@ class Viewport extends React.Component<ViewportProps, ViewportState> {
...
                 colorIndex = colorIndex + 1;
             });
-	    this.visGeometry.finalizeIdColorMapping();
+            this.visGeometry.finalizeIdColorMapping();
 
             onUIDisplayDataChanged(uiDisplayData);
         };
 
         simulariumController.postConnect = () => {
+            console.log("POST CONNECT CALLBACK");
```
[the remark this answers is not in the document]
toloudis: will remove

EricJIsaac (approved):

src/simularium/ISimulator.ts
toloudis: I want to clean up this whole interface in another PR

src/simularium/ClientSimulator.ts
```diff
@@ -0,0 +1,291 @@
...
+    ClientSimulatorParams,
+} from "./localSimulators/ClientSimulatorFactory";
+import { ISimulator } from "./ISimulator";
+
+// setInterval is the playback engine for now
+let simulatorIntervalId = 0;
+
+// a ClientSimulator is a ISimulator that is expected to run purely in procedural javascript in the browser client,
+// with the procedural implementation in a IClientSimulatorImpl
+export class ClientSimulator implements ISimulator {
```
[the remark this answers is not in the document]
meganrm: I kinda like `ProceduralSimulator` 🤷🏻‍♀️

src/simularium/ISimulator.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    pauseRemoteSim(): void;
+
+    resumeRemoteSim(): void;
+
+    abortRemoteSim(): void;
+
+    requestSingleFrame(startFrameNumber: number): void;
+
+    gotoRemoteSimulationTime(timeNanoSeconds: number): void;
```
meganrm, line 58: I still don't think i understand why you are using the functions here that are "remoteSim" instead of the names that were on the controller before. Ie in this case `SimulariumController.gotoTime`

toloudis: However, I do want to refine this interface but was hoping to do it in a subsequent PR.

schoinh (approved):

src/simularium/RemoteSimulator.ts
```diff
@@ -376,6 +402,9 @@ export class NetConnection {
...
+        // Note that it is possible for the first vis data to arrive before the TrajectoryFileInfo...
```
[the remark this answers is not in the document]
toloudis: A quick google search seems to say yes but there can be application software layers that reorder things. From what I read, we can assume that if trajectoryfileinfo is sent before the first data frame, then it should be received first too.

PR #104, opened 2021-04-28
Fix/ convert spatialUnits correctly in updateTrajectoryFileInfoFormat

schoinh:
Problem
=======
Scale bar labels for v1 trajectories look like this:
image

Solution
========
I was completely overthinking / mistaken about this before. Sorry about all the confusion during the last 2 meetings. Like @blairlyons was saying all along it's a really simple fix that just requires me to properly convert `spatialUnitFactorMeters` to `spatialUnits` during the conversion in `updateTrajectoryFileInfoFormat()`.

## Type of change

Steps to Verify:
----------------
Install this branch as a dependency for your local simularium-website and load some v1 trajectory files
1. in viewer, run `npm run build`
2. in website, delete the viewer from node-modules and put this in package.json `"@aics/simularium-viewer": "file:../simularium-viewer",`
3. Add `path.resolve(__dirname, "../", "../simularium-viewer/style")` as one of the CSS rules in webpack.config.js in website
4. in website, run `npm i` then `npm start`

-----------------------
1. src/simularium/versionHandlers.ts - the fix
2. src/viewport/index.tsx - renamed `newMsg` as `trajectoryFileInfo` per @meganrm 's suggestion

Thanks for contributing!


src/simularium/versionHandlers.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable-next-line @typescript-eslint/no-var-requires */
+const si = require("si-prefix");
```
toloudis, line 2: minor red flag when you have to do this;  the library is old and not maintained.  Makes me want to look for another one, like unitsnet-js for example.  Alternative: write a small module wrapper file that "hides" the require statement.

toloudis (approved):
Looks good! Does the unit test modification represent properly handling the "bad" behavior?

EricJIsaac (approved):

schoinh:
> Looks good! Does the unit test modification represent properly handling the "bad" behavior?

Yeah, it does! In simularium-website we're multiplying `tickLengthInterval` by `spatialUnits.magnitude`, which is usually like 1e-6 currently (as you can see in the test). So that creates a really small number for the scale bar label. We might still have to round the number in simularium-website, but that's no big deal

src/simularium/versionHandlers.ts
schoinh: Yeah, it is an old library but it appealed to me because it's small and does exactly what we need for this specific purpose. But I agree we should look for a better one. We need to add a more comprehensive unit library for unit validation (like we talked about before) at some point, so maybe that'll be a good time to research and upgrade our unit library. I just looked into unitsnet-js but couldn't find in their docs how to use it to take 1.5e-9 m for example and turn it into 1.5 nm, but I might just need to look more carefully.

I did some Googling and found that I can actually just use `import * as si from "si-prefix";`, so changed it to that!

PR #111, opened 2021-05-14
Feature/implement user-set camera transform

schoinh:
Problem
=======
The viewer doesn't do anything yet with the new `cameraDefault` data (camera position and orientation) in v2 trajectory files.

Resolves: [implement set initial camera transform from data](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1082)

Solution
========
Enable the viewer to use the `cameraDefault` data to set the initial camera orientation and position per user's specifications

## Type of change

- [x] New feature (non-breaking change which adds functionality)

Key files / Change summary:
---------------
* src/simularium/types.ts - Add `cameraDefault` to the `TrajectoryFileInfoV2` interface
* src/simularium/versionHandlers.ts - Add a `cameraDefault` block with default values to all v1 trajectories when converting to v2
* src/simularium/VisGeometry.ts
    * Created a `cameraDefault` property to hold the default camera transform values initially and to hold custom values if they are received
    * Created a `positionCamera()` method which does three.js operations to position and orient the camera
    * Invoke `positionCamera()` from `resetCamera()` so that any time a new file is loaded or when the user clicks the Reset Camera button, the camera's position and orientation is reset to what's in `this.cameraDefault` in addition to the camera's `OrbitControls` being reset
    * Minor refactoring
        * Removed `resetCameraOnNewScene` property because it should just always be `true` (per @toloudis 's suggestion)
        * Renamed `handleTrajectoryData()` method as `handleTrajectoryFileInfo()` to make it more accurate and less confusing

Steps to Verify:
----------------
1. Download [this test file](https://drive.google.com/file/d/1DJF8OYcFEIoh1KriNzqhKC5F2qAprsaN/view?usp=sharing). This file contains the following info in `trajectoryFileInfo`:
```
"cameraDefault":{
    "position":{
        "x":120,
        "y":0,
        "z":120
    },
    "lookAtPosition":{
        "x":0,
        "y":0,
        "z":0
    },
    "upVector":{
        "x":1,
        "y":2,
        "z":0
    },
    "fovDegrees": 90
}
```
2. Load it into our staging web site, notice you're looking at the bounding box straight-on (the camera params above don't have an effect)
3. Pull this branch, `npm start` and load it into the example viewer. Notice the bounding box is now farther away, rotated, and seen from a horizontally offset position

### Note
* The default fov angle [specified in simulariumio ](https://github.com/allen-cell-animated/simulariumio/blob/master/file_format.md) is 50 degrees. I talked with @blairlyons about this and it should be 75 degrees instead to match what's been the hard-coded default in the viewer. So I used 75 as the default. Blair will change it in simulariumio as well.
* If default values were used because the file was converted from v1 to v2 or because none were specified, then simularium-viewer emits a console warning about it. I don't think this will be visible to users on the simularium-website end though -- do we need it to be?

### DIFF
--- a/src/simularium/VisGeometry.ts
+++ b/src/simularium/VisGeometry.ts
@@ -56,8 +56,8 @@ const NO_AGENT = -1;
 
 const DEFAULT_CAMERA_POSITION: [number, number, number] = [0, 0, 120];
 const DEFAULT_CAMERA_LOOKAT: [number, number, number] = [0, 0, 0];
-const DEFAULT_CAMERA_UP: [number, number, number] = [0, 1, 0];
-const DEFAULT_CAMERA_FOV = 75;
+const DEFAULT_CAMERA_UP: [number, number, number] = [0, 1, 0]; // Must be a unit vector
+const DEFAULT_CAMERA_FOV = 75; // Degrees
 
 const CAMERA_DOLLY_STEP_SIZE = 10;
 export enum RenderStyle {
@@ -375,26 +375,13 @@ class VisGeometry {
             this.resetBounds(DEFAULT_VOLUME_DIMENSIONS);
         }
 
+        // Position and orient the camera
         this.resetCamera();
         this.positionCamera(trajectoryFileInfo.cameraDefault);
     }
 
     public positionCamera(cameraDefault: CameraTransform | undefined): void {
-        if (cameraDefault === undefined) {
-            this.logger.warn(
-                "Using default camera settings since none were provided"
-            );
-
-            this.camera.position.set(...DEFAULT_CAMERA_POSITION);
-            this.initCameraPosition = this.camera.position.clone();
-
-            this.camera.up.set(...DEFAULT_CAMERA_UP);
-
-            this.camera.lookAt(...DEFAULT_CAMERA_LOOKAT);
-            this.controls.target.set(...DEFAULT_CAMERA_LOOKAT);
-
-            this.camera.fov = DEFAULT_CAMERA_FOV;
-        } else {
+        if (cameraDefault) {
             const {
                 position,
                 upVector,
@@ -405,6 +392,7 @@ class VisGeometry {
             this.camera.position.set(position.x, position.y, position.z);
             this.initCameraPosition = this.camera.position.clone();
 
+            // Up vector needs to be a unit vector
             const normalizedUpVector = new Vector3(
                 upVector.x,
                 upVector.y,
@@ -428,8 +416,19 @@ class VisGeometry {
             );
 
             this.camera.fov = fovDegrees;
+        } else {
+            this.logger.warn(
+                "Using default camera settings since none were provided"
+            );
+            this.camera.position.set(...DEFAULT_CAMERA_POSITION);
+            this.initCameraPosition = this.camera.position.clone();
+            this.camera.up.set(...DEFAULT_CAMERA_UP);
+            this.camera.lookAt(...DEFAULT_CAMERA_LOOKAT);
+            this.controls.target.set(...DEFAULT_CAMERA_LOOKAT);
+            this.camera.fov = DEFAULT_CAMERA_FOV;
         }
 
+        // Apply the changes above
         this.camera.updateProjectionMatrix();
     }
 
--- a/src/simularium/VisGeometry.ts
+++ b/src/simularium/VisGeometry.ts
@@ -383,9 +383,12 @@ class VisGeometry {
         if (cameraDefault === undefined) {
             this.camera.position.set(...DEFAULT_CAMERA_POSITION);
             this.initCameraPosition = this.camera.position.clone();
+
             this.camera.up.set(...DEFAULT_CAMERA_UP);
+
             this.camera.lookAt(...DEFAULT_CAMERA_LOOKAT);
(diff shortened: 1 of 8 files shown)


blairlyons:
will a streaming file from the backend use the default values until Eric adds camera data to those?

schoinh:
> will a streaming file from the backend use the default values until Eric adds camera data to those?

Yup! Any file that doesn't have `cameraDefault` info will use the default values, and I checked that it was happening with streamed files

src/simularium/VisGeometry.ts
```diff
@@ -370,14 +388,67 @@ class VisGeometry {
...
             this.resetBounds(DEFAULT_VOLUME_DIMENSIONS);
         }
-        if (this.resetCameraOnNewScene) {
-            this.resetCamera();
+
+        // Get default camera transform values from data
+        if (trajectoryFileInfo.cameraDefault) {
+            this.cameraDefault = trajectoryFileInfo.cameraDefault;
+        } else {
+            this.logger.warn(
```
toloudis, line 396: this isn't really flipping back to the default, right?  It's just leaving the most recent camera settings?

schoinh: Ohhh good catch! I'll fix it

src/simularium/types.ts
```diff
@@ -15,6 +15,19 @@ export interface VisDataMessage {
     fileName: string;
 }
 
+interface Coordinates3d {
+    x: number;
+    y: number;
+    z: number;
+}
+
+export interface CameraTransform {
```
toloudis, line 24: I would call this something other than CameraTransform, because fov is part of a "projection transform" and the rest of the data is used for a "view transform".  I'd go for something like CameraSpec/CameraInfo/CameraData or even Camera(but that keyword tends to be already used in threejs etc)

src/simularium/localSimulators/PointSimulator.ts
```diff
@@ -128,6 +128,24 @@ export default class PointSim implements IClientSimulatorImpl {
...
+                    y: 0,
+                    z: 0,
+                },
+                upVector: {
+                    x: 0,
+                    y: 1,
+                    z: 0,
+                },
+                fovDegrees: 75,
+            },
```
toloudis, line 148: this z=120, fov=75 default camera could go in a constants file as DEFAULT_CAMERA_SPEC

src/simularium/VisGeometry.ts
toloudis: could set this.cameraDefault = DEFAULT_CAMERA_SPEC here, if that would help

src/simularium/types.ts
schoinh: Changed to `CameraSpec` :)

src/viewport/index.tsx
```diff
@@ -218,7 +218,7 @@ class Viewport extends React.Component<ViewportProps, ViewportState> {
             // Create a new bounding box and tick marks (via resetBounds()) and set
             // VisGeometry.tickIntervalLength, to make it available for use as the length of the
             // scale bar in the UI
-            this.visGeometry.handleTrajectoryData(trajectoryFileInfo);
+            this.visGeometry.handleTrajectoryFileInfo(trajectoryFileInfo);
```
toloudis, line 221: the comment above this could probably move into the actual function implementation

src/simularium/VisGeometry.ts
```diff
@@ -370,14 +371,68 @@ class VisGeometry {
...
+                "Using default camera settings since none were provided"
+            );
+            this.cameraDefault = DEFAULT_CAMERA_SPEC;
```
toloudis, line 382: small caution here: if we modify this.cameraDefault, then we run the risk of modifying DEFAULT_CAMERA_SPEC.  Could either clone or use some version of Object.freeze combined with readOnly to enforce it better?

toloudis (approved):

src/simularium/VisGeometry.ts
```diff
@@ -370,14 +371,68 @@ class VisGeometry {
...
     }
 
+    // Called when a new file is loaded, the Clear button is clicked, or Reset Camera button is clicked
     public resetCamera(): void {
         this.followObjectId = NO_AGENT;
         this.controls.reset();
+        this.positionCamera();
+    }
+
+    public positionCamera(): void {
```
meganrm, line 395: nit, but this is more `resetCameraPosition`, or `returnCameraToInitPosition`.  `positionCamera` sounds like I'm going to be able to give the function a position to set.

src/viewport/index.tsx
schoinh: Done

src/simularium/VisGeometry.ts
schoinh: 😱 Today I Learned JS objects are reference types..... I used lodash.deepClone to fix it. Thanks for catching it. Now I'm wondering how many lines of code I've written that's got this flaw in it has ever been merged 😬

src/simularium/VisGeometry.ts
schoinh: I see what you're saying. Renamed to `resetCameraPosition`

meganrm (approved):

PR #112, opened 2021-05-18
Feature/mesh instancing

toloudis:
Problem
=======
scenes with large numbers of spheres are too slow to render.
Also, the webGL 1 code path and the webgl2 code paths are too entangled in VisGeometry.  The VisGeometry is becoming more and more unreadable.

Solution
========
Optimize mesh rendering by using gpu instancing.
Simplify VisGeometry by refactoring big pieces of rendering code into a LegacyRenderer module.   Also simplify the VisAgent class by having it store an AgentData object of its own.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

Change summary:
---------------
Hopefully this does not break anything but it is a lot of new code.
Cpu javascript processing is becoming more and more the bottleneck for per-frame updates. 
Pdb is the last geometry type that is not using gpu instancing in the renderer yet.

Steps to Verify:
----------------
Just verify that highlighting, following, show/hide, and changing trajectory still works.
Scenes with high counts of meshes should render significantly faster.
A good test scene is `smoldyn_min1_output` or the point test simulator.

-----------------------
VisAgent, VisGeometry (in particular `updateScene()`), InstancedMesh
LegacyRenderer is entirely new but has code that hopefully will not be run very much and eventually deprecated
Thanks for contributing!


EricJIsaac:
Just to check - would switching to instances here affect our ability to occlude agents?
e.g. a scene where 1000 small agents of type 'A' surround 1 large agent of type 'B'; If 1 agent is in front of B, and 999 are behind B, then all 1000 would be drawn.

I remember this being a limitation of the game renderers I've seen (occlusion being all or nothing for groups of GPU instanced meshes). But I don't know if this is still a limitation and want to understand the trade-offs.

toloudis:
> Just to check - would switching to instances here affect our ability to occlude agents?
> e.g. a scene where 1000 small agents of type 'A' surround 1 large agent of type 'B'; If 1 agent is in front of B, and 999 are behind B, then all 1000 would be drawn.
> 
> I remember this being a limitation of the game renderers I've seen (occlusion being all or nothing for groups of GPU instanced meshes). But I don't know if this is still a limitation and want to understand the trade-offs.

Good question.  It depends on the occlusion method.  (Everything is still depth tested, for example).  If we were going to write some object-level occlusion system, then it would have to be computed prior to adding instances to these lists for drawing.  For gpu occlusion queries, I'm not sure how that would work -- that may be where the two methods are in conflict.  But I don't think webgl allows gpu occlusion queries anyway.

src/simularium/rendering/InstancedMesh.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    addInstance(
+        x: number,
+        y: number,
+        z: number,
+        scale: number,
+        rx: number,
+        ry: number,
+        rz: number,
+        instanceId: number,
```
toloudis, line 164: maybe rename this to agentId or uniqueId

src/simularium/rendering/LegacyRenderer.ts
```diff
@@ -0,0 +1,223 @@
...
+            transparent: true,
+            opacity: 0.6,
+        });
+        this.followMaterial = new MeshBasicMaterial({
+            color: new Color(FOLLOW_COLOR),
+        });
+        this.agentMeshGroup = new Group();
+    }
+
+    // public getGroup(): Group {
```
toloudis, line 51: remove this

src/simularium/rendering/LegacyRenderer.ts
```diff
@@ -0,0 +1,223 @@
...
+    //     return this.agentMeshGroup;
+    // }
+
+    public static membraneData: {
```
toloudis, line 55: remove this, membrane shader is dead for now

src/simularium/VisGeometry.ts
```diff
@@ -204,6 +208,7 @@ class VisGeometry {
 
         this.membraneAgent = undefined;
 
+        this.legacyRenderer = new LegacyRenderer();
         this.moleculeRenderer = new MoleculeRenderer();
```
toloudis, line 212: maybe rename this to somethign without molecule in the name

blairlyons (approved):
I'm not as familiar with this codebase, but the changes look ok. I tested in the example viewer with the Smoldyn Min1 trajectory, the first time through it was fairly slow, but second time was great. I'm assuming the limiting step is building the cache now? Also, I no longer see the accumulating agents bug with this change, I definitely still see it on master with the Smoldyn Min1 trajectory 🎉

toloudis:
> I'm not as familiar with this codebase, but the changes look ok. I tested in the example viewer with the Smoldyn Min1 trajectory, the first time through it was fairly slow, but second time was great. I'm assuming the limiting step is building the cache now? Also, I no longer see the accumulating agents bug with this change, I definitely still see it on master with the Smoldyn Min1 trajectory 🎉

If the second playthrough is faster, that's a sure sign that receiving the data across the wire and building the cache is the slow step there.  Also bear in mind that the cache is not bounded at this time, so very large trajectories could cause memory pressure.

src/simularium/rendering/InstancedMeshShader.ts
```diff
@@ -0,0 +1,151 @@
+import { FrontSide, Matrix4, ShaderMaterial } from "three";
+
+import { MultipassShaders } from "./MultipassMaterials";
+
+const vertexShader = `
+precision highp float;
+
+#ifdef WRITE_POS
```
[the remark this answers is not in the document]
toloudis: those are for configuration of the three separate render passes.  I want to use one vertex shader but 3 different fragment shaders.  If you just always write all the things from the vertex shader but the fragment shader doesn't read them all, you get glsl compiler warnings.  So this is a way to reuse the vertex shader code and still fix the warnings.

When we can use multiple render targets via ThreeJS, then those ifdefs will go away and there will be one vertex shader and one fragment shader only.

src/simularium/rendering/GBufferPass.ts
```diff
@@ -110,32 +114,22 @@ class GBufferPass {
         renderer.setRenderTarget(colorBuffer);
         renderer.autoClear = true;
 
-        const DO_MESHES = true;
         const DO_INSTANCED = true;
```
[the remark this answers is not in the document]
toloudis: I'd say that mostly these constants are for debugging purposes.  Generally they will always be true in normal operation.  If I want to debug or profile PDB drawing, for example, I can ensure that literally none of the instancing drawing will happen.

PR #113, opened 2021-05-21
Fix/floating point time errors

schoinh:
Problem
=======
Resolves [this bug](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1420) and hopefully prevents/fixes similar floating point time value bugs unbeknownst to us as well.

Solution
========
I added a `compareFloats()` util function that takes 2 numbers and lets us know which one is greater, or if they are pretty much equal with a configurable epsilon padding, so we can compare time values in functions like `VisData.gotoTime()` without getting thwarted by floating point precision issues.

`SimulariumController.gotoTime()` (which currently triggers bugs) calls either `this.visData.gotoTime()` or `this.simulator.gotoRemoteSimulationTime()` depending on whether we have frames cached in VisData or not. Time comparison logic is performed in `VisData.gotoTime()`, `VisData.hasLocalCacheForTime()`, and in `LocalFileSimulator.gotoRemoteSimulationTime()`. In these methods I now avoid floating point related errors by using the new `compareFloats()` util function to compare time values rather than just using greater/less/equal operators.

Thanks @toloudis for much help/advice.

## Type of change

Change summary:
---------------
* src/util.ts - new file containing the new `compareFloats()` function
* src/simularium/LocalFileSimulator.ts - refactoring for readability + use `compareFloats()` instead of plain comparisons
* src/simularium/VisData.ts - more refactoring for readability + use `compareFloats()` instead of plain comparisons

When `compareFloats()` is used to compare time values (like in this PR), epsilon is determined by the time step size for the trajectory file (epsilon = time step size * 0.01).

Steps to Verify:
----------------
1. Pull this branch
2. Run a local version of simularium-website with this branch as its viewer 
3. Load endocytosis
3. Try to repro the bug at the top of this PR description or just try to break the slider in general

After this merge there shouldn't be any bugs affecting the slider/playback buttons for networked trajectories. 🤞🏻🤞🏻 At least I'm not aware of any. There are, however, some non-floating point issues to be worked out for local/URL-download file playback though.

### DIFF
--- a/src/simularium/LocalFileSimulator.ts
+++ b/src/simularium/LocalFileSimulator.ts
@@ -157,20 +157,13 @@ export class LocalFileSimulator implements ISimulator {
     }
 
     public gotoRemoteSimulationTime(timeNs: number): void {
-        for (
-            let frame = 0,
-                numFrames = this.simulariumFile.spatialData.bundleData.length;
-            frame < numFrames;
-            frame++
-        ) {
-            const frameTime = this.simulariumFile.spatialData.bundleData[frame]
-                .time;
-            if (timeNs <= frameTime) {
-                const theFrameNumber = Math.max(frame, 0);
-                this.onTrajectoryDataArrive(this.getFrame(theFrameNumber));
-                break;
-            }
-        }
+        const { bundleData } = this.simulariumFile.spatialData;
+        const frameNumber = bundleData.findIndex(
+            (bundleData) => bundleData.time >= timeNs
+        );
+
+        // frameNumber is -1 if findIndex() above doesn't find a match
+        this.onTrajectoryDataArrive(this.getFrame(Math.max(frameNumber, 0)));
     }
 
     public requestTrajectoryFileInfo(_fileName: string): void {
--- a/src/simularium/LocalFileSimulator.ts
+++ b/src/simularium/LocalFileSimulator.ts
@@ -169,9 +169,7 @@ export class LocalFileSimulator implements ISimulator {
 
         // frameNumber is -1 if findIndex() above doesn't find a match
         if (frameNumber !== -1) {
-            this.onTrajectoryDataArrive(
-                this.getFrame(Math.max(frameNumber, 0))
-            );
+            this.requestSingleFrame(Math.max(frameNumber, 0));
         }
     }
 
--- a/src/simularium/LocalFileSimulator.ts
+++ b/src/simularium/LocalFileSimulator.ts
@@ -1,7 +1,7 @@
 import jsLogger from "js-logger";
 import { ILogger } from "js-logger";
 
-import { compareFloats } from "../util";
+import { compareTimes } from "../util";
 
 import {
     VisDataMessage,
@@ -164,7 +164,7 @@ export class LocalFileSimulator implements ISimulator {
 
         // Find the index of the frame that has the time matching our target time
         const frameNumber = bundleData.findIndex((bundleData) => {
-            return compareFloats(bundleData.time, timeNs, timeStepSize) === 0;
+            return compareTimes(bundleData.time, timeNs, timeStepSize) === 0;
         });
 
         // frameNumber is -1 if findIndex() above doesn't find a match
--- a/src/simularium/LocalFileSimulator.ts
+++ b/src/simularium/LocalFileSimulator.ts
@@ -169,7 +169,7 @@ export class LocalFileSimulator implements ISimulator {
 
         // frameNumber is -1 if findIndex() above doesn't find a match
         if (frameNumber !== -1) {
-            this.requestSingleFrame(Math.max(frameNumber, 0));
+            this.requestSingleFrame(frameNumber);
         }
     }
 
--- a/src/simularium/LocalFileSimulator.ts
+++ b/src/simularium/LocalFileSimulator.ts
@@ -1,6 +1,8 @@
 import jsLogger from "js-logger";
 import { ILogger } from "js-logger";
 
+import { compareFloats } from "../util";
+
 import {
     VisDataMessage,
     VisDataFrame,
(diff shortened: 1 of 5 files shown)


src/simularium/LocalFileSimulator.ts
```diff
@@ -157,20 +159,16 @@ export class LocalFileSimulator implements ISimulator {
...
+        const { bundleData } = this.simulariumFile.spatialData;
+        const { timeStepSize } = this.simulariumFile.trajectoryInfo;
+
+        // Find the index of the frame that has the time matching our target time
+        const frameNumber = bundleData.findIndex((bundleData) => {
+            return compareFloats(bundleData.time, timeNs, timeStepSize) === 0;
+        });
+
+        // frameNumber is -1 if findIndex() above doesn't find a match
+        this.onTrajectoryDataArrive(this.getFrame(Math.max(frameNumber, 0)));
```
toloudis, line 171: slight behavior change.  the old code never calls this function if it fails to find a time that satisfies the criteria

src/simularium/VisData.ts
```diff
@@ -369,36 +376,44 @@ class VisData {
...
+        const lastFrameTime = this.frameDataCache[
+            this.frameDataCache.length - 1
+        ].time;
+
+        // Edge cases
         if (
             this.frameDataCache.length > 0 &&
             timeNs === 0 &&
-            this.frameDataCache[0].time <= Number.EPSILON // allow for floating point errors
+            firstFrameTime <= Number.EPSILON // allow for floating point errors
```
toloudis, line 388: interesting: first frame time is assumed to be <= 0

src/simularium/VisData.ts
```diff
@@ -369,36 +376,44 @@ class VisData {
...
-        }
+        // Find the index of the frame that has the time matching our target time
+        const frameNumber = this.frameDataCache.findIndex((frameData) => {
+            return (
+                compareFloats(frameData.time, timeNs, this.timeStepSize) === 0
+            );
+        });
+
+        // frameNumber is -1 if findIndex() above doesn't find a match
+        this.cacheFrame = Math.max(frameNumber, 0);
```
toloudis, line 416: again, if the criteria is never met, do you want to still change this.cacheFrame?

src/test/util.test.ts
```diff
@@ -0,0 +1,48 @@
+import { compareFloats } from "../util";
+
+describe("util", () => {
+    describe("compareFloats", () => {
```
toloudis, line 4: Thank you!! Great job adding a unit test.

src/util.ts
```diff
@@ -0,0 +1,25 @@
+export const compareFloats = (
```
toloudis, line 1: Because of the 0.01 factor, I would consider naming this a little more specifically. Like compareTimes(time1, time2, timeStepSize) and the description would be `compares two times by testing to see whether they are within some small fraction of the time step size`.   The last 3 lines of the function are the true `compareFloats` but the epsilon multiplier makes it different.
Another way could be to pass the multiplier as a 4th argument but provide a default value:
compareTimes(time1, time2, timeStepSize, stepSizeFraction=0.01)
Just a suggestion. See what you think.

EricJIsaac (approved):

src/simularium/LocalFileSimulator.ts
toloudis: Do you want to return frame zero if the time passed in is out of range, or do nothing?  That's the behavior change I am seeing.  In the old code, it was possible to loop thru all values and never hit the condition, in which case nothing happens.  In the new code, it ALWAYS calls onTrajectoryDataArrive.

schoinh: I think you're right Dan, I will modify this (and VisData) to preserve the previous behavior in that regard

src/simularium/VisData.ts
schoinh: Oh true! But it should be ok since all other cases are handled by the final return statement `return notLessThanFirstFrameTime && notGreaterThanLastFrameTime;`. I'm not 100% sure these if statements are necessary, but I'll just keep them in for now?

src/test/util.test.ts
schoinh: 😄 It was necessary for my own sanity

src/simularium/VisData.ts
schoinh: Fixed!

src/util.ts
schoinh: Gotcha. I was trying to figure out the best place to put a global constant for that multiplier value and finally decided the best way to safeguard it from getting changed was to hard-code it inside the function 😆 I see what you're saying though. I'll turn it into a 4th argument with a default value.

src/simularium/LocalFileSimulator.ts
schoinh: I'm not sure if there is a point to `Math.max(frameNumber, 0))` anymore though. `frameNumber` should always be an integer greater than or equal to 0, or be -1 in the case `findIndex()` doesn't find a match

src/simularium/LocalFileSimulator.ts
```diff
@@ -157,19 +159,17 @@ export class LocalFileSimulator implements ISimulator {
...
+        const { timeStepSize } = this.simulariumFile.trajectoryInfo;
+
+        // Find the index of the frame that has the time matching our target time
+        const frameNumber = bundleData.findIndex((bundleData) => {
+            return compareTimes(bundleData.time, timeNs, timeStepSize) === 0;
+        });
+
+        // frameNumber is -1 if findIndex() above doesn't find a match
+        if (frameNumber !== -1) {
+            this.requestSingleFrame(Math.max(frameNumber, 0));
```
toloudis, line 172: yup it looks like you don't need the Math.max here.  it's not breaking anything but you could remove it for clarity

schoinh: Removed

toloudis (approved):

PR #116, opened 2021-06-03
add a function to switch between left-mouse pan or left-mouse rotate

toloudis:
Problem
=======
Allow switching between pan mode or rotate mode.
Resolves https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1337

Solution
========
Add a function called setPanningMode(boolean) where true causes left-drag to have pan behavior, and false causes left-drag to rotate.
The function is to be called explicitly on a tool button click.

## Type of change

Steps to Verify:
----------------
Can run the testbed viewer - there's a Toggle Pan/Rotate button.

-----------------------
Some minor changes to consolidate the button layout in the test viewer

### DIFF
--- a/src/simularium/VisGeometry.ts
+++ b/src/simularium/VisGeometry.ts
@@ -30,6 +30,7 @@ import {
     WebGLRendererParameters,
     Mesh,
     Quaternion,
+    MOUSE,
 } from "three";
 
 import * as dat from "dat.gui";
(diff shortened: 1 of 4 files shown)


src/simularium/VisGeometry.ts
```diff
@@ -30,6 +30,7 @@ import {
     WebGLRendererParameters,
     Mesh,
     Quaternion,
+    MOUSE,
```
schoinh, line 33: Strange, I can't find an entry for `Mouse` or `MOUSE` in the threejs docs!

src/simularium/VisGeometry.ts
```diff
@@ -477,6 +478,26 @@ class VisGeometry {
...
     }
 
+    public setPanningMode(pan: boolean): void {
+        if (!pan) {
+            this.controls.enablePan = false;
+            this.controls.enableRotate = true;
+            this.controls.mouseButtons = {
+                LEFT: MOUSE.ROTATE,
+                MIDDLE: MOUSE.DOLLY,
+                RIGHT: MOUSE.PAN,
```
schoinh, line 488: Whoa, we can do left = rotate and right = pan? This almost sounds like we can just have this and not bother with hot keys/buttons :P

schoinh (approved):

src/simularium/VisGeometry.ts
toloudis: https://threejs.org/docs/?q=orbit#examples/en/controls/OrbitControls.keys

toloudis: https://threejs.org/docs/?q=orbit#examples/en/controls/OrbitControls.mouseButtons

meganrm (approved):

PR #128, opened 2021-06-24
Fix/file switching bugs

### LINKED ISSUE #126: Displayed end time for local file is incorrect when loaded after networked file
## Description

Displayed end time for local file is incorrect when loaded after loading networked file.

## Expected Behavior

The displayed local file end time should be accurate.

## Reproduction

1. Download this file: https://www.dropbox.com/s/j2emic0jecq2pfm/kinesin002_01.h5.simularium?dl=0
2. Load the file to staging via drag and drop, notice the total time is 50 ns (correct)
3. Refresh the viewer and load endocytosis from dropdown
4. Load the kinesin file from Dropbox by drag and drop again
5. Notice the total time is now shown as 50.1 ns (unexpected/incorrect)

### LINKED ISSUE #127: Site crashes when switching models
## Description

Sometimes, when switching from one networked file to another, the site crashes, showing just a white background in the browser window, and the console shows multiple error messages about websocket connection.

image

## Expected Behavior

Site should change files without crashing.

## Reproduction

1. Go to staging or prod and load endocytosis
2. Click on the slider to skip to 10 seconds or so
3. Load COVID from dropdown
4. Site crashes

## Comments from Jira issue
* @toloudis - The first error looks pretty relevant:  attempting to send while socket is in CONNECTING state.  We have to disallow immediate sends if the socket is not yet ready. It could probably be prevented inside of RemoteSimulator but then we'd have to decide what to do on the UI side. can use RemoteSimulator.socketIsConnected
* @EricJIsaac - The simularium-viewer exhibits the correct behavior locally. Checking the simularium-website now.
* @EricJIsaac - This bug is not present in previous versions of the simularium-website. Testing to determine which commit(s) introduced it.
* @EricJIsaac - Bug introduced to simularium-website in commit:
7f0fb26d95a5543c11e1e9c8bbd207288156f949 Fix/floating point time errors (update viewer to v2.9.2) (#129)

schoinh:
Problem
=======
Closes #126
Closes #127 

Solution
========
### Short explanation: 
Prevent `simulariumController.gotoTime` from requesting a new time when the viewer is in the middle of changing files.

### Long explanation: 

The Antd Slider component's `componentDidUpdate` method is fired for some reason (probably because its parent's props change, but regardless of the reason, I don't think I have control over it?) when we're switching files but when we haven't finished loading the new file. And when that method fires, it triggers its `onChange` handler which eventually calls `simulariumController.gotoTime(arg)`, with the `arg` being the time that the slider was displaying for the trajectory we're trying to change out. This all leads to time data from the old trajectory file coming into and mixing with the cache of the new trajectory file and interfering with the determination of the new file's `firstFrameTime` and `lastFrameTime` among other things. 

`simulariumController` already has a boolean property called `fileChanged` that becomes `true` for the duration of a file change (loading period). If we disable `simulariumController.gotoTime()` when a file is still loading then we can prevent unwanted new time requests/messages during the file change period. This seems to fix the above two issues.

I think that the reason these bugs came up after [this simularium-website commit](https://github.com/allen-cell-animated/simularium-website/pull/129) is because the rounding of time values we were doing previously were simply masking some issues. 

## Type of change

Steps to Verify:
----------------
1. Pull this branch and link your local viewer to a locally running simularium-website
2. Try to reproduce the bugs above, and also check that you can switch files without problem in general

### DIFF
--- a/src/controller/index.ts
+++ b/src/controller/index.ts
@@ -64,7 +64,6 @@ export default class SimulariumController {
     private networkEnabled: boolean;
     private isPaused: boolean;
     private fileChanged: boolean;
-    private canRequestNewTime: boolean;
     private playBackFile: string;
     // used to map geometry to agent types
     private geometryFile: string;
@@ -119,7 +118,6 @@ export default class SimulariumController {
         this.networkEnabled = true;
         this.isPaused = false;
         this.fileChanged = false;
-        this.canRequestNewTime = true;
         this.playBackFile = params.trajectoryPlaybackFile || "";
         this.geometryFile = this.resolveGeometryFile(
             params.trajectoryGeometryFile || "",
@@ -244,7 +242,8 @@ export default class SimulariumController {
     }
 
     public gotoTime(timeNs: number): void {
-        if (this.canRequestNewTime === false) return;
+        // If in the middle of changing files, ignore any gotoTime requests
+        if (this.fileChanged === true) return;
         if (this.visData.hasLocalCacheForTime(timeNs)) {
             this.visData.gotoTime(timeNs);
         } else {
@@ -293,7 +292,6 @@ export default class SimulariumController {
         assetPrefix?: string
     ): Promise<FileReturn> {
         this.fileChanged = true;
-        this.canRequestNewTime = false;
         this.playBackFile = newFileName;
         this.geometryFile = this.resolveGeometryFile(
             geometryFile || "",
@@ -336,7 +334,6 @@ export default class SimulariumController {
             return this.start()
                 .then(() => {
                     if (this.simulator) {
-                        this.canRequestNewTime = true;
                         this.simulator.requestSingleFrame(0);
                     }
                 })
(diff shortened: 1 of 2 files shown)


toloudis (approved):
Nice fix but it leads me to think of two things:
1. should we be doing this check in more places?  you kind of want to disable a whole lot of things during that transitional time
2. should the variable have a better name like "isChanging" rather than "changed" and use something like `trajectory` instead of `file`...   naming things is hard.

blairlyons (approved):

EricJIsaac (approved):

schoinh:
> Nice fix but it leads me to think of two things:
> 
> 1. should we be doing this check in more places?  you kind of want to disable a whole lot of things during that transitional time

Probably. I'm not sure exactly where else at this point but will keep an eye out.

> 2. should the variable have a better name like "isChanging" rather than "changed" and use something like `trajectory` instead of `file`...   naming things is hard.

Agreed, I changed `fileChanged` to `isFileChanging` but left `file` as `file` for consistency with other functions that refer to `file`.

Issue #129, opened 2021-06-30
Skipping frames back sometimes makes viewer hang


schoinh:
## Description

Clicking the "skip 1 frame back" button on the networked endocytosis model sometimes makes the viewer hang. The playback buttons get stuck in a forever loading state and the only way to fix it is to refresh the page.

## Expected Behavior

Skip frame buttons should work without making things hang.

### When viewer is installed in simularium-website
1. On either staging or prod, load endocytosis
2. Click on the slider to jump 10 seconds or so
3. Click the Skip 1 Frame Back button
4. Click it again, and again
5. At about the 2nd or 3rd click, the viewer hangs (all playback components get stuck in a forever loading state)

### When viewer testbed is run alone
1. Launch the viewer testbed
2. Select endocytosis from the dropdown and click Load
3. Click on the slider to jump 10 seconds or so
3. Click the Skip 1 Frame Back button
4. Click it again, and again
5. After about 2 or 3 clicks, the Skip 1 Frame Back button stops working (doesn't do anything)

## Comments from Jira issue
* @schoinh: Looked into this more with debugger – I think the issue might be in the backend, because I don't see anything wrong with what happens on the client side after clicking the Skip 1 Frame Back button. The viewer code correctly determines that we don't have the desired time/frame in our cache and sends a websocket message like `{msgType: 13, time: 9.299999617040157}` via `RemoteSimulator.gotoRemoteSimulationTime`. But then it seems the client sits there forever waiting for a response.

schoinh:
Temporarily fixed with #131 , and resolving [this issue in simularium-engine](https://github.com/allen-cell-animated/simularium-engine/issues/14) will hopefully resolve this bug long term.

PR #131, opened 2021-07-01
Fix/skip frame bugs (temporary fix)

schoinh:
Problem
=======
At least 2 known bugs associated with the skip frame button: #129, #130 

The bugs above appeared (albeit not noticed right away) when [this commit](https://github.com/allen-cell-animated/simularium-website/pull/129) was merged to simularium-website. Prior to this merge, the front end was rounding all time values to an arbitrary number of significant figures before requesting a frame with a specific time from the back end. Since the merge, the front end has been sending unrounded, imprecise time values to the back end when requesting a frame at a specific time -- like 10.499999999991234 or 10.5000000000153, when the desired time is 10.5 seconds. This works fine when the actual desired time is less than the requested time (as in the case of 10.5000000000153), but when the desired time is slightly greater than the requested time (as in the case of 10.499999999991234), the back end sends a frame at 10.4 seconds instead of 10.5. 

Solution
========
I think the real fix for this problem is to update the simularium-engine to return the frame that has the timestamp closest to the requested time, instead of returning the frame with the timestamp closest to _and_ less than the requested time. However we're currently blocked on that task, so as a temporary fix, this PR proposes bringing back the arbitrary rounding of time values that simularium-website used to do before (see the old version of line 303 in simularium-website/src/containers/ViewerPanel/index.tsx [here](https://github.com/allen-cell-animated/simularium-website/pull/129/files)), but in simularium-viewer instead (immediately before a websocket request is made).

## Type of change

Steps to Verify:
----------------
1. Pull this branch and `npm start`
2. Launch the viewer testbed and check that the bugs above aren't reproducible anymore
3. Install this branch as a dependency for a local copy of simularium-website and check that the bugs aren't reproducible in the simularium-website dev environment either.

### DIFF
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -234,7 +234,7 @@ class Viewer extends React.Component<{}, ViewerState> {
     }
 
     public handleScrubTime(event): void {
-        simulariumController.gotoTime(event.target.value);
+        simulariumController.gotoTime(parseFloat(event.target.value));
     }
 
     public handleUIDisplayData(uiDisplayData: UIDisplayData): void {
--- a/src/controller/index.ts
+++ b/src/controller/index.ts
@@ -252,7 +252,7 @@ export default class SimulariumController {
                 //  and play remotely from the desired simulation time
                 this.visData.clearCache();
 
-                // This arbitrary rounding of timeNs is a temporary fix until
+                // NOTE: This arbitrary rounding of timeNs is a temporary fix until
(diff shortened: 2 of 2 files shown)


toloudis (approved):
we really need to get rid of the `Ns` in that variable name.

meganrm (approved):

schoinh:
> we really need to get rid of the `Ns` in that variable name.

Agreed! It's on my list

PR #135, opened 2021-07-01
Fix/zoom while pan

### LINKED ISSUE #125: Changing to pan mode breaks zoom in and out functions
## Description

After switching to pan mode, clicking the plus button to zoom in rotates and zooms. Clicking the minus often has no effect.

## Expected Behavior

The zoom in and out buttons should work the same regardless of pan/rotate mode state

## Reproduction

Load a model
click pan/rotate mode button
click (+) and (-)

meganrm:
Problem
=======
When you had panned off of center and then clicked the zoom buttons, the camera would rotate weirdly. 
resolves #125 

Solution
========
The camera position was not being set to the correct vector in world space. We had calculated the vector that was the correct length, and the correct orientation, and as long as the target was 0,0,0 that was enough. But once we pan, the vector needed to be added to the new target position, which was no longer 0,0,0

with @toloudis 

## Type of change

Change summary:
---------------
* changed the new camera position to the target vector + the vector that represents the new distance from the target
* I also combined the two functions into one dolly function that just takes the change by scalar as an argument

Steps to Verify:
----------------
1. npm start
2. load a model
3. switch to pan mode, pan off center
4. click the (+) and (-) buttons
5. should zoom in and out the same as scrolling your mouse does

### DIFF
--- a/src/simularium/VisGeometry.ts
+++ b/src/simularium/VisGeometry.ts
@@ -466,6 +466,7 @@ class VisGeometry {
     public zoomIn(): void {
         const position = this.camera.position.clone();
         const target = this.controls.target.clone();
+        console.log(target);
         const distance = position.distanceTo(target);
         const newDistance = distance - CAMERA_DOLLY_STEP_SIZE;
         const newPosition = new Vector3()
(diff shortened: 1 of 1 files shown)


toloudis (approved):

schoinh (approved):
Awesome! Do you know why the zoom in/out by mouse scroll worked regardless?

meganrm:
> Awesome! Do you know why the zoom in/out by mouse scroll worked regardless?

it's handled by the orbit controls in THREEjs but they don't expose that function, so we can't call it.

PR #136, opened 2021-07-02
Feature/remove dup camera function code

meganrm:
Problem
=======
At one point, visGeometry was not exposed to the controller, so we had to make stubbed out functions on the controller, and then pass those functions over through the viewer. However, at some point I added the line that hands the visGeometry class to the controller once the viewer is mounted. So this code to define the functions in both places isn't needed anymore. 

Solution
========
Removed camera control functions from the viewer. 

## Type of change

Steps to Verify:
----------------
1. npm start
2. load a model
3. click all the camera control buttons, verify that they work the same as before 

Thanks for contributing!


toloudis (approved):
Much cleaner. I like it.

schoinh (approved):
👏🏻

PR #138, opened 2021-07-06
add github action to show testing coverage

meganrm:
Problem
=======
We want to measure our test coverage 

Solution
========
1. Add coverage to dev report
2. Post coverage report on PRs using a [github action ](https://github.com/marketplace/actions/jest-coverage-report?version=v1.3). 
We can also have to do inline reporting that shows when a function isn't covered by a test, but that will be overwhelming at this moment. 

## Type of change

## Steps to verify:
1. npm run test
2. should see a coverage print out


schoinh (approved):
Thanks for doing this! I know we didn't talk about simularium-viewer specifically during the meeting earlier, but looks like we can set up thresholds for specific file paths, which would be more informative https://jestjs.io/docs/configuration#coveragethreshold-object

meganrm:
> Thanks for doing this! I know we didn't talk about simularium-viewer specifically during the meeting earlier, but looks like we can set up thresholds for specific file paths, which would be more informative https://jestjs.io/docs/configuration#coveragethreshold-object

yeah, that's a good idea. The print out that you get locally in also more informative than what gets posted here in the PR. (it is broken down by file). That will be especially helpful for wanting the best coverage on selectors on the website.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,15 @@
+name: 'coverage'
+on:
+    pull_request:
+        branches:
+            - master
```
toloudis, line 5: can we remove `master` from here?

toloudis (approved):

PR #168, opened 2021-08-02
update dependencies

toloudis:
Problem
=======
Bring some dependencies up to date and remove a few known vulnerabilities.

Solution
========
Update several dependency versions and do the few associated code and config fixes.
Functionality should not change.

Change summary:
---------------
Web Workers are simpler to load with the updated webpack.  Also, the options for copy-webpack-plugin are slightly different now.

Steps to Verify:
----------------
Functionality should not change.


toloudis:
oops, I need to fix this. will comment again when it's green.

toloudis:
I'm blocked. 
The remaining problem is that Jest doesn't play well with ESM modules.  Updating webpack simplifies the handling of Web Workers, but requires you to use `import.meta.url` for it.  Jest is not as far along as webpack in dealing with import.meta.url.  I'm still looking for the proper jest+ts+babel config to make the tests parse correctly.  Functionally they should be fine.

I'm going to turn this back into a draft PR until this is resolved.

toloudis:
the `esbuild` package allowed jest to do the right transforms to make the tests work.  It might be worth considering whether esbuild is a worthwhile addition to other parts of the tool chain.

schoinh (approved):
Yay! I just tested this out (symlinking to website and running) and seems like it's all working! I did notice something funky, and I'm not sure if this has to do with the dependency updates, but the cell membrane is not showing up in endocytosis anymore. (It shows up on staging)

toloudis:
> Yay! I just tested this out (symlinking to website and running) and seems like it's all working! I did notice something funky, and I'm not sure if this has to do with the dependency updates, but the cell membrane is not showing up in endocytosis anymore. (It shows up on staging)

I think this might be because .obj geometry is no longer working, because the file is missing a trajectoryfileinfo

EricJIsaac (approved):

PR #172, opened 2021-08-18
fix #170

### LINKED ISSUE #170: Perfectly aligned fiber does not render
## Description
If a fiber is aligned perfectly with an axis of the view, it does not render.

## Expected Behavior
Fiber should render regardless of its subpoint positions.

## Reproduction
The only difference between these two files is one of the fiber's z positions.
fiber_on_axis: filament at [-70, 0, 0] to [10, 0, 0]
fiber_off_axis: filament at [-70, 0, 0] to [10, 0, 0.001]
[fiber.zip](https://github.com/allen-cell-animated/simularium-viewer/files/6957548/fiber.zip)

## Environment
Can reproduce on production and staging

toloudis:
Problem
=======
fix #170 : a straight line fiber exactly aligned along a cardinal axis will not render.

Solution
========
Special case when tangent frame lines up with curve (perfectly linear curves) the Binormal and Normal are not well-defined, so establish a convention to get it to render.
This is at some slight performance cost but it doesn't seem to be an issue unless you have large numbers of fibers.

## Type of change

Change summary:
---------------
A test case was added to track this down ,which amounts to porting the GLSL shader to javascript so it can be run in a test.

### DIFF
--- a/src/simularium/rendering/InstancedFiberShader.ts
+++ b/src/simularium/rendering/InstancedFiberShader.ts
@@ -271,8 +271,9 @@ void createTube (float t, vec2 volume, out vec3 offset, out vec3 normal) {
 
   // if next-prev and next+prev are parallel, then
   // our normal and binormal will be ill-defined so we need to check for that
-  float check = dot(next - prev, next + prev);
-  if (check > 0.00001) {
+  // check parallel by dot the unit vectors and check close to +/-1
+  float check = dot(T, normalize(next + prev));
+  if (abs(check) < 0.999) {
     // cross product of parallel vectors is 0, so T must not be parallel to next+prev
     // hence the above check.
     B = normalize(cross(T, next + prev));
@@ -281,15 +282,15 @@ void createTube (float t, vec2 volume, out vec3 offset, out vec3 normal) {
     // special case for which N and B are not well defined. 
     // so we will just pick something
     float min = 1.0;
-    if (T.x <= min) {
-      min = T.x;
+    if (abs(T.x) <= min) {
+      min = abs(T.x);
       B.x = 1.0;
     }
-    if (T.y <= min) {
-      min = T.y;
+    if (abs(T.y) <= min) {
+      min = abs(T.y);
       B.y = 1.0;
     }
-    if (T.z <= min) {
+    if (abs(T.z) <= min) {
       B.z = 1.0;
     }
     vec3 tmpVec = normalize(cross(T, B));
--- a/src/simularium/rendering/InstancedFiberShader.ts
+++ b/src/simularium/rendering/InstancedFiberShader.ts
@@ -159,10 +159,10 @@ vec3 getTangent (vec3 a, vec3 b) {
   return normalize(b - a);
 }
 
-void rotateByAxisAngle (inout vec3 normal, vec3 axis, float angle) {
+void rotateByAxisAngle (inout vec3 normal, vec3 axis, float rotAngle) {
   // http://www.euclideanspace.com/maths/geometry/rotations/conversions/angleToQuaternion/index.htm
   // assumes axis is normalized
-  float halfAngle = angle / 2.0;
+  float halfAngle = rotAngle / 2.0;
   float s = sin(halfAngle);
   vec4 quat = vec4(axis * s, cos(halfAngle));
   normal = normal + 2.0 * cross(quat.xyz, cross(quat.xyz, normal) + quat.w * normal);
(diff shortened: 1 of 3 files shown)


src/simularium/rendering/InstancedFiberShader.ts
```diff
@@ -264,7 +264,30 @@ void createTube (float t, vec2 volume, out vec3 offset, out vec3 normal) {
 
   // compute the TBN matrix
   vec3 T = normalize(next - prev);
-  vec3 B = normalize(cross(T, next + prev));
+  vec3 B = vec3(0,0,0);
+  // if normalize(next-prev) and next+prev are parallel, then B will be zero
```
meganrm, line 268: This comment was initially more confusing than helpful, since `B` is initialized to be 0 right above it it seemed strange to reassert that it's zero. But then reading the whole section, I'm still a bit confused. Is it justifying why you initialized B as zero?  I also don't understand this well enough overall to understand the `check`/ "special case for which N and B are not well defined". If possible it would be nice to have clearer explanation

src/simularium/rendering/InstancedFiberShader.ts
```diff
@@ -264,7 +264,30 @@ void createTube (float t, vec2 volume, out vec3 offset, out vec3 normal) {
...
+  float check = dot(next-prev, next+prev);
```
meganrm, line 269: nit, inconsistent spacing around operators , not sure why the linter didn't pick this up

src/simularium/rendering/InstancedFiberShader.ts
```diff
@@ -264,7 +264,30 @@ void createTube (float t, vec2 volume, out vec3 offset, out vec3 normal) {
...
+  if (check > 0.00001) {
+    B = normalize(cross(T, next + prev));
+  }
+  else {
+    // special case for which N ad B are not well defined.
```
meganrm, line 274: typo: `ad`

meganrm (approved):

src/simularium/rendering/InstancedFiberShader.ts
toloudis: no prob, i think i can make it clearer

src/simularium/rendering/InstancedFiberShader.ts
toloudis: this doesn't get linted because it is a long inline string written in webGL

src/simularium/rendering/InstancedFiberShader.ts
toloudis: the init to zero was really just a safety thing and has nothing to do with what follows.  I tried to improve the comments

meganrm: thanks! it's a lot clearer to me now

toloudis:
commit 0437bd8 fixes an egregious error in my check for parallel vectors.  when you reverse the direction of the curves the mistake became apparent, so I added a unit test to check that.

blairlyons (approved):

src/simularium/rendering/InstancedFiberShader.ts
```diff
@@ -262,9 +262,42 @@ void createTube (float t, vec2 volume, out vec3 offset, out vec3 normal) {
...
+  // check parallel by dot the unit vectors and check close to +/-1
+  float check = dot(T, normalize(next + prev));
+  if (abs(check) < 0.999) {
+    // cross product of parallel vectors is 0, so T must not be parallel to next+prev
+    // hence the above check.
+    B = normalize(cross(T, next + prev));
+  }
+  else {
+    // special case for which N and B are not well defined. 
+    // so we will just pick something
```
schoinh, line 283: I think it would be helpful to put here "so we will just pick an arbitrary vector that's not parallel to T" (if I remembered that correctly?). Actually, how does this algorithm ensure finding a vector that's not parallel to T?

src/simularium/rendering/InstancedFiberShader.ts
```diff
@@ -171,11 +171,11 @@ void rotateByAxisAngle (inout vec3 normal, vec3 axis, float angle) {
 void createTube (float t, vec2 volume, out vec3 outPosition, out vec3 outNormal) {
   // Reference:
   // https://github.com/mrdoob/three.js/blob/b07565918713771e77b8701105f2645b1e5009a7/src/extras/core/Curve.js#L268
-  float nextT = t + (1.0 / lengthSegments);
+  float nextT = t + (1.0 / float(lengthSegments));
```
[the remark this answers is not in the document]
toloudis: These are in a code branch that is effectively commented out, but I was enabling it when testing/debugging.  It was not compiling without the float conversions.

schoinh (approved):

src/simularium/rendering/InstancedFiberShader.ts
toloudis: It's not this exact algorithm but it's something similar to these: 
https://math.stackexchange.com/questions/3122010/how-to-deterministically-pick-a-vector-that-is-guaranteed-to-be-non-parallel-to
I actually feel a bit more comfortable with those answers so I might change this at some point.

PR #173, opened 2021-08-19
Fix/warn if waiting too long to connect to server

### LINKED ISSUE #169: Warn if waiting too long for server connection
## Use Case
While we're working to set up autoscaling on the backend, if too many concurrent users use the website, they could overload the server. It would be nice if a client could warn them that the server is under high load rather than the blank viewer spinning forever.

## Acceptance Criteria
An app installing the viewer gets a warning if the viewer has been waiting on a connection request for more than 4(?) sec

## Details
Eventually the server should send a warning if it's overloaded, but until we get to that devops work, we can use a timeout to give the user some feedback.

Possible scenarios of issues with websockets connection:
**1. fail to establish connection**
2. connect, but becomes invalid
3. valid connection, but slow operation (lower priority since we're prioritizing posted examples which have been cached)

This handles the case when we fail to connect (1).

schoinh:
Problem
=======
Closes #169 

Solution
========
The timeout for connecting to the server was already built into `RemoteSimulator` but the error message was not getting propagated up as a `FrontEndError`. So I made that error bubble up from `connectToRemoteServer` to `startRemoteTrajectoryPlayback` where it would be thrown as a `FrontEndError` that would be displayed in the UI. I also added some tests.

While working on the above I refactored `connectToRemoteServer` to be more testable and hopefully a little clearer in general.

Thanks to @meganrm for help writing tests!

## Type of change

Steps to Verify:
----------------
1. Test should pass
2. `npm start` and check that networked models load in the viewer testbed with no new problems

-----------------------
1. src/simularium/RemoteSimulator.ts
2. src/test/RemoteSimulator.test.ts

Thanks for contributing!

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -19,3 +19,8 @@ export const DEFAULT_CAMERA_SPEC: CameraSpec = {
     },
     fovDegrees: 75,
 };
+
+export const TEST_CONNECTION_SETTINGS = {
+    serverIp: "staging-node1-agentviz-backend.cellexplore.net",
+    serverPort: 9002,
+};
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -19,8 +19,3 @@ export const DEFAULT_CAMERA_SPEC: CameraSpec = {
     },
     fovDegrees: 75,
 };
-
-export const TEST_CONNECTION_SETTINGS = {
-    serverIp: "staging-node1-agentviz-backend.cellexplore.net",
-    serverPort: 9002,
-};
--- a/src/simularium/ISimulator.ts
+++ b/src/simularium/ISimulator.ts
@@ -18,8 +18,6 @@ export interface ISimulator {
     /**
      * Connect
      * */
-    createWebSocket(uri: string): void;
-
     disconnect(): void;
 
     getIp(): string;
--- a/src/simularium/ISimulator.ts
+++ b/src/simularium/ISimulator.ts
@@ -18,7 +18,7 @@ export interface ISimulator {
     /**
      * Connect
      * */
-    connectToUri(uri: string): void;
+    createWebSocket(uri: string): void;
 
     disconnect(): void;
 
--- a/src/simularium/RemoteSimulator.ts
+++ b/src/simularium/RemoteSimulator.ts
@@ -1,5 +1,6 @@
 import jsLogger from "js-logger";
 import { ILogger } from "js-logger";
+import FrontEndError from "./FrontEndError";
 
 import { ISimulator } from "./ISimulator";
 import { TrajectoryFileInfoV2, VisDataMessage } from "./types";
@@ -264,7 +265,10 @@ export class RemoteSimulator implements ISimulator {
         return connectPromise;
     }
 
-    public connectToRemoteServer(address: string): Promise<string> {
+    public connectToRemoteServer(
+        address: string,
+        timeout = 1000
+    ): Promise<string> {
         const remoteStartPromise = new Promise<string>((resolve, reject) => {
             if (this.socketIsConnected()) {
                 return resolve("Remote sim successfully started");
@@ -276,15 +280,15 @@ export class RemoteSimulator implements ISimulator {
                 new Promise((resolve) =>
                     setTimeout(() => {
                         resolve(this.socketIsConnected());
-                    }, 1000)
+                    }, timeout)
                 );
 
             const handleReturn = async () => {
                 const isConnected = await waitForIsConnected();
-                secondsWaited++;
+                timeWaited += timeout;
                 if (isConnected) {
                     resolve("Remote sim successfully started");
-                } else if (secondsWaited < TOTAL_WAIT_SECONDS) {
+                } else if (timeWaited < MAX_WAIT_TIME) {
                     return await handleReturn();
                 } else if (connectionTries <= MAX_CONNECTION_TRIES) {
                     connectionTries++;
@@ -297,8 +301,8 @@ export class RemoteSimulator implements ISimulator {
                     );
                 }
             };
-            const TOTAL_WAIT_SECONDS = 2;
-            let secondsWaited = 0;
+            const MAX_WAIT_TIME = 2 * timeout;
+            let timeWaited = 0;
             const MAX_CONNECTION_TRIES = 2;
             let connectionTries = 1;
             return startPromise.then(handleReturn);
@@ -404,12 +408,16 @@ export class RemoteSimulator implements ISimulator {
 
         // begins a stream which will include a TrajectoryFileInfo and a series of VisDataMessages
         // Note that it is possible for the first vis data to arrive before the TrajectoryFileInfo...
-        return this.connectToRemoteServer(this.getIp()).then(() => {
-            this.sendWebSocketRequest(
-                jsonData,
-                "Start Trajectory File Playback"
-            );
-        });
+        return this.connectToRemoteServer(this.getIp())
+            .then(() => {
+                this.sendWebSocketRequest(
+                    jsonData,
+                    "Start Trajectory File Playback"
+                );
+            })
+            .catch((error) => {
+                throw new FrontEndError(error);
+            });
     }
 
     public pauseRemoteSim(): void {
--- a/src/simularium/RemoteSimulator.ts
+++ b/src/simularium/RemoteSimulator.ts
@@ -277,7 +277,7 @@ export class RemoteSimulator implements ISimulator {
 
     public async checkConnection(
         address: string,
-        timeout: number,
+        timeout = 1000,
         maxRetries = 1
     ): Promise<boolean> {
         // Initially wait for a max wait time of maxWaitTime, then retry
@@ -302,10 +302,7 @@ export class RemoteSimulator implements ISimulator {
         }
     }
 
-    public async connectToRemoteServer(
-        address: string,
-        timeout = 1000
-    ): Promise<string> {
+    public async connectToRemoteServer(address: string): Promise<string> {
         this.connectionTimeWaited = 0;
         this.connectionRetries = 0;
 
@@ -314,10 +311,7 @@ export class RemoteSimulator implements ISimulator {
         }
 
         this.createWebSocket(address);
-        const isConnectionSuccessful = await this.checkConnection(
-            address,
-            timeout
-        );
+        const isConnectionSuccessful = await this.checkConnection(address);
 
         if (isConnectionSuccessful) {
             return CONNECTION_SUCCESS_MSG;
--- a/src/simularium/RemoteSimulator.ts
+++ b/src/simularium/RemoteSimulator.ts
@@ -280,6 +280,9 @@ export class RemoteSimulator implements ISimulator {
         timeout = 1000,
         maxRetries = 1
     ): Promise<boolean> {
+        // Check if the WebSocket becomes connected within an allotted amount
+        // of time and number of retries.
+
         // Initially wait for a max wait time of maxWaitTime, then retry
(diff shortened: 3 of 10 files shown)


src/constants.ts
```diff
@@ -19,3 +19,8 @@ export const DEFAULT_CAMERA_SPEC: CameraSpec = {
     },
     fovDegrees: 75,
 };
+
+export const TEST_CONNECTION_SETTINGS = {
```
schoinh, line 23: I feel iffy about this variable name, better ideas welcome

src/simularium/RemoteSimulator.ts
```diff
@@ -255,53 +260,52 @@ export class RemoteSimulator implements ISimulator {
...
     }
 
-    private connectToUriAsync(address): Promise<string> {
-        const connectPromise = new Promise<string>((resolve) => {
-            this.connectToUri(address);
-            resolve("Successfully connected to uri!");
-        });
-
-        return connectPromise;
-    }
```
schoinh, line 265: After discussing WebSocket object creation with @EricJIsaac , I thought this function was unnecessary, since it's an async wrapper for a function (`connectToUri`) which doesn't have any async operations?

src/simularium/RemoteSimulator.ts
```diff
@@ -255,53 +260,52 @@ export class RemoteSimulator implements ISimulator {
...
-
-    public connectToRemoteServer(address: string): Promise<string> {
+    public connectToRemoteServer(
+        address: string,
+        timeout = 1000
```
schoinh, line 265: New parameter `timeout` allows testing with a very short timeout (like 1 instead of 1000 ms), as a way of simulating a slow connection

src/simularium/RemoteSimulator.ts
```diff
@@ -255,53 +260,52 @@ export class RemoteSimulator implements ISimulator {
...
                 );
 
-            const handleReturn = async () => {
-                const isConnected = await waitForIsConnected();
-                secondsWaited++;
+            /*
+            Initially wait for a max wait time of MAX_WAIT_TIME, then retry connecting 
+            <MAX_RETRIES> time(s). In a retry, only wait for the amount of time 
+            specified as timeout. (Can reset timeWaited to 0 upon retry if we want to
+            wait for MAX_WAIT_TIME in a retry too)
```
schoinh, line 289: Not positive if this is what we want, but I didn't change it

src/simularium/RemoteSimulator.ts
```diff
@@ -314,7 +318,7 @@ export class RemoteSimulator implements ISimulator {
         this.logger.debug("Web Socket Request Sent: ", whatRequest, jsonData);
     }
 
-    private sendWebSocketRequest(jsonData, requestDescription): void {
+    public sendWebSocketRequest(jsonData, requestDescription: string): void {
```
schoinh, line 321: Needed this public for testing

src/simularium/RemoteSimulator.ts
```diff
@@ -404,12 +408,16 @@ export class RemoteSimulator implements ISimulator {
...
+        return this.connectToRemoteServer(this.getIp())
+            .then(() => {
+                this.sendWebSocketRequest(
+                    jsonData,
+                    "Start Trajectory File Playback"
+                );
+            })
+            .catch((error) => {
+                throw new FrontEndError(error);
+            });
```
schoinh, line 420: This is the real fix that should get the error message "working"

src/test/RemoteSimulator.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+            try {
+                await simulator.connectToRemoteServer(
+                    simulator.getIp(),
+                    veryShortTimeout
+                );
+            } catch (error) {
+                expect(error).toEqual(new Error(CONNECTION_FAIL_MSG));
+            }
+        });
```
schoinh, line 44: Key test number 1

src/test/RemoteSimulator.test.ts
```diff
@@ -0,0 +1,73 @@
...
+            );
+
+            try {
+                await simulator.startRemoteTrajectoryPlayback(
+                    "endocytosis.simularium"
+                );
+            } catch (error) {
+                expect(error).toEqual(new FrontEndError("Mock error message"));
+            }
+        });
```
schoinh, line 71: Key test number 2

src/simularium/RemoteSimulator.ts
toloudis: It looks to me like we are waiting for a total of MAX_WAIT_TIME over many possible connection retries.  I am led to think that because of line 293 timeWaited+=timeout.  If I am correct, the comment should say something like "wait timeout time, try to reconnect, until a total of MAX_WAIT_TIME has elapsed before giving up"

src/test/RemoteSimulator.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Silence console.debug messages like this in Jest output:
+    // "[netconnection] WS Connection Request Sent:  wss://..."
+    jest.spyOn(global.console, "debug").mockImplementation(() => jest.fn());
+
+    describe("connectToUri", () => {
+        test("creates a valid WebSocket object", () => {
+            const simulator = new RemoteSimulator({});
+            expect(simulator.socketIsValid()).toBe(false);
+
+            simulator.connectToUri(simulator.getIp());
```
toloudis, line 19: is this testing a real live internet connection?

schoinh: Yeah, it's actually connecting to our staging server. Is that ok?

src/constants.ts
toloudis: I'll propose STAGING_SERVER_SETTINGS ?

src/test/RemoteSimulator.test.ts
toloudis: It just means you're no longer testing just simularium-viewer, you're now testing that the staging server and the internet connectivity to it are both alive and well.  There are so many reasons why it could fail.  I would resist doing this in the unit tests here.  Better if you can mock it somehow.  But if others prefer to have this be a true integration test, we can think about it that way.

meganrm: I think it's already mocked in the DummyRemoteSimulator ?

schoinh: Yeah that's a good point. I'll look into using the DummyRemoteSimulator or using Jest to mock some return values like I did in what I called the "key test number 2". I think I would prefer the second option because wouldn't it be better to test the RemoteSimulator class (even if some of its methods are mocked) rather than the DummyRemoteSimulator?

toloudis: I tend to agree, what you really want to mock is the back end, and not the RemoteSimulator itself.  Which i suppose is what the DummyRemoteSimulator is supposed to be trying to do, but it's extending RemoteSimulator.

toloudis:
Looking at this code, it strikes me that a good maintenance task on this repo would be to split apart the "viewing" code from the "streaming/downloading" code into separate subdirectories and have them talk to each other through a well thought out interface.  Something not to change any functionality but to start to decouple the two concerns a bit more.  Ideally this would organize the code better but have no functional or performance impact.

src/simularium/RemoteSimulator.ts
schoinh: That's what I thought/expected too at first, but if you look at line 296-297 it won't let you get to a connection retry (in line 299) until `timeWaited` is at least at `MAX_WAIT_TIME`, which is what (along with some console logs) led to my interpretation. But this function really needs a test so that we know for sure how long we're making the user wait and how many retries we're allowing. I tried to break it out into a class method so I can test it but it turned into a tangled mess, so I reverted it 😅 But seeing your comment makes me want to try again, so I'll give it another go at testing this recursion.

toloudis: I see what you mean.  Thanks.  It's basically just polling on a shorter interval (`timeout`) but allowing for a long interval (MAX_WAIT_TIME) for the total time to connect.

meganrm: Yeah, I think the idea overall was that if it had failed to connect after waiting a while, waiting a second time wouldn't help much, the problem isn't a slow connection, but no connection. 

Have you tried to factor out `checkConnection` and pass in the two counters?

schoinh:
I need to do a bit more cleanup

src/simularium/ISimulator.ts
```diff
@@ -18,7 +18,7 @@ export interface ISimulator {
     /**
      * Connect
      * */
-    connectToUri(uri: string): void;
+    createWebSocket(uri: string): void;
```
toloudis, line 21: I hope this can be removed from the ISimulator interface and it's really just private to RemoteSimulator.

schoinh:
Changes since @meganrm and @toloudis 's review: https://github.com/allen-cell-animated/simularium-viewer/pull/173/files/52903c395dae92977e6057e67feede30e32aa322..db1f0c55a73ffa9ec2e39f6ec6b5df2b0b71fbb7

- Factored out `checkConnection` as a class method that's more easily testable
- Used mock return values in all tests instead of real ones so that the tests aren't affected by any network/server issues
- I reverted the creation of the `TEST_CONNECTION_SETTINGS` constant because it was no longer needed in the tests.
- Renamed `connectToUri` method as `createWebSocket` because I think this is less confusing / more accurate
- I had to add a babel plugin `@babel/plugin-transform-runtime`. Something about me turning `waitForWebSocket` into an async class method was giving me a "regeneratorRuntime is not defined" error, so this was necessary to fix it. More info: https://babeljs.io/docs/en/babel-plugin-transform-runtime#regenerator

src/simularium/RemoteSimulator.ts
```diff
@@ -255,56 +266,61 @@ export class RemoteSimulator implements ISimulator {
...
-                    );
-                }
-            };
-            const TOTAL_WAIT_SECONDS = 2;
-            let secondsWaited = 0;
-            const MAX_CONNECTION_TRIES = 2;
-            let connectionTries = 1;
-            return startPromise.then(handleReturn);
-        });
+        this.createWebSocket(address);
```
toloudis, line 316: Is it possible to move this line inside of this.checkConnection? (is checkConnection called by anything else?)

src/simularium/RemoteSimulator.ts
toloudis: I wonder if there's a way to let a thing stay private but make it public for unit tests

src/simularium/RemoteSimulator.ts
schoinh: That makes sense @meganrm . I factored out `checkConnection` but instead of passing in the counters as params I turned them into class properties. Let me know if there's a good reason to make them function params instead.

meganrm (approved):

src/simularium/RemoteSimulator.ts
schoinh: I couldn't find a way to do that, but I modified the test so that we don't have to access this method.

src/simularium/ISimulator.ts
schoinh: Removed from ISimulator, ClientSimulator, and LocalSimulator and seems like everything still works fine.

src/simularium/RemoteSimulator.ts
toloudis: Cool.  You can do it in C++ by using the "friend" keyword which allows other classes access to my class's private members.

schoinh: Whaaaaat that's pretty cool.

toloudis (approved):

PR #174, opened 2021-08-20
Feature/custom geo

### LINKED ISSUE #149: Load geometry for data imported locally or from URL
## Use Case
It would be nice if users could set geometry files, either OBJ mesh files or PDB files/ids, and have them used for rendering agents. 

## Acceptance Criteria
Users can specify mesh or PDBs in their .simularium files, somehow provide those files to the viewer, and see their agents rendering with the correct mesh or PDB by their type.

## Details
Currently this works for streaming files (using an additional metadata file, vis_data.json) but not for files loaded by drag/drop or from a URL in the URL params.

Need to figure out how to let users provide multiple files or data. Could allow users to import zipped (or unzipped?) directories of files? Or follow URLs in trajectoryInfo from the .simularium file?

meganrm:
Problem
=======
This is the first step in converting to file format version 3 geo handling
 resolves #149 

Solution
========
I removed all the geometryFile request and handling. Thanks to @toloudis for some good discussion on the data processing. And added the data handling to the versionHandler to make sure all the new fields are present. 

## Type of change

Change summary:
---------------
* I started working on loading local files, I move that to another branch, but that's some of the early changes on here 
* refactor the path that used to use geometryFile to use the trajectoryFileInfo
* Add tests and typing
* Add error handling for user feedback

Steps to Verify:
----------------
1.  The existing example should work, although if they have geo, they'll only show as spheres now. 
2. I have some test files, I'll try to post them here if they arent too big, otherwise I'll share them on the dev channel. 

-----------------------
1. [src/simularium/VisGeometry.ts](https://github.com/allen-cell-animated/simularium-viewer/compare/feature/custom-geo?expand=1#diff-3a1fefcb5547dbeb43717d5829e7051157f9afe5eb71a06358a1d2e111adeef1)
2. [src/simularium/versionHandlers.ts](https://github.com/allen-cell-animated/simularium-viewer/compare/feature/custom-geo?expand=1#diff-c11b63de9abee754703762fead84e81e6a209e68bf3df4fb695e7483aa8d2ee7)

Thanks for contributing!


meganrm:
[actin-version3-pdb.simularium.zip](https://github.com/allen-cell-animated/simularium-viewer/files/7023863/actin-version3-pdb.simularium.zip)
This should show actin pdb, but error out on apr requests, and show spheres instead

meganrm:
[actin-version3-obj.simularium.zip](https://github.com/allen-cell-animated/simularium-viewer/files/7023873/actin-version3-obj.simularium.zip)
This will load meshes, but again, error out on apr

examples/Viewer.tsx
```diff
@@ -10,7 +10,7 @@ import SimulariumViewer, {
     SimulariumFileFormat,
 } from "../src";
 import "./style.css";
-import { isEqual } from "lodash";
+import { isEqual, findIndex} from "lodash";
```
toloudis, line 13: formatting : spaces at beginning/end of brackets

src/controller/index.ts
```diff
@@ -306,7 +272,7 @@ export default class SimulariumController {
                     connectionParams.clientSimulatorParams,
                     connectionParams.simulariumFile
                 );
-                this.networkEnabled = true;
+                this.networkEnabled = true; // This confuses me, because local files also go through this code path
```
toloudis, line 275: good point, this could be targeted for removal soon. It might be serving as "are we using a RemoteSimulator or not" but I 'm not sure.

src/simularium/versionHandlers.ts
```diff
@@ -1,27 +1,94 @@
...
+    SphereDisplayType,
     TrajectoryFileInfo,
     TrajectoryFileInfoAny,
     TrajectoryFileInfoV1,
 } from "./types";
+import { AgentDisplayDataWithGeometry } from "./VisGeometry";
 
+// the data may come in missing any of these values
+export interface AgentTypeVisDataPreProcessing {
+    displayType?:
```
toloudis, line 19: I wonder if "FIBER" could be one of these types... not for this PR...

src/simularium/versionHandlers.ts
```diff
@@ -1,27 +1,94 @@
...
+): { [key: number]: AgentDisplayDataWithGeometry } => {
+    return mapValues(typeMapping, (value: AgentDisplayDataPreProcessing) => {
+        let geometry = {};
+        if (value.geometry) {
+            let url = value.geometry.url || "";
+            let { displayType } = value.geometry;
+            if (!displayType) {
+                // we're relying on the data to have a displayType to tell us what sort of data the url is pointing at
+                // if the user fails to provide the displayType, we'll default to loading a sphere, and clear out the url
+                url = "";
```
toloudis, line 54: is this a good place to console.log?  since you are destroying any provided url, it might be a good chance to print the url here at this time.

src/simularium/versionHandlers.ts
```diff
@@ -1,27 +1,94 @@
...
+        ...msg,
+        typeMapping: sanitizeAgentMapGeometryData(msg.typeMapping),
+        version: LATEST_VERSION,
+    };
 
     switch (msg.version) {
         case LATEST_VERSION:
             break;
+        case 2:
+            break;
```
toloudis, line 91: is there any distinction between v2 and v3 now?  I didn't see any definitions for TrajectoryFileInfoV2 vs V3

src/simularium/versionHandlers.ts
```diff
@@ -1,27 +1,94 @@
...
+    CubeDisplayType,
+    GizmoDisplayType,
+    ObjDisplayType,
+    PdbDisplayType,
+    SphereDisplayType,
     TrajectoryFileInfo,
     TrajectoryFileInfoAny,
     TrajectoryFileInfoV1,
 } from "./types";
+import { AgentDisplayDataWithGeometry } from "./VisGeometry";
```
toloudis, line 15: this file shouldn't depend on VisGeometry, i'd define it here and import it the other way

toloudis: or in `types`

src/simularium/versionHandlers.ts
meganrm: currently no, but I was going to bring in the other changes for v3 in a different PR. So far all the v3 changes are optional.

src/simularium/VisGeometry.ts
```diff
@@ -86,6 +91,11 @@ interface AgentTypeGeometry {
     pdbName: string;
 }
 
+export interface AgentDisplayDataWithGeometry {
```
toloudis, line 94: move def to a lower level module

src/simularium/VisGeometry.ts
```diff
@@ -770,57 +779,79 @@ class VisGeometry {
...
-                } else {
-                    console.error(
-                        "Mesh loaded but could not find instanceable geometry in it"
-                    );
-                }
-                if (!object.name) {
-                    object.name = meshName;
-                }
-                this.onNewRuntimeGeometryType(meshName);
+    handleObjResponse(meshName: string, object: Object3D): void {
```
toloudis, line 784: mark as private?

src/simularium/VisGeometry.ts
```diff
@@ -1105,34 +1135,28 @@ class VisGeometry {
...
+        displayType: string,
+        url?: string,
+        color?: string
     ): void {
-        this.logger.debug(`Mesh for id ${id} set to '${meshName}'`);
-        this.logger.debug(`PDB for id ${id} set to '${pdbName}'`);
+        this.logger.debug(`Geo for id ${id} set to '${url}'`);
         const unassignedName = `${VisAgent.UNASSIGNED_NAME_PREFIX}-${id}`;
+        const isMesh = displayType === "OBJ";
+        const isPBD = displayType === "PDB";
```
toloudis, line 1145: isPDB

src/simularium/VisGeometry.ts
```diff
@@ -1105,34 +1135,28 @@ class VisGeometry {
...
+        if (!url) {
+            // displayType not either pdb or obj, will show a sphere
+            // TODO: handle CUBE, GIZMO etc
+            return;
+        }
         this.visGeomMap.set(id, {
-            meshName: meshName || DEFAULT_MESH_NAME,
-            pdbName: pdbName || "",
+            meshName: isMesh ? url : DEFAULT_MESH_NAME,
+            pdbName: isPBD ? url : "",
```
toloudis, line 1154: if mesh and pdb are mutually exclusive now, it seems to me the visGeomMap could just have one single object of displayType and url.  Not necessarily for this pr but I think it would work.

src/simularium/VisGeometry.ts
```diff
@@ -1202,82 +1226,23 @@ class VisGeometry {
...
-
-        Object.keys(jsonData).forEach((id) => {
-            const entry: AgentTypeVisData = jsonData[id];
-            if (id === "size") {
-                console.log("WARNING: Ignoring deprecated bounding box data");
-            } else {
-                // mesh name is entry.mesh
-                // pdb name is entry.pdb
-                this.mapIdToGeom(Number(id), entry.mesh, entry.pdb, assetPath);
-                this.setScaleForId(Number(id), entry.scale);
```
toloudis, line 1274: did we lose the `scale` value?

src/simularium/versionHandlers.ts
meganrm: you mean to display to the user?

src/simularium/VisGeometry.ts
meganrm: yeah, I think the whole, load, keep track of requests, and save the geo data could be its own class, and yeah, I think this sort of change would belong in that set

src/simularium/VisGeometry.ts
meganrm: Yeah, I think it's been gone since version 2? I'm going based on the doc. @blairlyons ?

src/simularium/versionHandlers.ts
toloudis: yes, to the user or to the programmer who is wondering why the heck the geometry looks wrong

meganrm: I added a response if the url gets cleared out. Also based on the acceptance criteria, I added another case where the geo defaults to spheres, and alerts the user about it.

src/simularium/versionHandlers.ts
```diff
@@ -1,27 +1,139 @@
...
+    typeMapping: EncodedTypeMappingPreProcessing,
+    onError?: (msg: string) => void
+): { [key: number]: AgentDisplayDataWithGeometry } => {
+    return mapValues(
+        typeMapping,
+        (value: AgentDisplayDataPreProcessing, key) => {
+            let geometry = {};
+            if (value.geometry) {
+                let url = value.geometry.url || "";
+                let { displayType } = value.geometry;
```
toloudis, line 70: is it possible to get an unrecognized displayType here?

toloudis (approved):
Code is looking good.
Only remaining concern is not having enough up-to-date test data to test changing trajectories, cancelling download requests etc because those are where the difficult bugs may come up.

src/test/versionHandlers.test.ts
```diff
@@ -91,7 +101,161 @@ const v2Data = {
...
+    });
+    describe("makeUrlErrorMessage", () => {
+        test("it will create a message for the user if the url is getting set to an empty string", () => {
+            const key = "1";
+            const url = "url-to-geo.com";
+            const message = makeMissingDisplayTypeErrorMessage(key, url);
+            expect(message).toEqual(
+                `Missing typeMapping[${key}].geometry.displayType, so we couldn't request ${url}. Geometry will default to spheres`
+            );
+        });
```
schoinh, line 212: Should these 2 tests be under `describe("makeMissingDisplayTypeErrorMessage")`?

src/test/versionHandlers.test.ts
```diff
@@ -91,7 +101,161 @@ const v2Data = {
...
+
+const v3Data = {
+    ...v2Data,
+    typeMapping: typeMappingWithDefaultGeo,
+    version: 3,
+};
+
 describe("Version handlers", () => {
+    describe("makeUrlErrorMessage", () => {
+        test("it will create a message for the user if there is not displayType", () => {
```
schoinh, line 195: This is a little confusing because the tested function doesn't actually know about the existence of `displayType`. I think it would be better to say something like "it will create a message for the user given an empty string for the url" and then "it will create a message for the user given a non-empty string for the url" for the following test

src/viewport/index.tsx
```diff
@@ -540,32 +528,14 @@ class Viewport extends React.Component<ViewportProps, ViewportState> {
...
-                });
-
+                this.visGeometry.render(totalElapsedTime);
+                this.lastRenderTime = Date.now();
+                this.lastRenderedAgentTime = -1;
                 simulariumController.markFileChangeAsHandled();
+           
+                this.animationRequestID = requestAnimationFrame(
+                    this.animate
+                );
```
schoinh, line 538: Is it ok to call this after line 534? (since it was being called before, it looks like?)

src/simularium/types.ts
```diff
@@ -55,11 +55,34 @@ interface Plot {
...
+    name: string;
+    geometry: AgentTypeVisData;
+}
+
 export interface AgentDisplayData {
     name: string;
-    pdb?: string;
-    mesh?: string;
+    geometry?: AgentTypeVisData;
 }
```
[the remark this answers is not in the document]
meganrm: It's so I could know geometry would be defined. I guess I could have one interface extend the other one? I haven't tried that

src/simularium/VisGeometry.ts
```diff
@@ -1105,34 +1131,28 @@ class VisGeometry {
...
+        url?: string,
+        color?: string
     ): void {
-        this.logger.debug(`Mesh for id ${id} set to '${meshName}'`);
-        this.logger.debug(`PDB for id ${id} set to '${pdbName}'`);
+        this.logger.debug(`Geo for id ${id} set to '${url}'`);
         const unassignedName = `${VisAgent.UNASSIGNED_NAME_PREFIX}-${id}`;
+        const isMesh = displayType === "OBJ";
+        const isPDB = displayType === "PDB";
+        console.log(color); // TODO: handle color
```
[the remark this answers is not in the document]
meganrm: I left this in so the linter wouldn't be mad about not using color.... since it's going to be handled soon

src/simularium/versionHandlers.ts
```diff
@@ -1,27 +1,139 @@
...
+                    if (onError) {
+                        onError(message);
+                    } else {
+                        console.log(message);
+                    }
+                    url = "";
+                    displayType = "SPHERE";
+                } else if (
+                    (!url && displayType === "PDB") ||
+                    displayType === "OBJ"
```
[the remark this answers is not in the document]
meganrm: yeah! thanks, I'm glad you caught this

src/simularium/types.ts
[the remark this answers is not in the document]
schoinh: Okay, yeah I'm not sure what the best way to do this is but I don't feel super strongly about it.

src/simularium/VisGeometry.ts
```diff
@@ -1202,82 +1222,23 @@ class VisGeometry {
...
+                entry.geometry.displayType,
+                entry.geometry.url,
+                entry.geometry.color
+            );
         });
-
-        if (callback) {
-            callback(jsonData);
-        }
+        console.log("AFTER", this.currentSceneAgents);
```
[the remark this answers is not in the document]
meganrm: A little bit? I am curious if this gets called with a non empty array ever, I can change it to a comment and check on the website using dev tools though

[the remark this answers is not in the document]
schoinh: Gotcha. It would be good to change `AFTER` to something more informative if you do leave it in :)

src/simularium/VisGeometry.ts
```diff
@@ -399,6 +402,7 @@ class VisGeometry {
         }
         // Reset then position and orient the camera
         this.resetCamera();
+        this.handleAgentGeometry(trajectoryFileInfo.typeMapping);
```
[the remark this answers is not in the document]
toloudis: it's a shame that `resetCamera` needs a comment that says it is resetting the camera 🤣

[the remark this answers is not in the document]
schoinh: 😆 It looks less ridiculous if you look at the rest of the method... but yeah now that you say it it does seem kinda silly....... @meganrm Feel free to remove it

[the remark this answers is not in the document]
meganrm: since this code needed to be called in order, but didn't look connected, I factored it out into it's own function that handles the camera data

src/viewport/index.tsx
meganrm: yeah, this mark as handled is a little weird here now because before it was waiting for an async event to be resolved, but now it's just being called in line

PR #175, opened 2021-08-22
Feature/multiple render targets

toloudis:
Problem
=======
Suboptimal rendering: geometry was being iterated over 3 times, to fill 3 different GPU buffers.
With recent updates to THREE.js, we can now access "multiple render targets" which allows drawing geometry once and writing outputs to separate buffers from one shader.

Solution
========
Update three js and refactor the shaders and the code that does the drawing to use one pass instead of three.

## Type of change

Steps to Verify:
----------------
This should not change any functionality.  Rendering code is simplified and cleaner, and should have a bit less cpu overhead to issue draw calls. There should be no loss of performance and we expect a gain overall.

Key files:
---------
the ones in src/simularium/rendering.  GBufferPass is the one that is affected.  GBufferPass and the shader files are the big changes.
There is a bonus bug fix in VisGeometry for loading pdb geometry and some minor tweaks to less significant code (e.g. the local client simulators, which are just live test cases really).


package.json
```diff
@@ -78,6 +81,10 @@
         "webpack-dev-server": "^3.11.2",
         "worker-loader": "^2.0.0"
     },
+    "peerDependencies": {
+        "react": "^16.9.0",
+        "react-dom": "^16.9.0"
+    },
```
toloudis, line 87: following recommended best practices for including react as peerDependency and devDependency, not a "dependency"

src/simularium/rendering/GBufferPass.ts
```diff
@@ -141,93 +129,13 @@ class GBufferPass {
...
-        if (DO_PDB) {
-            this.agentPDBGroup.visible = true;
-            this.instancedMeshGroup.visible = false;
-
-            setSceneRenderPass(scene, this.pdbGbufferMaterials, renderPass);
-            renderer.render(scene, camera);
-            renderer.autoClear = false;
-            scene.overrideMaterial = null;
-        }
-
```
toloudis, line 230: major change here: all this code is removed because we do drawing in a single pass now

src/simularium/rendering/HitTestHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+    ): Float32Array {
+        const pixel = new Float32Array(4).fill(-1);
+        // (typeId), (instanceId), fragViewPos.z, fragPosDepth;
+
+        this.hitTestMesh.material.uniforms.objectIdTexture.value = idBuffer;
+        this.hitTestMesh.material.uniforms.pixel.value = new Vector2(x, y);
+        renderer.setRenderTarget(this.hitTestBuffer);
+        renderer.render(this.hitTestScene, this.hitTestCamera);
+        renderer.setRenderTarget(null);
+        renderer.readRenderTargetPixels(this.hitTestBuffer, 0, 0, 1, 1, pixel);
```
toloudis, line 80: this implementation (how we extract the object id from a hit test read) had to change because of the multiple render target changes.  I refactored it into this helper class do keep it better encapsulated.

src/simularium/rendering/InstancedFiberShader.ts
```diff
@@ -371,39 +364,21 @@ void main()	{
...
 
-const positionShader = `
-precision highp float;
-
-in vec3 IN_viewPos;
-out vec4 fragColor;
-void main()	{
-    
-    vec3 fragViewPos = IN_viewPos;
-    fragColor = vec4(fragViewPos.x, fragViewPos.y, fragViewPos.z, 1.0);
```
toloudis, line 399: where there were 3 fragment shaders before, now there is a single one that writes 3 different outputs

src/test/AgentSimController.test.ts
```diff
@@ -1,3 +1,4 @@
+import "regenerator-runtime/runtime.js";
```
schoinh, line 1: Did you have to import this because you got a "regeneratorRuntime is not defined" error? If so, I added a Babel plugin that fixes that in the PR I just merged, so these imports may not be necessary anymore?

toloudis: Yup.

src/simularium/rendering/SimulariumRenderer.ts
```diff
@@ -364,43 +354,40 @@ class SimulariumRenderer {
 
         // 1 draw molecules into G buffers
         // TODO: MRT
```
schoinh, line 356: Remove TODO comment?

src/simularium/rendering/SimulariumRenderer.ts
```diff
@@ -364,43 +354,40 @@ class SimulariumRenderer {
...
-            camera,
-            this.colorBuffer,
-            this.normalBuffer,
-            this.positionBuffer
-        );
+        this.gbufferPass.render(renderer, scene, camera, this.gbuffer);
+
+        const POSITIONBUFFER = 2;
+        const NORMALBUFFER = 1;
+        const AGENTBUFFER = 0;
```
schoinh, line 361: Move these up as a global constant and use in lines 125-127, and in line 250?

src/simularium/rendering/HitTestHelper.ts
schoinh: I think it would be awesome if you added a brief description of what this class (and the `hitTest` method) does as a code comment!

schoinh (approved):

meganrm (approved):

PR #182, opened 2021-08-26
Feature/local custom geo

### LINKED ISSUE #177: Load folders of geometry files
## Use Case
It would be nice if users could drag a folder, rather than just a collection of files, onto the viewer to load local geometry files have them used for rendering agents.

## Acceptance Criteria
A user can drag a folder from their file system to the browser and the viewer uses the geometry files within to render the agents.

## Details
Folder should be able to contain both the .simularium file and any accompanying geometry files.

If not too complex, would be nice to handle any combination of folders and multiple selected files, e.g. the user could drag in a collection of files that includes a folder and all the files could be used.

meganrm:
Problem
=======
resolves #177 

Solution
========
If additional files are dropped along with the simularium file, cache the contents, and then look in the cache for data before making requests for data hosted remotely. 

## Type of change

Change summary:
---------------
* Pass additional files in as a mapping of file names to text strings
* add additional registries to store cached data 
* check cache before loading from url 

Steps to Verify:
----------------
1. npm start
2. unzip the file I'll attach 
3. drop the contents of the folder (grab the individual files) and drop them all together
4. Should load most of the geo except for the one that says "should fail"

Screenshots (optional):
-----------------------
Some of the geometry was loaded via url, some was dropped onto the viewer, and one is missing on purpose, so shows spheres. 
Screen Shot 2021-08-25 at 3 05 55 PM

Thanks for contributing!


src/simularium/VisGeometry.ts
```diff
@@ -88,8 +91,18 @@ function removeByName(group: Group, name: string): void {
...
 interface AgentTypeGeometry {
-    meshName: string;
-    pdbName: string;
+    name: string;
+    displayType: GeometryDisplayType;
+}
+
+interface PdbGeometry {
+    geometry: PDBModel;
+    displayType: PdbDisplayType;
```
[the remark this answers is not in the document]
toloudis: see #184, I at least refactored these two types into one thing kinda sorta

src/simularium/VisGeometry.ts
```diff
@@ -110,7 +123,11 @@ interface PathData {
     line: LineSegments;
 }
 
-type AgentTypeVisDataMap = Map<string, AgentTypeVisData>;
+interface CachedFileData {
+    displayType: ObjDisplayType | PdbDisplayType;
```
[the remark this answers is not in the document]
toloudis: see #184

src/simularium/VisGeometry.ts
```diff
@@ -746,6 +755,13 @@ class VisGeometry {
                             ${url}.
                             Showing spheres for this geometry instead`
                     );
+                    const ids = this.getAllTypeIdsForGeometryName(url);
+                    forEach(ids, (id) => {
+                        this.visGeomMap.set(id, {
+                            name: DEFAULT_MESH_NAME,
+                            displayType: "OBJ",
```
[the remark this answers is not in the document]
toloudis: see #184

src/simularium/VisGeometry.ts
```diff
@@ -725,7 +720,18 @@ class VisGeometry {
         this.initCameraPosition = this.camera.position.clone();
     }
 
-    public loadPdb(url: string): void {
+    public loadPdb(url: string, pdbmodel: PDBModel): void {
+        this.logger.debug("Finished loading pdb: ", url);
```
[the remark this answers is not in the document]
meganrm: I moved it into the download function

src/simularium/VisGeometry.ts
```diff
@@ -1166,64 +1232,24 @@ class VisGeometry {
         }
     }
 
-    public getGeomFromId(id: number): Object3D | null {
```
meganrm, line 1169: these functions were never called

EricJIsaac (approved):

src/simularium/VisGeometry.ts
```diff
@@ -843,13 +873,45 @@ class VisGeometry {
...
-        url: string,
+        urlOrPath: string,
         registry: Map<string | number, PDBModel | MeshLoadRequest>,
-        loadFunctionName: string
+        displayType: GeometryDisplayType
     ) {
-        if (!registry.has(url) && !this.geoLoadAttempted.get(url)) {
-            this.geoLoadAttempted.set(url, true);
-            return this[loadFunctionName](url);
+        if (this.localGeoFiles.has(urlOrPath)) {
```
toloudis, line 880: should this line check in the registry also, like line 898?  if item with the given key is in the registry we can skip all this?

meganrm: Since I delete it from the cache on line 896, after it's been loaded in the registry, the assumption is that it's only in the cache if it's never been loaded before.

toloudis (approved):

src/simularium/VisGeometry.ts
```diff
@@ -1129,35 +1192,38 @@ class VisGeometry {
...
-        if (isMesh) {
-            this.attemptToLoadGeometry(url, this.meshRegistry, "loadObj");
-        } else if (isPDB) {
-            this.attemptToLoadGeometry(url, this.pdbRegistry, "loadPdb");
+        if (isMesh || isPDB) {
+            this.attemptToLoadGeometry(
+                urlOrPath,
+                this.meshRegistry,
+                displayType
+            );
```
meganrm, line 1226: @toloudis, i've been meaning to ask about this: In lines 1227 -1231, are these a remnant of assuming if a model is a pdb geo, it's always going to be a pdb model, and of the separate registries? my thinking is that anything that fails will default to a SPHERE mesh instead of a single atom pdb.

toloudis: umm.. yes? I think both things might be true but I'm sure we want to use sphere mesh for the default/error fallback rather than single atom pdb.  At least for now.

toloudis: it's going to be a bit weird having all this `PDBModel | MeshLoadRequest` all over the place and we might need an even more generic "delayed instanced object container" kind of thing that hides that.

meganrm: yeah, I guess that's my actual question. If a request fails, should we revert the type to the default geometry (same as if they didn't supply any geometry info at all), or should we try to be consistent with the displayType provided and keep this single atom pdb as the fallback to a failed pdb request?

toloudis: I'm not sure if this single atom pdb code is ever hit

toloudis: took a bit more time to read and think about this, and I think this is dead code now.  Everything having to do with UNASSIGNED_NAME_PREFIX (including in VisAgent.ts).

meganrm: thanks

PR #184, opened 2021-08-28
cleanup, use hardcoded strings less

toloudis:
@meganrm here are the changes I was talking about.  I did have to leave in one extra "displayType" variable in the AgentGeometry type which I think shouldn't really be there, but maybe the next refactor will make it easier to get rid of that!
Note these changes go as part of #182

### DIFF
--- a/src/simularium/VisGeometry.ts
+++ b/src/simularium/VisGeometry.ts
@@ -44,8 +44,6 @@ import {
     CameraSpec,
     EncodedTypeMapping,
     AgentDisplayDataWithGeometry,
-    ObjDisplayType,
-    PdbDisplayType,
     GeometryDisplayType,
 } from "./types";
 import { AgentData } from "./VisData";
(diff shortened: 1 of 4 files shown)


toloudis:
The names of the things inside the enum should really now be something like:
enum GeometryDisplayType {
PDB="PDB",
OBJ="OBJ",
SPHERE="SPHERE",
etc
}
but I didn't want to do that much renaming because it would obfuscate the actual important changes.

meganrm:
yeah, thanks, I think this makes sense overall... and I do agree with your second comment that the enum is redundantly named. Is there a reason you prefer the enum/string constant to just letting typeScript tell you if you've spelled a variable wrong?  I think `GeometryDisplayType.PDB` is probably slightly clearer than just "PDB" but i was wondering if there was a reason for the change other than getting to have a clear variable name

meganrm:
I'm just going to go ahead and merge this and then update the enum on my open PR

toloudis:
>  i was wondering if there was a reason for the change other than getting to have a clear variable name

Clear variable name for readability, lets the symbol name be different from the string value (so you could make it more desciptive), ease of extension (just add new values in one place, never need to import single values), can validate against existing values more easily, and lastly, I want to think of the strings as VALUES and not TYPES in the programming sense.  "PDB" is not a type, it's a value. the TYPE is GeometryDisplayType and it's allowed only certain values, which is a good use for an enum.

PR #185, opened 2021-08-31
Feature/geometry store

### LINKED ISSUE #181: create a module for requesting and storing geometry files
## What needs to happen?
Pull out the mesh and pdb registry handling from visGeometry. 

## Why should we do this?
visGeometry is getting really large, and this is a chunk of work that could be separated out. 

## When does this need to get done?

would be good to do with the custom geo changes since it's expanding the visGeometry file

meganrm:
Problem
=======
Factors out geometry handling into it's own module 
resolves #181 

Solution
========
What I/we did to solve this problem

with @pairperson1

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


src/simularium/PDBModel.ts
```diff
@@ -74,24 +73,24 @@ class PDBModel {
     private bounds: Box3;
     // cancelled means we have abandoned this pdb
     // and if it is still initializing, it should be dropped/ignored as soon as processing is done
-    private cancelled: boolean;
+    private _cancelled: boolean;
```
meganrm, line 76: made this change so I could call `geometry.cancelled` for either geometry type and it would work.

src/simularium/VisGeometry/GeometryStore.ts
```diff
@@ -0,0 +1,358 @@
...
+                ),
+            },
+        });
+    }
+
+    public get registry(): Registry {
+        return this._registry;
+    }
+
+    public updateMeshes(): void {
```
meganrm, line 61: I think it's also reasonable for VisGeometry to do this, curious what people's opinions are

src/simularium/VisGeometry.ts
```diff
@@ -1489,6 +1231,93 @@ class VisGeometry {
         return agent;
     }
 
+    private renderPdb(typeId: number, visAgent: VisAgent, pdbEntry: PDBModel) {
```
meganrm, line 1234: these are unchanged, I just wanted to be able to see what was happening in updateScene more clearly

src/simularium/PDBModel.ts
```diff
@@ -108,27 +107,6 @@ class PDBModel {
...
-                    );
-                }
-                return response.text();
-            })
-            .then((data) => {
-                if (this.cancelled) {
-                    return Promise.reject(REASON_CANCELLED);
-                }
-                // note pdb atom coordinates are in angstroms
-                // 1 nm is 10 angstroms
```
toloudis, line 127: can you move this comment into the parsePdb call? I don't want to lose that info

src/simularium/VisGeometry/GeometryStore.ts
```diff
@@ -0,0 +1,352 @@
...
+        // note that this leaves cancelled things in the registries.
+        // This should be called before the registries are cleared and probably
+        // only makes sense to do if they are indeed about to be cleared.
+
+        // don't process any queued requests
+        TaskQueue.stopAll();
+        // signal to cancel any pending geometries
+        this._registry.forEach((value) => {
+            value.geometry.cancelled = true;
+        });
```
meganrm, line 87: dan, shouldn't this clear out `_geoLoadAttempted` too?

meganrm:
Sorry this took so long to finally be submitted, there was an error switching between my files that I finally fixed.

meganrm:
[pdb-test.zip](https://github.com/allen-cell-animated/simularium-viewer/files/7156398/pdb-test.zip)

meganrm:
> I still need to look more carefully, but right off the bat I noticed that we now have a `src/simularium/VisGeometry.ts` and a `src/simularium/VisGeometry` (folder) that contains the geometry store. Is that what we want to do? Would it not make sense to move VisGeometry.ts into `src/simularium/VisGeometry` as index.ts?

Yes, sorry I meant to make a note of this in the PR: I did this so the diff is readable. Once it's approved I'll do a different PR that renames the file.

src/simularium/VisGeometry/GeometryStore.ts
toloudis: I don't think it really matters.  The main point here is that any async processing is either shut down (the task queue) or when results arrive they can be discarded (the cancelled=true).

toloudis:
> 
> Yes, sorry I meant to make a note of this in the PR: I did this so the diff is readable. Once it's approved I'll do a different PR that renames the file.

We also need to think about separating "viewer" from "data connection and relay" -- all the Simulators and Controller stuff could be in a separate subdirectory from the Viewer, VisGeometry and rendering stuff.

src/simularium/VisGeometry/GeometryStore.ts
```diff
@@ -0,0 +1,352 @@
...
+    GeometryStoreLoadResponse,
+    MeshGeometry,
+    MeshLoadRequest,
+} from "./types";
+
+export const DEFAULT_MESH_NAME = "SPHERE";
+
+type Registry = Map<string, AgentGeometry>;
+
+class GeometryStore {
```
schoinh, line 25: Can we have documented somewhere (in a readme, or as code comments for class definitions) the intended purpose/scope of `GeometryStore` and `VisGeometry`? That would be sooo helpful. (I would love them for `VisAgent` and `VisData` too but I don't mean in this PR 🙂)

src/simularium/VisGeometry.ts
```diff
@@ -1259,24 +963,62 @@ class VisGeometry {
...
+                        geometry,
+                        errorMessage,
+                    } = returned;
+                    const newDisplayType = returnedDisplayType || displayType;
+                    this.onNewRuntimeGeometryType(
+                        lookupKey,
+                        newDisplayType,
+                        geometry
+                    );
+                    // handle additional async update to LOD for pdbs
```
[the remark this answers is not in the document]
toloudis: Level Of Detail.  Each PDB asynchronously generates several reduced size representations to be drawn at different distances from the camera.

src/simularium/VisGeometry/GeometryStore.ts
```diff
@@ -0,0 +1,352 @@
...
+                    return Promise.resolve(undefined);
+                }
+                pdbModel.parsePDBData(data);
+                const pdbEntry = this._registry.get(url);
+                if (pdbEntry && pdbEntry.geometry === pdbModel) {
+                    this.mlogger.info("Finished downloading pdb: ", url);
+
+                    return pdbModel;
+                } else {
+                    // TODO: what should happen here?
```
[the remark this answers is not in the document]
meganrm: Yeah, this wasn't my note. I can probably just get rid of the else block all together, but I wasn't sure if this TODO was something that would remind someone (probably @toloudis ) of work he wanted to do

src/test/GeometryStore.test.ts
```diff
@@ -0,0 +1,176 @@
...
+            const returned = await store.mapKeyToGeom(1, {
+                displayType: GeometryDisplayType.SPHERE,
+                url: "",
+                color: "",
+            });
+            const registry = store.registry;
+
+            expect(returned).toBeTruthy();
+            const savedMesh = registry.get("1-SPHERE");
+            if (returned && savedMesh) {
```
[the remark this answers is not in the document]
meganrm: I will aways be, this was just a typeScript thing to keep it from telling me it could possibly be undefined, I could put a note.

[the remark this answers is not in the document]
schoinh: Maybe a `expect savedMesh to be truthy` statement before the if statement would do the trick? I think I saw something like that in another test here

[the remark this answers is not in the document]
meganrm: No, typeScript doesn't care if the test passed. Same with `returned` here. I shouldn't need this if statement at all

src/simularium/VisGeometry/GeometryStore.ts
meganrm: I added some comments to the file itself. Maybe the planned file reorg can also add a README to this folder?

schoinh: That sounds good. I don't see it in GeometryStore.ts though, did you push it?

src/simularium/VisGeometry.ts
```diff
@@ -1026,9 +782,15 @@ class VisGeometry {
 
             // collect up the meshes that have > 0 instances
             const meshTypes: InstancedMesh[] = [];
-            for (const entry of this.meshRegistry.values()) {
-                meshTypes.push(entry.instances);
-                this.instancedMeshGroup.add(entry.instances.getMesh());
+            for (const entry of this.geometryStore.registry.values()) {
+                const { displayType } = entry;
+                if (displayType !== GeometryDisplayType.PDB) {
```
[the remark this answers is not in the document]
meganrm: only Meshes, ie not PDB files, have geometry.instances.  Before we had two seperate registries, one for Meshes and one for PDBs with the idea we might have both a mesh representation and a pdb for one agent. But with the new v3 file format, we're really saying that there is only one geometry per agent type, so the duplicate registry seemed unnecessary. However, in the few cases where we have to do something only to one geometry type, we need to filter the registry down. I thought about making a `geometryStore.getMeshes` but it would only be for this one use case, and it would then make us loop over the meshes twice.

src/simularium/VisGeometry/GeometryStore.ts
meganrm: Now I did. :)

src/simularium/VisGeometry.ts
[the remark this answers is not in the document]
schoinh: Ah, thanks! 👍🏻 👍🏻

src/simularium/VisGeometry/GeometryStore.ts
schoinh: Beautiful thank you 😍

schoinh (approved):

schoinh:
Am I supposed to see some spheres in this model? I pulled this branch and drag-and-dropped the files from your zip file but I don't think I see any spheres? 🤔 
image

meganrm:
> Am I supposed to see some spheres in this model? I pulled this branch and drag-and-dropped the files from your zip file but I don't think I see any spheres? 🤔
> image
Yeah, they should show up when you play it.

meganrm:
Screen Shot 2021-09-20 at 11 43 21 AM

src/simularium/VisGeometry/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export enum GeometryDisplayType {
+    PDB = "PDB",
+    OBJ = "OBJ",
+    SPHERE = "SPHERE",
+    CUBE = "CUBE",
+    GIZMO = "GIZMO",
+}
+
+export type PrimitiveDisplayType =
```
toloudis, line 19: what is PrimitiveDisplayType used for?

src/simularium/VisGeometry/types.ts
```diff
@@ -0,0 +1,40 @@
...
+    geometry: PDBModel;
+    displayType: GeometryDisplayType.PDB;
+}
+
+export interface MeshGeometry {
+    geometry: MeshLoadRequest;
+    displayType: GeometryDisplayType.OBJ | PrimitiveDisplayType;
+}
+
+export type AgentGeometry = PDBGeometry | MeshGeometry;
```
toloudis, line 34: just looking at this, would it be cleaner if you just had 
```
export interface AgentGeometry {
  geometry: PDBModel | MeshLoadRequest;
  displayType: GeometryDisplayType;
}
```
?

toloudis: and then that starts to look like the GeometryStoreLoadResponse too

src/simularium/VisGeometry/GeometryStore.ts
```diff
@@ -0,0 +1,400 @@
...
+        forEach(assets, (value, key) => {
+            const path = checkAndSanitizePath(key);
+            this._cachedAssets.set(path, value);
+        });
+    }
+
+    private createNewSphereGeometry(meshName: string): MeshLoadRequest {
+        /** create new default geometry */
+        return {
+            mesh: new Mesh(VisAgent.sphereGeometry),
```
toloudis, line 131: I don't think there's much need to have sphereGeometry live in VisAgent anymore.  either now, or as a TODO, especially when we support CUBE etc

src/simularium/VisGeometry/GeometryStore.ts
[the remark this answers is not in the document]
toloudis: it's still a TODO ... I can't even remember exactly how it would get there... setGeometryInRegistry (line 151) seems like it would ensure that the if clause is satisfied.

src/simularium/VisGeometry/types.ts
meganrm: Right, but I want to be able to type a return type as either PDBGeometry or MeshGeometry so that I can access attributes only on PDBModel or MeshLoadRequest without having to do a bunch of additional if statements

src/simularium/VisGeometry.ts
```diff
@@ -1259,24 +961,62 @@ class VisGeometry {
...
-                entry.geometry.color
-            );
+            const { url, displayType } = entry.geometry;
+            const lookupKey = url ? checkAndSanitizePath(url) : displayType;
+            this.visGeomMap.set(Number(id), lookupKey);
+            this.geometryStore
+                .mapKeyToGeom(Number(id), entry.geometry)
+                .then((returned) => {
+                    if (!returned) {
+                        // no new geometry to load
```
toloudis, line 974: when does this happen?

src/simularium/VisGeometry/types.ts
meganrm: For instance: https://github.com/allen-cell-animated/simularium-viewer/pull/185/files/572bcfc73001255d9e1f2bda1c6dd3294d4b5a70#diff-be85a101dc76608957bdc5a6860c22c7dc95827dc01a9974a8cf2980d3ca264fR90. 

I also think it's a more accurate typing since you should never have `displayType === PDB` and `geometry === MeshLoadRequest`

src/simularium/VisGeometry.ts
```diff
@@ -1586,66 +1411,19 @@ class VisGeometry {
...
-                            this.getColorForTypeId(typeId)
-                        );
-                    } else {
-                        if (pdbEntry !== visAgent.pdbModel) {
-                            // race condition? agents arrived after pdb did?
-                            this.resetAgentPDB(visAgent, pdbEntry);
-                        }
-                        visAgent.updatePdbTransform(1.0);
-                    }
+                    this.renderPdb(typeId, visAgent, pdbEntry);
```
toloudis, line 1423: these really should be named "addPdbToDrawList" or something like that.  it's not quite rendering that happens inside there

src/simularium/VisGeometry/types.ts
toloudis: i agree those shouldn't happen but typechecking is not going to stop that from happening ,as it's all coming from runtime data

src/simularium/VisGeometry.ts
meganrm: If the same geometry is assigned to the two different agent ids, so there is no new geometry needed to load.

toloudis (approved):
I think this is a huge step in the right direction and there's more incremental stuff I want to do on top of it.  Some of it mentioned in the various comments.

src/simularium/VisGeometry.ts
toloudis: i see, so returned could be thought of as something like "newGeometry", roughly

meganrm: yeah, I added some comments and renamed it so hopefully this will be clearer

src/simularium/VisGeometry/types.ts
meganrm: Yeah, I just was more thinking of the typing as also a form of documentation, and trying to have it be clear what the expected input data should be, not that it would stop the problem.

toloudis: I might do a PR just to see what the code looks like with this change :)

PR #193, opened 2021-10-01
Feature/move visgeo file

meganrm:
Problem
=======
I had held off moving VisGeometry so that the diff wasn't too complicated when it had a lot of edits

Solution
========
Moved and renamed VisGeometry

## Type of change

* Maintenance

Steps to Verify:
----------------
1. npm start
2. test app should work normally


src/simularium/VisGeometry/index.ts
```diff
@@ -29,32 +29,35 @@ import jsLogger from "js-logger";
...
-import { InstancedMesh } from "./rendering/InstancedMesh";
-import { LegacyRenderer } from "./rendering/LegacyRenderer";
-import GeometryStore, { DEFAULT_MESH_NAME } from "./VisGeometry/GeometryStore";
+} from "../types";
+import { AgentData } from "../VisData";
+
+import SimulariumRenderer from "../rendering/SimulariumRenderer";
+import { InstancedFiberGroup } from "../rendering/InstancedFiber";
+import { InstancedMesh } from "../rendering/InstancedMesh";
+import { LegacyRenderer } from "../rendering/LegacyRenderer";
```
toloudis, line 52: VisGeometry is the only one that imports all these `rendering/` files. It looks odd that they became further removed from VisGeometry.  Do you have any objection to my doing a followup that brings all the visualization code together and starts to separate it from the controller/simulation code?  Or doing it in this PR.

meganrm: you mean just moving the rendering folder into VisGeometry/ too? Or more extensive changes. The former seems fine for this PR, the latter maybe separate PR

toloudis: Yes - please just move rendering/ into VisGeometry and we can do the rest in a separate PR! thank you

meganrm: The `worker/KMeansWorker` imports `rendering/KMeans3d` just FYI

package.json
```diff
@@ -49,6 +49,7 @@
         "@types/react": "^16.14.3",
         "@typescript-eslint/eslint-plugin": "~4.29.0",
         "@typescript-eslint/parser": "~4.29.0",
+        "@webpack-cli/serve": "^1.5.2",
```
toloudis, line 52: can remove?

toloudis (approved):

package.json
meganrm: I needed it to run `npm start`, is it just assumed people already have this installed?

toloudis: I have not needed it yet, so not sure.  I'm surprised there wouldn't also be a package-lock update then too.

schoinh (approved):

PR #194, opened 2021-10-04
move visgeometry-associated files inside of visgeometry

toloudis:
Problem
=======
Bring files into directories where they more logically belong.

Solution
========
Move files.  In particular, PDBModel and VisAgent needed to be under VisGeometry.  Also the web worker was moved into VisGeometry.

## Type of change

Organizing files

### DIFF
--- a/src/simularium/VisGeometry/KMeansWorker.ts
+++ b/src/simularium/VisGeometry/KMeansWorker.ts
@@ -1,7 +1,7 @@
 import "regenerator-runtime/runtime";
 import * as Comlink from "comlink";
 
-import KMeans3d from "../VisGeometry/rendering/KMeans3d";
+import KMeans3d from "./rendering/KMeans3d";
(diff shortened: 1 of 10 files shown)


schoinh (approved):

src/simularium/VisGeometry/KMeansWorker.ts
```diff
@@ -17,7 +17,3 @@ class KMeansWorker {
 export type KMeansWorkerType = typeof KMeansWorker;
 
 Comlink.expose(KMeansWorker);
-
-// I know of no other way to deal with this, see the documentation for webpack's worker-loader.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export default self as any;
```
toloudis, line 23: The above change is a bonus that should have been removed when updating webpack a few weeks ago.

PR #208, opened 2021-10-18
Fix/color types

meganrm:
Problem
=======
The viewer was throwing a typescript error if the default color list was a list of strings when I was trying to install it in the website. 

Solution
========
Typescript was inferring the type of the defaultProp, agentColors, which currently is an array of numbers, so I told it that it was actually either an array of numbers or an array of strings. I also changed the typing from `(string | number)[]` to `(string[] | number[])` partially because I thought it might have fixed the problem (it didn't), but also because it seems more accurate, but I don't feel strongly about it and can change it back. 

I didn't notice this typescript error in our example app because it has had two other typescript errors we've been ignoring for a while, so I fixed those too, so any future issues will be more apparent. 

## Type of change

Change summary:
---------------
* Fixed existing typescript errors in the example app
* fixed the typing of default props in viewer

### DIFF
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -393,21 +393,13 @@ class Viewer extends React.Component<{}, ViewerState> {
                             <label htmlFor={id}>{id}</label>
                             <input
                                 type="checkbox"
-                                onClick={(event) => {
-                                    const target: { value: string; }  = event.target as HTMLInputElement;
-                                    return this.turnAgentsOnOff(target.value)
-                                }}
+                                onClick={(event) => this.turnAgentsOnOff((event.target as HTMLInputElement).value)}
                                 value={id}
                                 defaultChecked={true}
                             />
                             <input
                                 type="checkbox"
-                                onClick={(event) => {
-                                    const target: { value: string; }  = event.target as HTMLInputElement;
-                                    this.turnAgentHighlightsOnOff(
-                                        target.value
-                                    )
-                                }}
+                                onClick={(event) => this.turnAgentHighlightsOnOff((event.target as HTMLInputElement).value)}
                                 value={id}
                                 defaultChecked={false}
                             />


examples/Viewer.tsx
```diff
@@ -394,19 +393,21 @@ class Viewer extends React.Component<{}, ViewerState> {
                             <label htmlFor={id}>{id}</label>
                             <input
                                 type="checkbox"
-                                onClick={(event) =>
-                                    this.turnAgentsOnOff(event.target.value)
-                                }
+                                onClick={(event) => {
+                                    const target: { value: string; }  = event.target as HTMLInputElement;
+                                    return this.turnAgentsOnOff(target.value)
+                                }}
```
meganrm, line 399: target.value was giving a typescript error that I had gotten in the habit of ignoring since it had been there awhile

examples/Viewer.tsx
```diff
@@ -394,19 +393,21 @@ class Viewer extends React.Component<{}, ViewerState> {
...
                                 value={id}
                                 defaultChecked={true}
                             />
                             <input
                                 type="checkbox"
-                                onClick={(event) =>
+                                onClick={(event) => {
+                                    const target: { value: string; }  = event.target as HTMLInputElement;
```
toloudis, line 406: would it be more accurate to type `const target` as HTMLInputElement?

toloudis (approved):

examples/Viewer.tsx
```diff
@@ -7,7 +7,6 @@ import type {
 import SimulariumViewer, {
     SimulariumController,
     RenderStyle,
-    SimulariumFileFormat,
```
schoinh, line 10: Was this not being used?

examples/Viewer.tsx
schoinh: Is it possible to just do `onClick={(event: React.MouseEvent<HTMLInputElement>) => {`?

examples/Viewer.tsx
schoinh, line 10: Was this not being used before?

examples/Viewer.tsx
toloudis: That sounds good. Also, I was even thinking 
```
this.turnAgentHighlightsOnOff((event.target as HTMLInputElement).value)
```

schoinh: Ugh 🤦🏻‍♀️ okay 😁 thanks for checking

examples/Viewer.tsx
meganrm: nope

PR #221, opened 2021-10-29
add a nightly build script

meganrm:
Problem
=======
We want to be able to test and use new features without having to have a local build 

Solution
========
Copied over the nightly build script from simularium website, and changed for this setup: change the folder name from dist to public. 

## Type of change

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -1,0 +1,42 @@
+name: github pages
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  deploy:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+
+      - name: Cache dependencies
+ 
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+            
+      - run: npm ci
+      - run: npm run gh-build
+      - name: Prepare tag
+        id: prepare_tag
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          TAG_NAME="${GITHUB_REF##refs/tags/}"
+          echo "::set-output name=tag_name::${TAG_NAME}"
+          echo "::set-output name=deploy_tag_name::deploy-${TAG_NAME}"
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,42 @@
...
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          TAG_NAME="${GITHUB_REF##refs/tags/}"
+          echo "::set-output name=tag_name::${TAG_NAME}"
+          echo "::set-output name=deploy_tag_name::deploy-${TAG_NAME}"
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./public
```
meganrm, line 40: This is the folder that gets uploaded to the gh-pages branch

toloudis (approved):
is this testable prior to merge?

meganrm:
> is this testable prior to merge?

I dunno, I think it's possible if we left this open until 0 UTC time the branch would run

toloudis:
I suppose you could just force it to run once, make sure it worked, and then put the cron back in

schoinh:
Question: Could you clarify what this will allow us to do again? Will we be able to go to some URL like xxx.github.io/xxx and use the viewer testbed with all (or almost all, depending on what has been merged since the last nightly build) the changes that are in the main branch?

toloudis:
Yes -- the main branch will be built nightly and the test viewer available on a github.io page, so you don't need to be on VPN to access the dev-aics-dtp-001 server that relies on being manually updated whenever I get around to it.

schoinh:
Another question: The test viewer is rendered through examples/index.html, so I was expecting to see a reference to `examples` somewhere in the config? Don't we need something like that if we want the github.io URL to launch the test viewer? Or does the `public` directory contain / take care of all that?

toloudis:
> Another question: The test viewer is rendered through examples/index.html, so I was expecting to see a reference to `examples` somewhere in the config? Don't we need something like that if we want the github.io URL to launch the test viewer? Or does the `public` directory contain / take care of all that?

The webpack config uses examples as the source but builds into public.
https://github.com/allen-cell-animated/simularium-viewer/blob/main/webpack.dev.js#L10

schoinh:
> The webpack config uses examples as the source but builds into public. https://github.com/allen-cell-animated/simularium-viewer/blob/main/webpack.dev.js#L10

Ohhh! Thank you thank you

schoinh (approved):
This will be a great improvement!!

EricJIsaac (approved):

Issue #187, opened 2021-09-09
Patchy textures


blairlyons:
## Description
On some machines, textures look patchy, like they are scaled incorrectly. See screenshot:
![image](https://user-images.githubusercontent.com/12102071/132740490-cfbaa565-04ce-4986-9770-1bc2e1cfcb85.png)

## Expected Behavior
Materials should render smoothly on any machine/browser that supports WebGL2

## Reproduction
Load this link on specific machines: https://simularium.allencell.org/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/actin012_3.h5.simularium

## Environment
This was only reproduced on a collaborator's machine. Their specs:

OS: Arch linux (latest, 5.13.13-arch1-1)
browser: chrome (93.0.4577.63)
GPU:
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 470.63.01    Driver Version: 470.63.01    CUDA Version: 11.4     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Quadro M1000M       On   | 00000000:01:00.0 Off |                  N/A |
| N/A   48C    P0    N/A /  N/A |    432MiB /  2000MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
```

PR #222, opened 2021-11-02
fix issue with color varying across same agent

toloudis:
Problem
=======
On some platforms the rendering is causing colors to vary across the surface of each agent, as if it is flipping between two different color indices.

chrome_0dTpn5FmYs

Solution
========
Add a rounding function when determining agent color index.

## Type of change

Steps to Verify:
----------------
1. I'm only aware of repro on Windows+nvidia graphics.

Screenshots (optional):
-----------------------

chrome_fRq5DScNiy


meganrm (approved):

schoinh (approved):

toloudis:
possible fix for #187

PR #223, opened 2021-11-03
Feature/error levels

### LINKED ISSUE #201: Errors should have consistent type
## What needs to happen?
Currently, some of the errors that get emitted by the viewer are strings and others are `Error` objects (see example of a string error being thrown [here](https://github.com/allen-cell-animated/simularium-viewer/blob/main/src/simularium/VisGeometry/GeometryStore.ts#L266)). All errors thrown by the viewer should be have a consistent type, preferably `Error` or its derivatives rather than string.

I think this would entail doing a global search for words like `reject`, `error`, and `throw` and making sure all errors are thrown as `Error`s, not strings.

## Why should we do this?
A developer of an application that uses simularium-viewer (like simularium-website) should not have to worry about errors coming from the viewer having different types.

## When does this need to get done?
Not aware of any specific errors that are not being handled in simularium-website because of this issue currently, but would be good to get this done in the next ~month to be sure we're not mishandling any errors that come in in an unexpected format.

### LINKED ISSUE #219: Send level with emitted error
## Use Case

We want to give info to the user, but differentiate between things that are actually broken, and things that are failing gracefully or just information that would be useful for them. 

## Acceptance Criteria

Along with the error message, send a level: "error", "warning", "info"?

## Details

We are sending errors in more than one way, one is through "HtmlError" and one is through throwing errors that are then caught through the `onError` handler. Need to make sure these can both have the same format.

meganrm:
Problem
=======
closes #219 
closes #201 

Solution
========
Added a level to front end error
Made all emitted errors FrontEndErrors
Left htmlData as a field because of how it's being used on the front end.

## Type of change

Steps to Verify:
----------------
I pushed my local build to the website repo if you want to test out these changes
2. View this [diff](https://github.com/allen-cell-animated/simularium-website/compare/local-viewer-install?expand=1)
3. Copy changes locally, plus install your viewer locally
4. Drag and drop the actin file without any geometry. 
5. You should never see the panel flash into error state, it goes from loading to success.

### DIFF
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -7,6 +7,7 @@ import type {
 import SimulariumViewer, {
     SimulariumController,
     RenderStyle,
+    ErrorLevel,
 } from "../src";
 import "./style.css";
 import { isEqual, findIndex } from "lodash";
@@ -14,6 +15,7 @@ import { isEqual, findIndex } from "lodash";
 import PointSimulator from "./PointSimulator";
 import PdbSimulator from "./PdbSimulator";
 import CurveSimulator from "./CurveSimulator";
+import FrontEndError from "../src/simularium/FrontEndError";
 
 const netConnectionSettings = {
     serverIp: "staging-node1-agentviz-backend.cellexplore.net",
@@ -134,6 +136,16 @@ class Viewer extends React.Component<{}, ViewerState> {
         e.preventDefault();
     };
 
+    public onError = (error: FrontEndError) => {
+        if (error.level === ErrorLevel.ERROR) {
+            window.alert(`ALERT, something is broken: ${error.htmlData}`);
+        } else if (error.level === ErrorLevel.WARNING) {
+            window.alert(`User warning, but not terrible: ${error.htmlData}`);
+        } else {
+            console.log(`Just for your info: ${error.htmlData}`);
+        }
+    };
+
     public onDrop = (e: Event): void => {
         this.onDragOver(e);
         const event = e as DragEvent;
@@ -150,7 +162,12 @@ class Viewer extends React.Component<{}, ViewerState> {
             const simulariumFileIndex = findIndex(filesArr, (file) =>
                 file.name.includes(".simularium")
             );
-            const simulariumFile = JSON.parse(parsedFiles[simulariumFileIndex]);
+            let simulariumFile;
+            try {
+                simulariumFile = JSON.parse(parsedFiles[simulariumFileIndex]);
+            } catch (error) {
+                return this.onError(new FrontEndError(error.message));
+            }
             const fileName = filesArr[simulariumFileIndex].name;
             const geoAssets = filesArr.reduce((acc, cur, index) => {
                 if (index !== simulariumFileIndex) {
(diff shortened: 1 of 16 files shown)


examples/Viewer.tsx
```diff
@@ -134,6 +136,16 @@ class Viewer extends React.Component<{}, ViewerState> {
         e.preventDefault();
     };
 
+    public onError = (error: FrontEndError) => {
+        if (error.level === ErrorLevel.ERROR) {
+            window.alert(`ALERT, something is broken: ${error.htmlData}`);
+        } else if (error.level === ErrorLevel.WARNING) {
+            window.alert(`User warning, but not terrible: ${error.htmlData}`);
+        } else {
+            console.log(`Just for your info: ${error.htmlData}`);
```
toloudis, line 145: here's the case I brought up where you are `console.log`ing html and it would make sense to have a non-html message

toloudis: i wonder if there is a utility function to just remove any `<tags>` from a string

toloudis: ... or maybe the console.log should log error.message?

toloudis (approved):

examples/Viewer.tsx
meganrm: I was a bit mistaken when we talked this morning: the errors always have a message, they may also have htmlData. I updated the example viewer to log just the message

schoinh (commented):
When I search for `throw new Error` in this branch, I still see a lot of them (like [here](https://github.com/allen-cell-animated/simularium-viewer/blob/8d433af321fac55e8981becd285c02e4248a7324/src/simularium/RemoteSimulator.ts#L331)). Just want to check if you made sure these are all caught and emitted as a FrontEndError when needed? I wish it were easier to tell from our code

src/controller/index.ts
```diff
@@ -46,7 +47,7 @@ export default class SimulariumController {
     public tickIntervalLength: number;
     public handleTrajectoryInfo: (TrajectoryFileInfo) => void;
     public postConnect: () => void;
-    public onError?: (errorMessage: string) => void;
+    public onError?: (errorMessage: FrontEndError) => void;
```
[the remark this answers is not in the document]
meganrm: Yeah I can change those, thanks

meganrm:
> When I search for `throw new Error` in this branch, I still see a lot of them (like [here](https://github.com/allen-cell-animated/simularium-viewer/blob/8d433af321fac55e8981becd285c02e4248a7324/src/simularium/RemoteSimulator.ts#L331)). Just want to check if you made sure these are all caught and emitted as a FrontEndError when needed? I wish it were easier to tell from our code

I went through and added comments to the code. Mostly those were all Errors being created within Promises or try catch blocks, that would then be passed up by the function that called it. I didn't want to make `FrontEndErrors` except for the places were it's truly the error being sent to the front end. If you're throwing an error within a function to prematurely exit, that should just be a normal Error. 

Thanks for asking though, because I did find one I missed, and I updated the logs on the example viewer so if an error gets sent in that doesn't have a level (ie is not a FrontEndError) we will hopefully be aware of it. 

There was one place where we are catching and error and logging it that might be better as a FrontEndError.

schoinh (approved):
Wow thank you for hunting down and fixing up all the errors, this is so much better than before

Issue #228, opened 2021-11-09
View and edit current camera transform


blairlyons:
## Use Case
It would be nice if users could view and edit the current camera transform.

## Acceptance Criteria
A user can choose to display or hide the current camera transform, and they can edit the values and see the view update.

## Details
camera transform includes: camera position, look at position, up vector, and field of view

Likely will want a hotkey to bring this up too

blairlyons:
duplicate

PR #232, opened 2021-11-12
Feature/headsup controls

toloudis:
Problem
=======
#228 and https://github.com/allen-cell-animated/simularium-website/issues/276
and also: 
* dat.gui (the existing way of showing our tweakable render settings for debugging) is really old and not actively developed, and a bit clunky in appearance
* we can't bring up the rendering settings controls in the simularium-website app but it would be nice to
* it would be nice to be able to take a quick screen capture while tweaking these parameters

Solution
========
Find `tweakpane` which has a very similar api to dat.gui, and integrate it!  
I added a button to do a png snapshot of the 3d view.
I also added live editable camera position and camera target coordinates.
The UI is now hidden behind a hotkey, currently ctrl-alt-1 (or control-option-1) and toggles on and off with that key combo.
This is meant to be hidden from all users except those "in the know".  If that is still too discoverable then we can just turn it off / on with special builds or debug flags.

Thanks to @meganrm and @schoinh for fine tuning on the CSS positioning to make it less clunky.

## Type of change

Steps to Verify:
----------------
1. try it out!

Screenshots (optional):
-----------------------
Simularium viewer 2021-11-12 11-27-23

-----------------------
I will put PR comments where the interesting bits of code are.


src/simularium/VisGeometry/index.ts
```diff
@@ -586,6 +597,12 @@ class VisGeometry {
 
     public setUpControls(element: HTMLElement): void {
         this.controls = new OrbitControls(this.camera, element);
+        this.controls.addEventListener("change", () => {
```
toloudis, line 600: this is how the camera moves cause the gui panel to update itself

src/viewport/index.tsx
```diff
@@ -333,6 +318,13 @@ class Viewport extends React.Component<ViewportProps & DefaultProps, ViewportSta
         if (prevProps.height !== height || prevProps.width !== width) {
             this.visGeometry.resize(width, height);
         }
+        if (prevState.showRenderParamsGUI !== this.state.showRenderParamsGUI) {
```
toloudis, line 321: this is where the show/hide happens.  this component has a piece of state called showRenderParamsGui but it has to be explicitly done through the visGeometry api below.

src/viewport/index.tsx
```diff
@@ -457,6 +463,7 @@ class Viewport extends React.Component<ViewportProps & DefaultProps, ViewportSta
                 false
             )
         );
+        document.addEventListener("keydown", this.handleKeyDown, false);
```
toloudis, line 466: I tried to add the keydown event to elements within this component but only adding it to the document seemed to actually work.  If anyone can make this part more elegant that would be great, since this component does already do touch and pointer events in a neater way.

src/simularium/VisGeometry/index.ts
```diff
@@ -268,43 +265,57 @@ class VisGeometry {
         this.threejsrenderer.setClearColor(this.backgroundColor, 1);
     }
 
-    public setupGui(): void {
-        const gui = new dat.GUI();
+    public setupGui(container?: HTMLElement): void {
+        this.gui = new Pane({
+            title: "AdvancedSettings",
```
schoinh, line 270: [nit] Can we do `"Advanced Settings"`

src/viewport/index.css
```diff
@@ -18,4 +18,15 @@
...
     border: solid 1px #979797;
-}
\ No newline at end of file
+}
+
+.tp-dfwv {
+    z-index: 1000;
+}
+
+.tp-rotv {
```
schoinh, line 27: Would be awesome to have some comments in here noting (1) that these are for the control pane (2) the difference between these 2 classes (why is z-index specified twice)

schoinh (approved):
Sweet!

src/viewport/index.css
toloudis: yes good idea. i'll add something

EricJIsaac (approved):

src/viewport/index.tsx
meganrm: this seems fine to me. If you wanted to I guess you could make a `this.documentHandlers = {"keydown" : this.handleKeyDown`. Or are you saying you'd prefer it to be attached to a different element?

meganrm (approved):

PR #233, opened 2021-11-23
Fix/dependabot

toloudis:
Problem
=======
github reports known vulnerabilities in dependencies

Solution
========
clean up the vulnerabilities by updating a few dependencies

## Type of change

Steps to Verify:
----------------
1. Should get rid of the reported vulnerabilities on github


package.json
```diff
@@ -57,7 +57,7 @@
         "babel-plugin-const-enum": "^1.1.0",
         "copy-webpack-plugin": "~9.0.1",
         "css-loader": "^3.2.0",
-        "cssnano": "^4.1.10",
+        "cssnano": "^5.0.11",
```
schoinh, line 60: Are you sure we're still using this dependency? When I do a global search for `cssnano` I get nothing, looks like we would see it in one of the config files? https://cssnano.co/docs/getting-started/

toloudis: I did the same, didn't find it, but didn't feel like removing it and accidentally breaking something, since there are zero code changes happening here.  I don't know what cssnano is supposed to do, or how it folds in.  I'll check that link.

toloudis: Took a look, yeah maybe we can remove it.

schoinh (approved):

EricJIsaac (approved):

PR #235, opened 2022-01-10
Fix/frame time issues when streamed trajectory doesn't start at t=0

### LINKED ISSUE #234: Frame time issues with endocytosis
## Description

When and only when streaming from the server, endocytosis doesn't play back properly. When the same file is downloaded from S3 and drag-and-dropped, it plays properly. Symptoms:

* When clicking step forward button, the time jumps from 0.1 to 0.3, then 0.5 and so on, instead of 0.1 to 0.2 to 0.3 and so on (the timestep is supposed to be 0.1).
* Clicking the step back button makes the viewer hang forever.

My first guess was this was related to floating point math error, but that should have been fixed in both the front and backend now. 

## Expected Behavior

* The step forward button should increment time by the expected time step interval.
* I should be able to click the Skip 1 Frame Back button and go 1 frame back, without the app breaking.

## Reproduction

1. Go here: https://simularium.allencell.org/viewer?trajFileName=endocytosis.simularium
2. Press step forward and step back buttons

I have verified that this is not an issue for any other networked trajectories.

## Environment

Staging and prod

schoinh:
Problem
=======
Closes #234 

The above bug was being caused because the backend incorrectly assumes that all trajectories start at t = 0. This is the quickest way to patch the issue so that endocytosis can be successfully streamed again, while we work out a better long-term solution.

Solution
========
* Grab the first frame's timestamp when a new trajectory arrives from the backend and VisData parses that data into frames, and save that as `firstFrameTime`, a property of class `VisData`. 
* When requesting a frame at a specific time from the backend, request the frame at `time - firstFrameTime` instead of `time` to compensate for the backend assumption that trajectories start at time 0.

Thank you @toloudis for finding the root of the issue in the backend and suggesting this fix!

## Type of change

Steps to Verify:
----------------
1. Pull this branch, and try streaming endocytosis in the viewer testbed. Skipping frames forward and backward should successfully increment or decrement by 0.1.
2. Install this in simularium-website and make sure it works when wrapped in simularium-website as well
3. Other streamed and drag and drop trajectories should work without problem as before.

### DIFF
--- a/src/simularium/VisData.ts
+++ b/src/simularium/VisData.ts
@@ -460,6 +460,11 @@ class VisData {
         this.netBuffer = new ArrayBuffer(0);
     }
 
+    public clearForNewTrajectory(): void {
+        this.clearCache();
+        this.firstFrameTime = null;
+    }
+
     public cancelAllWorkers(): void {
         // we need to be able to terminate any queued work in the worker during trajectory changeovers
         if (
--- a/src/simularium/VisData.ts
+++ b/src/simularium/VisData.ts
@@ -41,7 +41,6 @@ class VisData {
     private frameCache: AgentData[][];
     private frameDataCache: FrameData[];
     private webWorker: Worker | null;
-    public firstFrameTime: number;
 
     private frameToWaitFor: number;
     private lockedForFrame: boolean;
@@ -51,6 +50,7 @@ class VisData {
     // eslint-disable-next-line @typescript-eslint/naming-convention
     private _dragAndDropFileInfo: TrajectoryFileInfo | null;
 
+    public firstFrameTime: number;
     public timeStepSize: number;
 
     /**
--- a/src/simularium/VisData.ts
+++ b/src/simularium/VisData.ts
@@ -471,6 +471,18 @@ class VisData {
         }
     }
 
+    // Add parsed frames to the cache and save the timestamp of the first frame
+    private addFramesToCache(frames: ParsedBundle): void {
+        Array.prototype.push.apply(this.frameDataCache, frames.frameDataArray);
+        Array.prototype.push.apply(
+            this.frameCache,
+            frames.parsedAgentDataArray
+        );
+        if (this.firstFrameTime === -1) {
+            this.firstFrameTime = frames.frameDataArray[0].time;
+        }
+    }
+
     public parseAgentsFromNetData(msg: VisDataMessage | ArrayBuffer): void {
         if (msg instanceof ArrayBuffer) {
             const floatView = new Float32Array(msg);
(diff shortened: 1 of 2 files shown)


src/simularium/VisData.ts
```diff
@@ -357,6 +358,7 @@ class VisData {
         }
         this.frameCache = [];
         this.frameDataCache = [];
+        this.firstFrameTime = -1; // Assumes negative timestamps don't exist
```
toloudis, line 361: you could use null or undefined as the uninitialized value to avoid this assumption

schoinh: Oh good point, for some reason I thought it would be difficult to deal with TS that way, but it was no problem, fixed!

src/simularium/VisData.ts
```diff
@@ -469,6 +471,18 @@ class VisData {
...
 
+    // Add parsed frames to the cache and save the timestamp of the first frame
+    private addFramesToCache(frames: ParsedBundle): void {
+        Array.prototype.push.apply(this.frameDataCache, frames.frameDataArray);
+        Array.prototype.push.apply(
+            this.frameCache,
+            frames.parsedAgentDataArray
+        );
+        if (this.firstFrameTime === null) {
+            this.firstFrameTime = frames.frameDataArray[0].time;
```
toloudis, line 482: I just noticed this- do you want to use `this.frameDataCache[0].time` instead of `frames.frameDataArray[0].time` here?  i.e if this function is always appending to the cache, do you want to use the first item in the cache or the first item of those being appended?  i guess the null check is protecting that anyway.

toloudis (approved):

src/simularium/VisData.ts
meganrm: when does this get reset to null? Is visData re-constructed with each new trajectory? I just dont remember

schoinh: It doesn't get re-constructed with each new trajectory. I reset it manually when a trajectory is cleared (ie with the "Clear Trajectory" button in the viewer testbed) or when a trajectory is changed, in lines 271 and 287 in src/controller/index.ts

meganrm: oh ok, I was looking for it in this file. Probably not a big deal, but is there any sort of "reset" function in this file? If not it might make more sent to make a `clearForNewTrajectory` function that both clears the cache and resets the `firstFrameTime`

meganrm: also is `clearFile` called before `changeFile`? If so it seems like you should only need to reset it in one place

schoinh: No, `clearFile` is only used when file is being cleared not being replaced by a new file. But I will make a `clearForNewTrajectory` function, makes a lot of sense!

meganrm (approved):

src/simularium/VisData.ts
schoinh: @toloudis I looked into your question. The cache gets cleared at various points during playback, e.g., when stepping through frames (not sure why it's implemented that way, but I think that's for another discussion) so it's not any more stationary than `frames.frameDataArray`. It doesn't matter which I use because of the null check, like you said, so I'll just keep it like this.

PR #236, opened 2022-01-12
fix the sample app build: css file was not being pre built

toloudis:
Problem
=======
the gh-pages build has been failing.  

Solution
========
a necessary css file was not being preprocessed.  Now it will be.

## Type of change

Change summary:
---------------
I just made sure the npm commands pre-generate the css file.  This could probably be folded into the webpack config instead, but this is the quick version of the fix.   
Note that this bug does not affect production builds at all - only the example test app.

Steps to Verify:
----------------
1. `rm style/style.css`
1. `npm start`
1. prior to this change, the demo app would fail to start.
1. alternately, observe that the gh-actions gh-pages workflow succeeds.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,35 +22,35 @@
             }
         },
         "@babel/code-frame": {
-            "version": "7.14.5",
-            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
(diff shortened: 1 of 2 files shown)


schoinh (approved):
Thank you!

meganrm (approved):

PR #244, opened 2022-01-28
Feature/update three

toloudis:
Problem
=======
Dependabot says we need to update three.js

Solution
========
Update threejs to latest, and make corresponding code changes. In particular, DataTexture needs to make sure `needsUpdate` is set, and RGBFormat is deprecated in favor of RGBAFormat.

## Type of change
* Security/Maintenance (non-breaking change which updates dependencies)

Steps to Verify:
----------------
1. No behavior should change

### DIFF
--- a/src/simularium/VisGeometry/GeometryStore.ts
+++ b/src/simularium/VisGeometry/GeometryStore.ts
@@ -297,25 +297,29 @@ class GeometryStore {
         const objLoader = new OBJLoader();
         this.prepMeshRegistryForNewObj(url);
         return new Promise((resolve, reject) => {
-            objLoader.load(
-                url,
-                (object: Object3D) => {
-                    this.handleObjResponse(url, object);
-                    resolve(object);
-                },
-                (xhr) => {
-                    this.mlogger.info(
-                        url,
-                        " ",
-                        `${(xhr.loaded / xhr.total) * 100}% loaded`
-                    );
-                },
-                (error) => {
-                    // if the request fails, leave agent as a sphere by default
-                    this.mlogger.warn("Failed to load mesh: ", error, url);
-                    return reject(`Failed to load mesh: ${url}`);
-                }
-            );
+            try {
+                objLoader.load(
+                    url,
+                    (object: Object3D) => {
+                        this.handleObjResponse(url, object);
+                        resolve(object);
+                    },
+                    (xhr) => {
+                        this.mlogger.info(
+                            url,
+                            " ",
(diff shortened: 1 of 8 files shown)


src/simularium/VisGeometry/GeometryStore.ts
```diff
@@ -297,25 +297,29 @@ class GeometryStore {
...
-                        `${(xhr.loaded / xhr.total) * 100}% loaded`
-                    );
-                },
-                (error) => {
-                    // if the request fails, leave agent as a sphere by default
-                    this.mlogger.warn("Failed to load mesh: ", error, url);
-                    return reject(`Failed to load mesh: ${url}`);
-                }
-            );
+            try {
```
schoinh, line 300: This try-catch addition isn't related to the three.js update, right? (Fine with me, just wondering)

src/test/GeometryStore.test.ts
```diff
@@ -149,6 +149,9 @@ describe("GeometryStore module", () => {
         });
         test("if a request fails, returns a sphere with an error message", async () => {
             const store = new GeometryStore();
+            // TODO threejs' OBJLoader now uses the Request api under the hood, and Request is
+            // not available in nodejs (the environment our tests run in) without a polyfill
+            // such as node-fetch. I tried to add node-fetch here but I couldn't get it to work.
             const returned = await store.mapKeyToGeom(1, {
```
schoinh, line 155: Confused...  So `store.mapKeyToGeom()` uses threejs' OBJLoader which uses the Request API which doesn't work in nodejs? And using node-fetch wouldn't work? So how come the tests still pass?

toloudis: the tests pass because of the extra try-catch I added.  The error is "Request is not defined" which was throwing but not caught as part of the FileLoader's own error handling.

src/simularium/VisGeometry/GeometryStore.ts
toloudis: the try catch allows me to handle the error in nodejs not knowing about Request (the fetch API) when we run unit tests that call this code.

schoinh: Oh... maybe if the `error` type caught in the catch block in the test environment is specific enough it'd be good to only handle that type of error? And maybe also add a comment saying the try-catch block was added for the test environment? Otherwise I'm scratching my head why we need error handling in both line 314 and line 320. Not a huge deal though...

schoinh (approved):

meganrm (approved):

PR #247, opened 2022-02-21
Feature/load binary trajectories

toloudis:
Problem
=======
Drag and drop can be inefficient for larger trajectories when the file is in JSON format.  Also, the current binary streamed format is not suitable for combining with the parts of the data that are JSON into a single file.

Solution
========
Binary files take up less disk space and can theoretically be parsed faster.
Add binary file format and parsing to the viewer for drag and drop usage.
The "collection of blocks" paradigm allows us to pack text data into blocks just as easily as binary data.

## Type of change

Change summary:
---------------
* Some simplifying assumptions: the new binary format is only readable as a drag n drop file, and is not understood as a network data stream yet.   It is currently only written out from simulariumio python package.
* Only one frame is parsed at a time and it is done as a copy.


jest.config.js
```diff
@@ -5,11 +5,13 @@ module.exports = {
...
     testEnvironment: "jsdom",
+    roots: ["<rootDir>/src/"],
     // From https://jestjs.io/docs/webpack#mocking-css-modules
     moduleNameMapper: {
         "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
             "<rootDir>/scripts/jestAssetTransformer.js",
         "\\.(css|less)$": "identity-obj-proxy",
     },
     transformIgnorePatterns: ["<rootDir>/node_modules/three/examples/(?!jsm/)"],
+    setupFiles: ["<rootDir>/src/test/testSetup.ts"],
```
toloudis, line 16: this is fixing problems with running the tests due to some missing dependencies that are present in the browser but not in the jsdom test environment

package.json
```diff
@@ -96,7 +97,7 @@
     },
     "dependencies": {
         "@fortawesome/fontawesome-free": "^5.15.2",
-        "@fortawesome/fontawesome-svg-core": "^1.2.34",
+        "@fortawesome/fontawesome-svg-core": "~1.2.34",
```
toloudis, line 100: this is fixing a spontaneous bug that popped up with weird incompatibilities due to updates in fontawesome.  I had a github link to the actual issue somewhere....

examples/Viewer.tsx
```diff
@@ -161,43 +157,62 @@ class Viewer extends React.Component<{}, ViewerState> {
...
+            loadSimulariumFile(filesArr[simulariumFileIndex])
+                .then((simulariumFile: ISimulariumFile) =>
+                    // textify every file that was not the simularium file:
+                    Promise.all(
+                        filesArr.map((element, index) => {
+                            if (index !== simulariumFileIndex) {
+                                // is async call
+                                return element.text();
+                            } else {
+                                return simulariumFile;
```
meganrm, line 177: For clarity, it seems like you should return the async `loadSimulariumFile(element)` here, so then you're mapping over an array and no matter what you're returning a Promise instead of having this nested Promise all within another Promise.

toloudis: I'm not sure that would make it clearer to me but it will certainly make it more concise without any loss of clarity.  I'll make that change.

src/simularium/JsonFileReader.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        const spatialData = this.simulariumFile.spatialData;
+        if (!spatialData) {
+            throw new FrontEndError(
+                "Simularium files need 'spatialData' array"
+            );
+        }
+        spatialData.bundleData.sort(
+            (a: VisDataFrame, b: VisDataFrame): number =>
+                a.frameNumber - b.frameNumber
```
meganrm, line 24: did we do this sorting before? Is it not ok to assume they are in order?

src/util.ts
```diff
@@ -47,3 +51,32 @@ export function getFileExtension(pathOrUrl: string): string {
         pathOrUrl
     );
 }
+
+export function isBinarySimulariumFile(fileContents: Blob): Promise<boolean> {
+    const first16blob: Blob = fileContents.slice(0, 16);
+    // compare this with "SIMULARIUMBINARY"
+    return first16blob.text().then((text) => {
+        return text === "SIMULARIUMBINARY";
```
meganrm, line 59: You make this a constant as "secret", is there a reason you're using the string here instead of exporting that constant?

src/util.ts
```diff
@@ -47,3 +51,32 @@ export function getFileExtension(pathOrUrl: string): string {
...
+        .then((isBinary): Promise<ArrayBuffer | string> => {
+            if (isBinary) {
+                return file.arrayBuffer();
+            } else {
+                return file.text();
+            }
+        })
+        .then((fileContents: ArrayBuffer | string) => {
+            if (typeof fileContents === "string") {
+                console.log("TEXT JSON FILE");
```
meganrm, line 74: Don't we have a logger for this sort of stuff?

src/simularium/LocalFileSimulator.ts
```diff
@@ -155,13 +134,7 @@ export class LocalFileSimulator implements ISimulator {
     }
 
     public gotoRemoteSimulationTime(time: number): void {
```
meganrm, line 136: I can't remember, why are all these functions called "remote" something when it's in the `localFileSimulator`? is it because of the shared interface? It's definitely confusing

toloudis: Yes, remnants of the original code Eric wrote. When I refactored I didn't want to make the diff worse by renaming things so I was saving it for another maintenance PR.  See comment here:
https://github.com/simularium/simularium-viewer/blob/main/src/simularium/ISimulator.ts#L5

src/util.ts
toloudis: That was for my own debugging really... I'll get rid of it.

src/util.ts
toloudis: Yes, I'll do something better to remove this redundancy.

src/simularium/JsonFileReader.ts
toloudis: that code was there before, and I didn't have any reason to change it.  I don't know for sure but we probably have other implicit assumptions that they are in-order in the file.

meganrm (approved):

src/simularium/LocalFileSimulator.ts
```diff
@@ -155,13 +134,7 @@ export class LocalFileSimulator implements ISimulator {
...
-        const { bundleData } = this.simulariumFile.spatialData;
-        const { timeStepSize } = this.simulariumFile.trajectoryInfo;
-
-        // Find the index of the frame that has the time matching our target time
-        const frameNumber = bundleData.findIndex((bundleData) => {
-            return compareTimes(bundleData.time, time, timeStepSize) === 0;
-        });
+        const frameNumber = this.simulariumFile.getFrameIndexAtTime(time);
```
schoinh, line 137: I love ISimulariumFile!

schoinh (approved):
Looks good to me!

src/simularium/BinaryFileReader.ts
```diff
@@ -0,0 +1,269 @@
...
+interface BlockInfo {
+    type: BlockTypeEnum;
+    offset: number;
+    size: number;
+}
+interface Header {
+    version: number;
+    blocks: BlockInfo[];
+}
+const SECRET = "SIMULARIUMBINARY";
```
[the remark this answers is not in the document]
toloudis: yeah, that's better

PR #249, opened 2022-03-01
Feature/cam capture

toloudis:
Problem
=======
The last UI meeting generated some ideas on how to capture cameras and save them.  This implements some of those core ideas into the "advanced settings" dialog, accessible via `control-option-1` keypress.

Solution
========
There are 3 "stored" cameras.  Click "Save" to store the current camera settings into one of the saved cameras.  Click "View" to load a saved camera's setting into the current viewing camera.  Click "Export Camera" to output the current camera's settings to a json file, and "Import Camera" to load a json file with saved camera settings.  Currently only `position` and `target` are saved.  
An interesting future update would be to try to smoothly animate the camera into position instead of just setting it directly.

## Type of change

Screenshots (optional):
-----------------------
Simularium viewer 2022-03-01 15-04-05


package.json
```diff
@@ -99,14 +99,15 @@
         "@fortawesome/fontawesome-svg-core": "^1.2.34",
         "@fortawesome/free-solid-svg-icons": "^5.15.2",
         "@fortawesome/react-fontawesome": "^0.1.14",
-        "@tweakpane/core": "^1.0.6",
+        "@tweakpane/core": "^1.0.8",
         "@tweakpane/plugin-essentials": "^0.1.4",
+        "acorn": "^8.7.0",
```
toloudis, line 104: I was getting some inexplicable `npm install` warnings about acorn needing to be installed. I'm not really sure what acorn is for - I am going to look into this now.

src/viewport/index.tsx
```diff
@@ -365,7 +365,7 @@ class Viewport extends React.Component<
 
         const event = e as KeyboardEvent;
         // control-option-1 (mac) or ctrl-alt-1 (windows)
-        if (event.key === "1" && event.altKey && event.ctrlKey) {
+        if (event.code === "Digit1" && event.altKey && event.ctrlKey) {
```
toloudis, line 368: This had been broken on Firefox.  I believe this fixes it cross-browser.

src/simularium/VisGeometry/index.ts
```diff
@@ -276,6 +306,88 @@ class VisGeometry {
         fcam.addInput(this.camera, "position");
         fcam.addInput(this.controls, "target");
 
+        const grid1: ButtonGridApi = this.gui.addBlade({
```
meganrm, line 309: this could be reduced down to a `forEach` loop on an array of `[this.cam1, this.cam2, this.cam3]` so that there isn't so much repeated code in here.  Or `[{camera: this.cam1, label: "Cam 1"}, ...]`

src/simularium/VisGeometry/index.ts
```diff
@@ -276,6 +306,88 @@ class VisGeometry {
...
+            view: "buttongrid",
+            size: [2, 1],
+            cells: (x, y) => ({
+                title: [["View", "Save"]][y][x],
```
meganrm, line 313: Small thing, but maybe "Go to" instead of "View". "View" makes me think i might be looking at the camera instead of through the camera. But I know this isn't really a true UX design so could wait

toloudis: I also considered "Activate"...

schoinh (approved):

package.json
toloudis: yes, or peer... I think it's really supposed to be peer

src/simularium/VisGeometry/index.ts
```diff
@@ -265,6 +273,28 @@ class VisGeometry {
         this.threejsrenderer.setClearColor(this.backgroundColor, 1);
     }
 
+    private loadCamera(cameraSpec: CameraSpec): void {
+        // TODO add other parameters from CameraSpec?
```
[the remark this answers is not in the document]
toloudis: not much more complicated, but users don't generally directly manipulate up and fov when interacting with the viewer.  Easy enough to add them later as needed.

src/simularium/VisGeometry/index.ts
toloudis: done!

package.json
toloudis: i reset everything and was able to get rid of the acorn dep.

schoinh: Awesome!

meganrm (approved):

PR #251, opened 2022-03-09
Prototype movie saving

### LINKED ISSUE #240: Prototype movie saving
## Use Case
As a developer, I need to know the best way to save a series of rendered images as a movie file on my local machine.

## Acceptance Criteria
A developer can save a series of rendered images as a movie locally.

Specific questions to answer to unblock UX design:
- What can we show the users during recording? Can we show these?
    - Duration (clock time of recording)
    - Number of frames
    - Simulated time
- What can we show at the end of recording (before download)? Can we show the above plus these?
    - File size

## Details
There's already discussion about this on this issue simularium/simularium-planning#22

schoinh:
Problem
=======
Closes #240 

Solution
========
Bare-bones test implementation -- Use the [MediaRecorder API](https://developer.mozilla.org/en-US/docs/Web/API/Media_Streams_API) to capture canvas frames as a webm video and [ffmpeg.wasm](https://github.com/ffmpegwasm/ffmpeg.wasm#ffmpegwasm) to convert it to an mp4 file before downloading it to the user's machine.

### Some questions that were answered through prototyping:
* Can we output videos of satisfactory quality?
    * Yes it seems we can raise the quality of the output video by raising `videoBitsPerSecond` where we instantiate a `new MediaRecorder`
* Can we output videos as mp4 files?
    * Yes it is possible with ffmpeg.wasm (web version of ffmpeg) 
* Can we display the clock duration of the recording to the user while recording?
    * Yes I have implemented a duration indicator using basic `setInterval()` but there might be a better way to do this. I found sometimes I'm off by 1 second with the current implementation.
* Can we display file size in a modal before the user confirms they want to download the file?
    * Yes, we should be able to, but depending on the user's browser setting the browser will also do this for us automatically.
* Can we display a progress indicator while the file is processing?
    * Yes it seems like we should be able to with ffmpeg (the transcoding to mp4 makes up almost all of the processing time): https://github.com/ffmpegwasm/ffmpeg.wasm/blob/master/docs/api.md#createffmpegoptions-ffmpeg

### Remaining issues to address
* ffmpeg requires [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) to be enabled which requires a page to be "cross-origin isolated"... I was able to make it with the dev server work by supplying in the necessary headers but not sure how it would work for non-dev environment. See note under "Browser" in the [ffmpeg readme](https://github.com/ffmpegwasm/ffmpeg.wasm#installation)
* I think these two issues only arise with ffmpeg in use but I'm not 100% positive:
    * Current implementation breaks the page if I stop recording while a trajectory is still playing. I got around this for now by forcing the trajectory to pause first when "Stop Recording" button is pressed
    * Current implementation fails (output video is empty) if I start playing the trajectory before pressing the "Start Recording" button. I tried one solution which was to pause the trajectory, start recording, then resume the trajectory playback anytime the Start Recording button is clicked, but that resulted in videos with a slo-mo part in the middle. I think this might be some kind of race condition? And/or something to do with a variable frame rate: https://video.stackexchange.com/questions/24576/output-of-ffmpeg-has-a-slow-motion-part-what-is-it-and-how-to-avoid/24578#24578

## Type of change

* Proto-New feature (non-breaking change which adds functionality)

Screenshots (optional):
-----------------------
image

### DIFF
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -75,6 +75,7 @@ interface ViewerState {
     totalDuration: number;
     uiDisplayData: UIDisplayData;
     recordingTime: number; // in seconds
+    videoOutputStatus: string;
 }
 
 const simulariumController = new SimulariumController({});
@@ -102,6 +103,7 @@ const initialState = {
         hiddenAgents: [],
     },
     recordingTime: 0,
+    videoOutputStatus: "",
 };
 
 class Viewer extends React.Component<{}, ViewerState> {
@@ -150,6 +152,9 @@ class Viewer extends React.Component<{}, ViewerState> {
     }
 
     private handleVideoDataAvailable = async (event: BlobEvent) => {
+        this.setState({
+            videoOutputStatus: "Processing..."
+        });
         console.log("Video data available:", event.data)
         // Following example from:
         // https://github.com/ffmpegwasm/react-app/blob/master/src/App.js
@@ -165,7 +170,9 @@ class Viewer extends React.Component<{}, ViewerState> {
         }
         const mp4Data = await transcodeToMp4(webmBlob);
         const mp4Blob = new Blob([mp4Data.buffer], { type: "video/mp4" });
-        console.log(`Size of output video: ${mp4Blob.size / 1000} KB`);
+        this.setState({
+            videoOutputStatus: `Video file size: ${mp4Blob.size / 1000} KB`
+        });
         const url = URL.createObjectURL(mp4Blob);
         let a = document.createElement("a");
         document.body.appendChild(a);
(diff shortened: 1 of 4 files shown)


toloudis:
some suggested video bitrates: https://support.google.com/youtube/answer/1722171?hl=en#zippy=%2Cbitrate (5 MBps is still on the low side!)

toloudis:
https://github.com/allen-cell-animated/nucmorph-colorizer/issues/33 has a solution that works without any shared memory requirements or workers.   The solution is only Chrome-compatible (and must be used in a secure https session only)  but it was straightforward to implement.  It's using a web API which could be implemented in the future in other browsers.

PR #257, opened 2022-03-21
Feature/refactor viewer from data

toloudis:
Problem
=======
The visualization code is entangled with the simularium data transfer code.  The repository could use a more logical organization and eventually the viewer/renderer could be separated from the data handlers.

Solution
========
No functionality changes.

Currently just move some files into a different folder structure.  "simularium" holds all the simularium data handling code, and "visGeometry" holds the display code.  Very much open to renaming the "visGeometry" folder with something like "vis" or "display" or "graphics" or "rendering"

There is still some common code used by both simularium and visGeometry which could be examined and maybe stored in a "common" folder or utils modules.

I also refactored implementation details of the path drawing code into agentPath.ts

## Type of change

* code maintenance / refactor


src/simularium/types.ts
```diff
@@ -1,4 +1,4 @@
-import { GeometryDisplayType } from "./VisGeometry/types";
+import { GeometryDisplayType } from "../visGeometry/types";
```
toloudis, line 1: this is an example of a data structure that should be factored out of visGeometry and into a common module so that simularium doesn't have to depend on visGeometry.

src/simularium/VisData.ts
```diff
@@ -198,7 +198,7 @@ class VisData {
         let j = AGENTS_OFFSET;
         for (let i = 0; i < expectedNumAgents; i++) {
             const agentData: AgentData = {
-                visType: -1,
+                "vis-type": -1,
```
meganrm, line 201: Whats the reason for this change? Was this not working correctly before?

meganrm (approved):

src/simularium/VisData.ts
toloudis: unfortunately, the server provides this data in json as "vis-type".  I'm just making it all consistent here in keeping that name everyplace.

toloudis: I suppose the alternative is to go the other way and modify it as soon as it's read, which would probably be better but a bit uglier at parse time.

meganrm: Ok, and all the other names here are how the server provides them? I was just surprised this is `vis-type` while there are other camelCase names

toloudis: it has been like that since early days.  I have a feeling Eric named it from the server side before the js code was even written.

Issue #190, opened 2021-09-28
Support "Cube" displayType


meganrm:
## Use Case

If the user specifies "Cube" as their geometry type in their trajectory, they want to see cubes. 

## Acceptance Criteria

The agents show as cubes when the agent's `geometry.displayType` is set to Cube.

PR #260, opened 2022-03-29
Feature/cubes

meganrm:
Problem
=======
Support for cubes #190 

Solution
========
Used the same set up we had for spheres and tested it with cellpack file
I also moved the sphere geometry creation to the geometry store because that was the only place we used it. 

## Type of change

Steps to Verify:
----------------
1. Change the display type on a simularium file to CUBE
2. Drag and drop on the test viewer

Screenshots (optional):
-----------------------
Screen Shot 2022-03-29 at 11 46 36 AM

### DIFF
--- a/src/visGeometry/GeometryStore.ts
+++ b/src/visGeometry/GeometryStore.ts
@@ -12,7 +12,6 @@ import {
 import { checkAndSanitizePath, getFileExtension } from "../util";
 import PDBModel from "./PDBModel";
 import { InstancedMesh, InstanceType } from "./rendering/InstancedMesh";
-import VisAgent from "./VisAgent";
 import TaskQueue from "../simularium/TaskQueue";
 import { AgentTypeVisData } from "../simularium/types";
 
--- a/src/visGeometry/GeometryStore.ts
+++ b/src/visGeometry/GeometryStore.ts
@@ -434,8 +434,8 @@ class GeometryStore {
         const isPDB = displayType === GeometryDisplayType.PDB;
         if (!url || (!isMesh && !isPDB)) {
             const lookupKey = displayType;
-            let geometry;
-            // displayType not either pdb or obj, will show a sphere
+            let geometry: MeshLoadRequest;
+            // TODO: handle gizmo here
             if (displayType === GeometryDisplayType.CUBE) {
                 geometry = this.createNewCubeGeometry(lookupKey);
                 this.setGeometryInRegistry(
@@ -444,6 +444,7 @@ class GeometryStore {
                     GeometryDisplayType.CUBE
                 );
             } else {
+                // default to a sphere
                 geometry = this.createNewSphereGeometry(lookupKey);
                 this.setGeometryInRegistry(
                     lookupKey,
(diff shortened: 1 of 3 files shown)


src/visGeometry/GeometryStore.ts
```diff
@@ -408,16 +433,25 @@ class GeometryStore {
         this.mlogger.debug(`Geo for id ${id} set to '${url}'`);
         const isMesh = displayType === GeometryDisplayType.OBJ;
         const isPDB = displayType === GeometryDisplayType.PDB;
-        if (!url) {
+        if (!url || (!isMesh && !isPDB)) {
```
toloudis, line 436: is this if statement even needed here? 
Do we even need to check for url if displayType is CUBE or SPHERE? 
If not, then you could just do `if (!url && displayType===GeometryDisplayType.CUBE)` and likewise for `SPHERE` without the outer if statement.

toloudis (approved):
I think that one if statement should be simplified to be more readable (if it doesn't break anything) but LGTM.

meganrm:
> I think that one if statement should be simplified to be more readable (if it doesn't break anything) but LGTM.

I just factored it out into a function. It's not the most concise it can be, but to me it's the clearest. Let me know what you think.

src/visGeometry/GeometryStore.ts
meganrm: I tried to make this clear in the function I just wrote, but it's checking for the case where someone selected "OBJ" or "PDB" but forgot to include a url.

src/visGeometry/GeometryStore.ts
```diff
@@ -43,6 +48,31 @@ class GeometryStore {
...
+        1
+    );
+
+    private static loadPrimitive = (
+        displayType: GeometryDisplayType,
+        url?: string
+    ) => {
+        if (!url) {
+            // if there isn't an url to load, even if they selected PDB or OBJ
+            // we have to default to a sphere. May change depending on how we handle the gizmo
```
toloudis, line 65: I think this is really trying to say that some displaytypes require a url to work.

if there isn't a url AND they selected PDB or OBJ, then we have to default to a sphere.  Otherwise we can use the specified type that doesn't require a url.
Is that correct?

src/visGeometry/GeometryStore.ts
```diff
@@ -43,6 +48,31 @@ class GeometryStore {
...
     public mlogger: ILogger;
+    public static sphereGeometry: SphereBufferGeometry =
+        new SphereBufferGeometry(1, 32, 32);
+    public static cubeGeometry: BoxBufferGeometry = new BoxBufferGeometry(
+        1,
+        1,
+        1
+    );
+
+    private static loadPrimitive = (
```
toloudis, line 59: `shouldLoadPrimitive`

src/visGeometry/GeometryStore.ts
toloudis: `if displaytyperequiresurl(displayType) && !url) { use sphere } ...` ?

blairlyons (approved):

src/visGeometry/GeometryStore.ts
meganrm: yeah that's what I was trying to say in the comment

PR #261, opened 2022-04-12
Feature/metaballs

toloudis:
Problem
=======
One whole mesh per agent per time step is remarkably inefficient memory usage.  In particular this is exhibited by one of the ABM trajectories.  

Solution
========
Because of the nature of the grid-based data and an approximation where a sphere is used at each filled grid point, I hypothesized that a metaball representation would look good and provide some savings in memory and bandwidth for trajectory playback.  

Metaballs will use the existing per-agent "subpoints".  Each agent can have N metaballs. where each ball is 4 floats: x,y,z,radius.  The xyz will be a translation relative to the agent's xyz position.   
The triangle mesh will be reconstructed per-timestep from the metaball representation.  In this way we have a sort of mesh compression format.

## Type of change

## Note
This is a draft pr with some todo's still to occur. 
It is also a very slow brute-force implementation but might be ok to merge before doing the next optimization work to help it scale a little better to more agents.


src/visGeometry/index.ts
```diff
@@ -845,7 +846,7 @@ class VisGeometry {
             this.scene.autoUpdate = false;
 
             // collect up the meshes that have > 0 instances
-            const meshTypes: InstancedMesh[] = [];
+            const meshTypes: (InstancedMesh | MetaballMesh)[] = [];
```
toloudis, line 849: TODO:  abstract all these `InstancedMesh | MetaballMesh` to an interface

src/visGeometry/rendering/GBufferPass.ts
```diff
@@ -39,7 +39,7 @@ class GBufferPass {
     public instancedMeshGroup: Group;
     // instancedMeshGroup consists of fibers and meshes:
     public fibers: InstancedFiberGroup;
-    public meshTypes: InstancedMesh[];
+    public meshTypes: (InstancedMesh | MetaballMesh)[];
```
toloudis, line 42: TODO: abstract all these InstancedMesh | MetaballMesh to an interface

src/visGeometry/rendering/GBufferPass.ts
```diff
@@ -54,7 +54,7 @@ class GBufferPass {
     public setMeshGroups(
         instancedMeshGroup: Group,
         fibers: InstancedFiberGroup,
-        meshes: InstancedMesh[]
+        meshes: (InstancedMesh | MetaballMesh)[]
```
toloudis, line 57: TODO: abstract all these InstancedMesh | MetaballMesh to an interface

src/visGeometry/types.ts
```diff
@@ -1,11 +1,12 @@
 import { Object3D } from "three";
 import PDBModel from "./PDBModel";
 import { InstancedMesh } from "./rendering/InstancedMesh";
+import { MetaballMesh } from "./rendering/MetaballMesh";
 
 export interface MeshLoadRequest {
     mesh: Object3D;
     cancelled: boolean;
-    instances: InstancedMesh;
+    instances: InstancedMesh | MetaballMesh;
```
toloudis, line 9: TODO: abstract all these InstancedMesh | MetaballMesh to an interface

PR #262, opened 2022-04-19
Feature/export i simularium file

toloudis:
Problem
=======
The implementation of Drag and drop file loading has changed, and the simularium-website needs to be updated, but the necessary types are not visible to simularium-website yet.

Solution
========
Export ISimulariumFile and loadSimulariumFile from the viewer

## Type of change


toloudis:
Note that this PR will require a version bump and release before it can be integrated with simularium-website

meganrm (approved):

blairlyons (approved):

Issue #237, opened 2022-01-18
Prototype initial autoconversion UI - Smoldyn


blairlyons:
## Use Case
As a developer using the example viewer, I need to be able to provide additional display information about a trajectory through a UI before autoconverting it.

## Acceptance Criteria
- When a Smoldyn trajectory is loaded locally in the viewer, a UI menu opens to collect display data from the user according to the corresponding template
- When the user clicks a submit button, example data loads in the viewer
- Using the data the user provided is out of scope, for next issue https://github.com/simularium/simularium-lambdas/issues/1

## Details
This depends on UI templates being defined in https://github.com/simularium/simulariumio/issues/74 

The scope of this issue is only a prototype for developers, another later issue will make this into a production UI.

This is only in the example viewer. 

Only render required fields for now

This is only the first template. We will use this ticket to make a decision (ticket no.) on whether we will proceed in this manner for other templates.

blairlyons:
I started XML templates for Smoldyn, I'm not sure if this is the best way, so we should discuss:
https://github.com/simularium/simulariumio/tree/feature/ui-templates/ui-templates

@meganrm @schoinh

blairlyons:
@meganrm UI templates for Smoldyn are ready for implementation: https://github.com/simularium/simulariumio/tree/feature/ui-templates/ui-templates

toloudis:
A/C still say "When the user clicks a submit button, the file is autoconverted and loads in the viewer" - is that still true for this ticket?

PR #274, opened 2022-06-29
Feature/UI template

meganrm:
Problem
=======
What is the problem this work solves, including
#237 

Solution
========
Load the templates from the simulariumio package. This renders the form and submits the values, but I am not yet unpacking those values, so it will have no effect on the trajectory you view. 

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. Start the test app
2. Click the "load smoldyn" button 
3. Select the `example_3D.txt` from the simulariumio test data 
4. Leave the form blank, click submit 

Screenshots (optional):
-----------------------
Show-n-tell images/animations here


meganrm:
@toloudis @blairlyons this is ready to be looked over, ie, see if we like both this and the template schema setup

examples/Viewer.tsx
```diff
@@ -208,17 +222,83 @@ class Viewer extends React.Component<{}, ViewerState> {
...
+                Promise.all(
+                    map(fileRefs, (ref) =>
+                        fetch(ref.download_url).then((data) => data.json())
+                    )
+                )
+            );
+        const typeMap = reduce(
+            customTypes,
+            (acc, cur) => {
+                const key = Object.keys(cur)[0];
```
toloudis, line 252: what does Object.keys(cur) look like?  giving customTypes a type (or a comment describing what's expected) would help understand this code

examples/Viewer.tsx
```diff
@@ -208,17 +222,83 @@ class Viewer extends React.Component<{}, ViewerState> {
...
+                acc[key] = cur[key];
+                return acc;
+            },
+            {}
+        );
+        baseTypes["base_types"].forEach((type) => {
+            typeMap[type.id] = { ...type, isBaseType: true };
+        });
+        return typeMap;
```
toloudis, line 261: would love to see a type description for typeMap to know what it's supposed to look like

examples/ConversionForm/BaseInput.tsx
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+
+interface BaseInputProps {
+    dataType: string;
+    handler: (event) => void;
+    options?: [];
+    name: string;
+}
+const BaseInput = (props: BaseInputProps) => {
+    switch (props.dataType) {
```
toloudis, line 10: my only concern here is that this switch stuff is going to get worse as you add functionality to the component like validation etc.   I would lean toward just making a component type for each dataType.

toloudis: Still, at some level you will need to map a string datatype to a component. so this is a start

toloudis:
This looks like a great start. Couple high level comments:
1. I'm not sure why exactly it should be implemented in here rather than in the website repo.  It seems that this repo shouldn't have any responsibilities for showing these forms.
2. It would be nice to do a verbal walkthrough of the main logic for rendering
3. I wonder if things would be simpler if the "compound" built in types were just their own components too, rather than forcing them to build on the other primitives

meganrm:
> This looks like a great start. Couple high level comments:
> 
> 1. I'm not sure why exactly it should be implemented in here rather than in the website repo.  It seems that this repo shouldn't have any responsibilities for showing these forms.
- this is a prototype to see if we could get everything hooked up without having to mess with the main repo until we feel good about it. 

> 2. It would be nice to do a verbal walkthrough of the main logic for rendering
> 3. I wonder if things would be simpler if the "compound" built in types were just their own components too, rather than forcing them to build on the other primitives

examples/ConversionForm/CollectionInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+class CollectionInput extends React.Component<CollectionInputProps> {
+    constructor(props: CollectionInputProps) {
+        super(props);
+        this.state = {};
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange(key, index, type, targetValue) {
```
toloudis, line 22: this function is kind of begging for unit tests as documentation of how it's supposed to work in practice

examples/ConversionForm/InputSwitch.tsx
```diff
@@ -0,0 +1,103 @@
...
+                    parameter={currentParameter}
+                    templateData={templateData}
+                    dataType={currentDataType}
+                    handler={handler}
+                    path={[...path, key]}
+                />
+            );
+        }
+    };
+    const hasChildren = templateData[dataType].parameters!!;
```
toloudis, line 64: is postfix `!` what you want here?

examples/ConversionForm/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+class InputForm extends React.Component<InputFormProps> {
+    constructor(props: InputFormProps) {
+        super(props);
+        this.state = {};
+        this.handleChange = this.handleChange.bind(this);
+        this.handleSubmit = this.handleSubmit.bind(this);
+    }
+
+    handleChange(path, key, value) {
```
toloudis, line 25: again would be great to have unit tests and typings for this if possible, to help understand how it's supposed to work.

examples/Viewer.tsx
```diff
@@ -69,6 +77,27 @@ interface ViewerState {
...
+    id: string;
+    data: string;
+    match: string;
+}
+
+interface CustomType {
+    [key: string]: {
+        "python::module": string;
+        "python::object": string;
+        parameters: { [key: string]: any };
```
toloudis, line 99: Do we really know more about parameters than this?  It would be cool if you could get to something like this, for example: `{ [key:string] : BaseType | CustomType}`

toloudis (approved):

meganrm:
- [ ] Add file input as one of the switch options
- [ ] Add more typings 
- [ ] Make list extendable

examples/ConversionForm/CollectionInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+        const index = this.state.length + 1;
+        this.setState({ length: index });
+    }
+
+    render() {
+        const { name, path, parameter } = this.props;
+        const jsx = [] as JSX.Element[];
+        for (let index = 0; index < this.state.length; index++) {
+            jsx.push(
+                <div>
```
frasercl, line 120: should this element take a unique `key` property to go in an array? as in `<div key={index}>`

examples/ConversionForm/CollectionInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+                            path,
+                            index,
+                            "value"
+                        )}
+                    </div>
+                </div>
+            );
+        }
+        if (parameter.extendible) {
+            jsx.push(<button onClick={this.addItem}>Add another {name} item</button>);
```
frasercl, line 146: same as above - I believe this element should also get a unique `key`.

frasercl (approved):
LGTM, though check that React isn't complaining about a lack of keys in `CollectionInput`.

Issue #275, opened 2022-02-25
Implement ISimulator with editable state


blairlyons:
## Use Case
As a developer, I want to see a local javascript simulation running live in the Simularium viewer with a button to change the position of an agent, in order to have a proof of concept for viewing and modifying live simulations.

## Acceptance Criteria
A ISimulator simulation running locally can be visualized in a Simularium example viewer client also running locally. Can click a button to set an agent's position live. Must use an interface that can be used over the network as well, with support for play, pause, go to frame index (maybe go to time value), and updateState.

## Details
implement a javascript `ISimulator` with editable parameters

implement updateAgentPosition(id, position)

In this case there need not be any network communication, but the parameter change must demonstrate a prototypical interface that would work across the network as well.

ISimulator needs to support: play, pause, go to frame index (maybe go to time value), and updateState in order to be demonstrated.

This will build toward implementation for click and drag object later

blairlyons:
@toloudis do you want to merge this in order to be done? If so, we're getting close to end of sprint if you want to open a PR

toloudis:
https://github.com/simularium/simularium-viewer/pull/279

PR #279, opened 2022-07-22
Feature/proto live mode

toloudis:
Problem
=======
We wanted to get a quick demo of a typical live mode interaction, such as picking up an agent and moving it spatially.
Addresses #275 

Solution
========
Using the already existing ClientSimulator interface, we add methods to send data updates from viewer to server, including a new Net Message ID that just passes json data.  Entire implementation is done within the browser, using the ClientSimulator as a proxy for our remote server.
A button is added to the test viewer interface to translate agents every time it is clicked.

## Type of change

Change summary:
---------------
The ISimulator file has been shuffled a bit just to start to group methods more logically.

Steps to Verify:
----------------
1. npm start
2. select the test live points demo
3. Load, and press Play
4. click the Translate agent and see agents move while playback is occurring

### DIFF
--- a/examples/PointSimulatorLive.ts
+++ b/examples/PointSimulatorLive.ts
@@ -65,9 +65,9 @@ export default class PointSimLive implements IClientSimulatorImpl {
         // const agents = data["agents"];
         //        for (const agent of agents) {
         for (let ii = 0; ii < this.nPoints; ++ii) {
-            this.pointsData[ii * 3 + 0] += 5; //agentpos[0];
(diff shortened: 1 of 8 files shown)


examples/PointSimulatorLive.ts
```diff
@@ -0,0 +1,172 @@
+import {
```
toloudis, line 1: this whole file is basically a copy of PointSimulator but with the update state function implemented.

blairlyons (approved):

Issue #283, opened 2022-08-02
Show all and hide all are not working


meganrm:
## Description
On the test bed app the hide all and show all button is not working 

## Expected Behavior

it should be able to unselect all the agents

## Reproduction

Load a file, click hide all 

## Environment

_Any additional information about your environment. E.g. OS version, python version, browser etc._

PR #284, opened 2022-08-02
fix the "Hide All" in the demo viewer

toloudis:
Problem
=======
when working with test scenes with large numbers of agent types it's convenient to use "hide all".  However "Hide All" has been broken for a while.

Solution
========
Use the selection state data structure already present in the app to do the hiding.

## Type of change

### DIFF
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -506,6 +506,11 @@ class Viewer extends React.Component<{}, ViewerState> {
                                 }
                                 value={id}
                                 defaultChecked={true}
+                                checked={
+                                    this.state.selectionStateInfo.hiddenAgents.find(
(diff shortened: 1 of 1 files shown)


toloudis:
fixes #283

examples/Viewer.tsx
```diff
@@ -603,7 +622,6 @@ class Viewer extends React.Component<{}, ViewerState> {
                         )}
                         loadInitialData={true}
                         agentColors={this.state.agentColors}
-                        hideAllAgents={this.state.hideAllAgents}
```
toloudis, line 606: future work: remove this prop from the component?  I don't think we would use it when we can do it using the Selection interface instead.

meganrm (approved):

PR #285, opened 2022-09-13
increase number of pdb atoms for levels of detail

toloudis:
Problem
=======
A couple of pdb molecules have a lower density of atoms, making them look significantly worse at low LOD.

Solution
========

Adjust the global settings for how much decimation we do.  Also adjust distance thresholds to increase amount of detail.

There are several parameters we can tweak to control this.  This is one of the very simplest changes.  We could also clamp numbers of atoms (say, LOD 1 should be 1/4 the number of atoms but can have no more than X atoms and no less than Y atoms).  Furthermore, we could have this depend on average density (number of atoms per bounding box volume).   The distance values where the LODs change is also very global and hardcoded and doesn't change with scene dimensions.

### DIFF
--- a/src/visGeometry/index.ts
+++ b/src/visGeometry/index.ts
@@ -802,7 +802,7 @@ class VisGeometry {
                         visAgent.agentData.instanceId,
                         visAgent.signedTypeId(),
                         // a scale value for LODs
-                        (index + 1) * 0.25
+                        0.25 + index * 0.25
                     );
                     break;
                 }
--- a/src/visGeometry/index.ts
+++ b/src/visGeometry/index.ts
@@ -235,7 +235,7 @@ class VisGeometry {
         this.currentSceneAgents = [];
         this.colorsData = new Float32Array(0);
         this.lodBias = 0;
-        this.lodDistanceStops = [40, 100, 150, Number.MAX_VALUE];
+        this.lodDistanceStops = [100, 200, 400, Number.MAX_VALUE];
         this.agentsWithPdbsToDraw = [];
         this.agentPdbsToDraw = [];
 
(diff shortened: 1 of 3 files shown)


src/visGeometry/index.ts
```diff
@@ -802,7 +802,7 @@ class VisGeometry {
                         visAgent.agentData.instanceId,
                         visAgent.signedTypeId(),
                         // a scale value for LODs
-                        (index + 1) * 0.25
+                        0.25 + index * 0.25
```
toloudis, line 805: this isn't really a change but it makes it a little easier to tweak by starting with a level 0 radius of 0.25 and then an additive term based on LOD.

meganrm:
Do you mean for this to still be a draft?

toloudis:
The next step was to make the LOD distance thresholds into parameters we can edit, and then it can become a full PR.  I think the plan was to set them up to make the blood plasma trajectory look good and then fix them at those distances until we are ready to do something more general.

src/visGeometry/rendering/PDBGBufferShaders.ts
```diff
@@ -87,7 +87,9 @@ layout(location = 2) out vec4 gPos;
 
         // uncomment the following line to test LOD.  IN_radius is a measure of lod.
         //gAgentInfo = vec4(IN_radius*4.0, IN_instanceAndTypeId.x, fragViewPos.z, fragPosDepth);
+
         gAgentInfo = vec4(IN_instanceAndTypeId.y, IN_instanceAndTypeId.x, fragViewPos.z, fragPosDepth);
+        
         gNormal = vec4(normal * 0.5 + 0.5, 1.0);
```
toloudis, line 93: swap the commenting of the above two lines to do very simple local testing of LOD levels on pdb geometries.  it will color the pdbs according to their discrete LOD level

toloudis:
TODO:  generalize these settings for scenes with different bounding box sizes?

meganrm (approved):
Looks a lot better

blairlyons (approved):

PR #289, opened 2022-10-03
Fix/workflows

meganrm:
Problem
=======
1. Like our other repos, we had redundant actions in our workflows. 
3. We want to be able to install the development version of the app for testing: https://github.com/simularium/simularium-website/issues/352

Solution
========
Consolidated workflows and added a development publish to npm

## Type of change


toloudis (approved):

toloudis (approved):
Can you just keep overwriting the same dev build over and over on npm?

meganrm:
> Can you just keep overwriting the same dev build over and over on npm?

My understanding is that it will still have the version number, but plus a tag. So yes and no, in between releases it will keep overwriting the same dev build, but each release will have one dev build

blairlyons (approved):

PR #290, opened 2022-10-07
Fix/worker loading

toloudis:
Problem
=======

web workers failing to load in simularium-website

Solution
========

This is half of the solution.  The other half will be in some updates in simularium-website.

1. This library's build is updated to remove all other bundling apart from transpiling to es modules.
The lib and dist directories are removed from package.json as they are not being used at all.

2. web worker is moved into a subdirectory and receives a custom typescript config file so that it can use the Workers typescript lib.


toloudis:
fixes https://github.com/simularium/simularium-website/issues/366

meganrm (approved):

PR #292, opened 2022-11-18
Feature/send smoldyn data to websocket

ascibisz:
Problem
=======
Originally, we thought we were going to send the result object of the Smoldyn form to an AWS lambda via a HTTP request. We have since decided to do things differently and instead send this via a websocket message to Octopus, which will handle the autoconversion of the Smoldyn object. This PR makes the relevant changes so that the form sends the smoldyn trajectory in the expected way.

[Off shoot of this ticket](https://github.com/simularium/simularium-planning/issues/92)

Solution
========
* Changed up the callback from submitting the `ConversionForm` so that it'll send a websocket request
* Created a new message type, and at first I just added the message handling for autoconversion in with `RemoteSimulator`. Since this is autoconversion and inherently not a simulation, that felt bad so on Dan's suggestion, I created a `IConverter` interface.
* Rather than just have the `IConverter` implemented just in `RemoteSimulator` alongside `ISimulator` (which I did at first, if you look back in earlier commits), I built out a new `WebsocketClient` class and brought the logic for handling websocket connections there, so that we could have `RemoteSimulator` and a new `RemoteConverter` use the same `WebsocketClient`

## Type of change

Open Questions
-----------------------
1. Should I have a `IWebsocketClient` interface? I couldn't decide if I was supposed to or not, so I didn't for now, but it's easy enough to add
2. Is there a less convoluted way for me to handle all the callbacks associated with submitting the smoldyn form?? I did something that works, but if there's some cleaner way to do it please let me know
3. I should probably write a test or something, shouldn't I?

This is the most TypeScript I've ever written! Please help me make this look like proper TypeScript 😅


src/simularium/WebsocketClient.ts
```diff
@@ -0,0 +1,317 @@
...
+//     fileName: string; // identifies the trajectory this connection is dealing with
+//     payload: Object; // the JS object with the message data itself
+// }
+
+export interface MessageEventLike {
+    data: string;
+}
+
+// these have been set to correspond to backend values
+export const enum NetMessageEnum {
```
toloudis, line 26: Half baked idea: 
I wonder if the message ids should ultimately live inside the RemoteSimulator etc, and just keep this class for managing the connection state.  The json and binary message handlers would be much more generic then.  And anyone who wants to add a handler can have their own message interpreter.  You could still have this idea where each handler can either say "pass this message along to other handlers" or "stop, we have completely handled it" by returning a boolean or some other value type.

src/simularium/RemoteConverter.ts
```diff
@@ -0,0 +1,78 @@
...
+import jsLogger from "js-logger";
+
+export class RemoteConverter implements IConverter {
+    protected logger: ILogger;
+    public loadFile: (NetMessage) => void;
+    public handleError: (error: FrontEndError) => void | (() => void);
+    private webSocketClient: WebsocketClient;
+
+    public constructor(
+        webSocketClient: WebsocketClient,
```
toloudis, line 15: it could be very interesting to create an entirely separate websocket client instance that connects to the same server! (rather than sharing the same connection...)  We might get a bit more asynchronous/parallelism by doing that and it could keep the message passing cleaner.  I'm not sure if there would be any significant overhead.

src/simularium/RemoteSimulator.ts
```diff
@@ -120,244 +65,124 @@ export class RemoteSimulator implements ISimulator {
...
-            this.webSocket.readyState === this.webSocket.CLOSED
-        );
-    }
-
-    private socketIsConnected(): boolean {
-        return (
-            this.webSocket !== null &&
-            this.webSocket.readyState === this.webSocket.OPEN
-        );
+        return this.webSocketClient.socketIsValid();
```
toloudis, line 69: who calls socketIsValid? could we remove this?

src/simularium/WebsocketClient.ts
```diff
@@ -0,0 +1,317 @@
...
+        // where we receive websocket messages
+        if (!this.socketIsValid()) {
+            return;
+        }
+
+        if (event.data instanceof ArrayBuffer) {
+            // Handle binary message
+            const floatView = new Float32Array(event.data);
+            const binaryMsgType = floatView[0];
+            for (let msgType in this.binaryMessageHandlers) {
```
toloudis, line 151: instead of a for loop, I wonder if this could be written as something like
```
if this.binaryMessageHandlers.has(msgType)
  this.binaryMessageHandlers[msgType](event)
```

examples/Viewer.tsx
```diff
@@ -311,6 +312,17 @@ class Viewer extends React.Component<{}, ViewerState> {
         });
     }
 
+    public convertFile(obj: Record<string, any>, fileType: string) {
+        simulariumController.configureNetwork(netConnectionSettings);
```
toloudis, line 316: I can't remember what configureNetwork does.  is it something that might have async failure states such that you want to do 
controller.configureNetwork().then(controller.convertTrajectory())...

src/simularium/WebsocketClient.ts
toloudis: at the same time, there is something very nice and simple about just adding handlers for each known message type right here...

src/simularium/WebsocketClient.ts
```diff
@@ -0,0 +1,317 @@
...
+export class WebsocketClient {
+    private webSocket: WebSocket | null;
+    private serverIp: string;
+    private serverPort: number;
+    public connectionTimeWaited: number;
+    public connectionRetries: number;
+    protected jsonMessageHandlers: Map<NetMessageEnum, (NetMessage) => void>;
+    protected binaryMessageHandlers: Map<
+        NetMessageEnum,
+        (MessageEvent) => void | ((MessageEventLike) => void)
```
toloudis, line 69: Instead of having `MessageEvent | MessageEventLike` in many different places, I think what would look cleaner here is to define a type called `MessageEventLike` that is `MessageEvent | {data:string}`, and then use MessageEventLike everywhere.
You could name the {data:string} as SimulariumEvent or something like that.

toloudis:
I left a few suggestions, most of them are not super critical.  There are also a bunch of lint warnings to clean up.  I think the new WebSocketClient class is a really good and important refactoring.

src/simularium/RemoteSimulator.ts
ascibisz: it's called in `configureNetwork()` and `disableNetworkCommands()` inside `SimulariumController`. This was mostly me just trying to maintain the existing functionality from before as I refactored everything, I don't 100% understand everything that the `SimulariumController` does so I can't tell how important it is 😬

src/simularium/WebsocketClient.ts
```diff
@@ -150,18 +150,15 @@ export class WebsocketClient {
...
             const floatView = new Float32Array(event.data);
             const binaryMsgType = floatView[0];
-            for (let msgType in this.binaryMessageHandlers) {
-                if (binaryMsgType == Number(msgType)) {
-                    this.binaryMessageHandlers[msgType](event);
-                    // TODO: is it ok if we have multiple handlers for the same
-                    // message type? should we return here? for now, we will
-                    return;
-                }
+            if (binaryMsgType in this.binaryMessageHandlers) {
```
toloudis, line 153: I like this change and looking at it, it looks like a great candidate for some kind of unit test.

src/simularium/WebsocketClient.ts
```diff
@@ -150,18 +150,15 @@ export class WebsocketClient {
...
+                this.binaryMessageHandlers[binaryMsgType](event);
+            } else {
+                this.logger.error(
+                    "Unexpected binary message arrived of type ",
+                    binaryMsgType
+                );
```
toloudis, line 159: Is this really an error?  Either we got a message id that is bad or we are possibly intentionally not handling it by virtue of never having added a handler.  Am I overthinking this?

ascibisz: Is it a reasonable case that we'd receive messages that we haven't implemented a handler? Like should that not be an error?

I don't have particularly strong opinions, I'd happily just console log it to indicate that we received and didn't do anything with it.

examples/Viewer.tsx
ascibisz: Nope, `configureNetwork()` just returns void, and throws an error if it runs into one. `configureNetwork()` is also called elsewhere in this file and it looks like what I have (with no `.then()`)

src/simularium/WebsocketClient.ts
ascibisz: ok so I just wrote out some unit tests for this, only to realize that I'd need to change `onMessage` to be `public` instead of `protected` for the tests to run... how do we feel about that?

ascibisz: Gah I stand corrected, the typechecker doesn't like that I'm trying to call `protected onMessage()` from my test, is there a hacky way around that or should I change `onMessage` to public or do you potentially have a different suggestion? @toloudis

src/simularium/WebsocketClient.ts
toloudis: reasonable point that we should always handle all messages and it's an error if we don't.  I can be on board with that.  Maybe the code could enforce that even more strongly, then.

src/simularium/WebsocketClient.ts
toloudis: One way is to do this: in your test make a class that derives from the main class, and add a public function that calls the protected one.  It's kind of a nasty workaround but I bet it would do the trick.

```
class TestMyClass extends MyClass {
  public exposeProtectedFuncForTesting() { return protectedFunc(); }
}
```

toloudis: I guess I'm not strictly opposed to making onMessage public, either.  Another way is to refactor the code to make the onMessage in its own little class where it makes more sense to be public.

src/controller/index.ts
```diff
@@ -133,13 +139,21 @@ export default class SimulariumController {
...
             this.simulator.setTrajectoryDataHandler(
                 this.visData.parseAgentsFromNetData.bind(this.visData)
             );
+            this.converter = new RemoteConverter(webSocketClient, this.onError);
+            this.converter.setLoadFileHandler((result: ConvertedFileData) => {
+                const file = JSON.parse(result["simulariumData"]);
+                const simulariumFile = new JsonFileReader(file);
+                // TODO: make file name more informative
+                this.handleFileChange(simulariumFile, "test.simularium");
+            });
```
meganrm, line 156: I get why you're initializing this here, but it also seems like it could be done separately, and only if the user requests a conversion? 
Otherwise, I think it would be good to separate out this section contained in the `else if` block and call it `setUpRemoteConnection` or something, because this goes beyond `createSimulatorConnection`

src/simularium/RemoteConverter.ts
meganrm: I agree with this, the code to set up the simulator and the convertor seems repetitive as it is now

src/controller/index.ts
```diff
@@ -288,6 +312,22 @@ export default class SimulariumController {
...
+        if (!fileName.includes(".simularium")) {
+            throw new Error("File must be a .simularium file");
+        }
+
+        if (geoAssets && geoAssets.length) {
+            return this.changeFile({ simulariumFile, geoAssets }, fileName);
+        } else {
+            return this.changeFile({ simulariumFile }, fileName);
+        }
+    }
```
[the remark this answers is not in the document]
ascibisz: Ok so I just looked and I took this code from `examples/Viewers.tsx` `loadFile()`.  I'll make that function call this one and remove the redundant code

src/simularium/RemoteConverter.ts
ascibisz: Alrighty! I'm not sure if there's any significant overhead to this change either, but I'll give it a whirl and we'll see how it goes

src/simularium/WebsocketClient.ts
ascibisz: My only concern with moving these is making sure the values stay the same, since it's important for them to match with what octopus has. Like is it confusing if we define separate enums in separate files that are connecting to the WebsocketClient, ie in RemoteConverter we have something like
```
const enum ConverterEnum {
    ID_CONVERT_TRAJECTORY_FILE = 16,
    ID_CONVERTED_TRAJECTORY = 17,
}
```
Like is it weird to have this on its own separate from enums numbered 1-15 and 18+? Or am I overthinking this...

src/simularium/WebsocketClient.ts
ascibisz: It's an interesting question of how best to handle unknown messages. Should we have some sort of error message type that we send back to the server in that case? Since we own both pieces, we could log the error in both places, and hopefully eventually attach some monitoring to it (once all this is in production).

It might be worth having a larger conversation on error handling, because as we start sending large trajectories as multiple packets and come up with some sort of progress measuring, we're just going to get more errors. It'd be good to clearly define the error cases in which we retry vs abort vs alert the user, and where we want to log all this.

src/simularium/types.ts
```diff
@@ -129,3 +129,8 @@ export interface FileReturn {
     status: FileStatus;
     message?: string;
 }
+
+export interface ConvertedFileData {
+    msgType: number;
+    simulariumData: string;
```
toloudis, line 135: converter only returns strings (aka json data)?    ultimately binary, right?

toloudis (approved):

src/simularium/types.ts
ascibisz: correct, at least for now

src/controller/index.ts
```diff
@@ -233,14 +219,25 @@ export default class SimulariumController {
...
+            this.configureNetwork(netConnectionConfig);
+            if (!(this.simulator instanceof RemoteSimulator)) {
+                throw new Error("Autoconversion requires a RemoteSimulator");
+            }
+        } catch (e) {
+            return Promise.reject(e);
         }
-        return this.converter.convertTrajectory(obj, fileType);
+
+        return this.simulator.convertTrajectory(obj, fileType).then(() => {
```
ascibisz, line 235: TODO: check the status of this potential race condition, at least leave a comment indicating the potential

examples/Viewer.tsx
```diff
@@ -311,6 +312,16 @@ class Viewer extends React.Component<{}, ViewerState> {
         });
     }
 
+    public convertFile(obj: Record<string, any>, fileType: string) {
```
ascibisz, line 315: we should add a specific type for obj here

toloudis: To clarify, if obj is known to consist of form data plus file blob, then it makes sense to have a type for that.

src/controller/index.ts
toloudis: if it's not easy to create the race condition now, then it might make sense to defer it to a new issue/PR that is resolved soon-ish.  one thing we used to talk about a lot is having every message contain the id of the simulation it's dealing with, and so you can do checks on both ends to make sure the id matches what the server or client thinks is current for the given connection.

src/controller/index.ts
```diff
@@ -214,6 +218,27 @@ export default class SimulariumController {
         }
     }
 
+    public convertAndLoadTrajectory(
+        netConnectionConfig: NetConnectionParams,
+        obj: Record<string, unknown>,
```
meganrm, line 223: in addition to typing this, I would call it something clearer, like `dataToConvert` or `trajectoryPayload` (open to suggestions :)

src/controller/index.ts
```diff
@@ -214,6 +218,27 @@ export default class SimulariumController {
...
+        fileType: string
```
meganrm, line 224: we mentioned that it would be good to have this typed as an enum that currently has just one item, but to make it clear that fileType has to be one of our specified engines

meganrm (approved):

PR #293, opened 2022-11-29
Feature/dev build

meganrm:
Problem
=======
#221 
I previously tried using npm tags to get a build on merge, but the tags are not sub versions as I thought, so you can't publish them without also versioning the app. 

Solution
========
This builds the app, copies everything needed into one folder, and then pushes that to the `build` branch. I've tested this by turning off the conditional that only builds on main, so if you look in the recent actions you can see how it works. 

## Type of change

Steps to Verify:
----------------
1. [action that ran](https://github.com/simularium/simularium-viewer/actions/runs/3595831276)
2. check the build artifact
3. check branch `build`


.github/workflows/ci.yml
```diff
@@ -34,32 +34,53 @@ jobs:
...
+            - run: mkdir dev-build
+            - run: |
+                cp -r es dev-build | 
+                cp -r lib dev-build | 
+                cp -r dist dev-build |
+                cp -r style dev-build |
+                cp -r type-declarations dev-build |
+                cp README.md dev-build |
+                cp package.json dev-build
+            - uses: actions/upload-artifact@v3
```
toloudis, line 65: is upload-artifact being used by the next step (push-to-branch-action)?  I can't tell why this step is needed

.github/workflows/ci.yml
```diff
@@ -34,32 +34,53 @@ jobs:
...
             - run: npm run build
+            - run: mkdir dev-build
+            - run: |
+                cp -r es dev-build | 
+                cp -r lib dev-build | 
+                cp -r dist dev-build |
+                cp -r style dev-build |
+                cp -r type-declarations dev-build |
+                cp README.md dev-build |
+                cp package.json dev-build
```
toloudis, line 64: all these cp lines look a bit clunky - it would be cool if they were encapsulated in a `npm run copy_build_artifacts` script so that this file doesn't have to know about all those directories.

.github/workflows/ci.yml
meganrm: no it's not. I can delete this, I was using to confirm the directory was correct before I added the push to gh step

.github/workflows/ci.yml
meganrm: I'll try using this: https://www.npmjs.com/package/copyfiles

meganrm: ok I got it to work. there are a lot of commits because I mistakenly thought the dist and lib folders needed to be in there too and I couldn't figure out why they were getting copied

toloudis:
The contents of https://github.com/simularium/simularium-viewer/tree/build/es look wrong to me.  Am I looking in the right place?  Also are you sure the copy is recursive wrt subdirectories and everything?

meganrm:
> The contents of https://github.com/simularium/simularium-viewer/tree/build/es look wrong to me. Am I looking in the right place? Also are you sure the copy is recursive wrt subdirectories and everything?

thanks for checking, yeah I was missing the sub directories, they are there now. I needed to change 'es/*' to 'es/**/*'

toloudis:
something still wrong with the test files both in the es/ dir and inside of es/test... they are supposed to be excluded from the transpileES step so I'm not sure how they got here into your build/ dir.  Otherwise this is looking great

toloudis (approved):

meganrm:
> something still wrong with the test files both in the es/ dir and inside of es/test... they are supposed to be excluded from the transpileES step so I'm not sure how they got here into your build/ dir. Otherwise this is looking great

That seems to be happening if I run `npm run build` so it's not a problem with this new code. I made a new issue for it

ascibisz (approved):

PR #302, opened 2023-04-06
Update websocket messaging for octopus

ascibisz:
Problem
=======
We updated the protocol of the messaging Octopus expects for kicking off a new simulation, so this PR is meant to update the simularium-viewer to be compatible.  
[Octopus Ticket](https://github.com/simularium/octopus/issues/26)

Solution
========
This change diverges slightly from the communication protocol used in simularium-engine, so in order to preserve functionality with simularium-engine backed for other developers, I wanted to add some sort of flag/gate so we can have both communication protocols available in the viewer, ideally that developers can just define in the commandline when starting up the simularium-viewer. After poking around for a bit and reading some webpack documentation, I came up with _a_ way to do this. This is out of my area of expertise, so feel free to let me know if you have suggestions of better ways to do this!

While I was coming up with a way to gate this octopus communication protocol, I figured I may as well do the same thing for running local backend vs remote, since I'm tired of going into the code and manually switching the net config params. So now here's what we've got:

`npm run start` will run the simularium-viewer as currently expected, pointed to the remote simularium-engine server
`npm run start --octopus` will run simularium-viewer pointing to the remote octopus EC2 instance (which is currently down, next on my todo list is to figure out why)
`npm run start --localserver` will run the viewer expecting that you have a server running locally at `0.0.0.0:8765` for handling the backend, with the simularium-engine messaging protocol (although I strongly doubt anyone's running the simularium-engine code locally these days)
`npm run start --octopus --localserver` will run the viewer and communicate with your locally running octopus server for the backend

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,8 @@ This will run the example in `/examples/Viewer.tsx`, demonstrating the viewer's 
 | transpileES       | run babel on `src` directory; _do not_ transpile `import/export` statements for an ES module compatible build (used by bundlers for tree-shaking) |
 | test              | run `jest`; searches for any files matching the pattern "src/\*_/_.test.js"                                                                      |
 | typeCheck         | run `tsc` in type-check only mode                                                                                                                 |
-| start             | runs an example app from `examples` for testing. Runs at `localhost:8080/public/`.                                                                |
+| start             | runs an example app from `examples` for testing. Runs at `localhost:8080/public/`. Use `--octopus` to connect to octopus backend and/or           |
+|                   | `--localserver` to run backend locally. With no flags, this script will default to using remote simularium-engine as backend                      |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -118,8 +118,7 @@ This will run the example in `/examples/Viewer.tsx`, demonstrating the viewer's 
 | transpileES       | run babel on `src` directory; _do not_ transpile `import/export` statements for an ES module compatible build (used by bundlers for tree-shaking) |
 | test              | run `jest`; searches for any files matching the pattern "src/\*_/_.test.js"                                                                      |
 | typeCheck         | run `tsc` in type-check only mode                                                                                                                 |
-| start             | runs an example app from `examples` for testing. Runs at `localhost:8080/public/`. Use `--octopus` to connect to octopus backend and/or           |
-|                   | `--localserver` to run backend locally. With no flags, this script will default to using remote simularium-engine as backend                      |
+| start             | runs an example app from `examples` for testing. Runs at `localhost:8080/public/`. Use `--octopus` to connect to octopus backend and/or `--localserver` to run backend locally. With no flags, this script will default to using remote simularium-engine as backend |
 
 ---
 
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -31,10 +31,7 @@ import ConversionForm from "./ConversionForm";
 import MetaballSimulator from "./MetaballSimulator";
 import { TrajectoryType } from "../src/constants";
 
-const netConnectionSettings = {
-    // to test local server: (also may have to change wss to ws in the url)
-    // serverIp: "0.0.0.0",
-    // serverPort: 8765,
+let netConnectionSettings = {
     serverIp: "staging-node1-agentviz-backend.cellexplore.net",
     serverPort: 9002,
 };
@@ -114,12 +111,17 @@ interface CustomType {
     };
 }
 
-const simulariumController = new SimulariumController({});
+interface InputParams {
+    localBackendServer: boolean;
+    useOctopus: boolean;
+}
+
+let simulariumController;
 
 let currentFrame = 0;
 let currentTime = 0;
 
-const initialState = {
+const initialState: ViewerState = {
     renderStyle: RenderStyle.WEBGL2_PREFERRED,
     pauseOn: -1,
     particleTypeNames: [],
@@ -145,8 +147,10 @@ class Viewer extends React.Component<{}, ViewerState> {
     private viewerRef: React.RefObject<SimulariumViewer>;
     private panMode = false;
     private focusMode = true;
+    private useOctopus = false;
+    private localBackendServer = false;
 
-    public constructor(props) {
+    public constructor(props: ViewerState & InputParams) {
         super(props);
         this.viewerRef = React.createRef();
         this.handleJsonMeshData = this.handleJsonMeshData.bind(this);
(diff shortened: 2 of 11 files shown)


webpack.dev.js
```diff
@@ -26,6 +27,10 @@ module.exports = {
                 },
             ],
         }),
+        new webpack.DefinePlugin({
+            SIMULARIUM_USE_OCTOPUS: JSON.stringify(Boolean(process.env.npm_config_octopus)),
+            SIMULARIUM_USE_LOCAL_BACKEND: Boolean(process.env.npm_config_localserver),
```
ascibisz, line 32: todo: make these lines the same

src/controller/index.ts
```diff
@@ -102,6 +109,8 @@ export default class SimulariumController {
             }
         }
 
+        console.log("controller oct: " + this.useOctopus);
+        console.log("controller server: " + this.localBackendServer);
```
ascibisz, line 113: remove console logs

src/simularium/WebsocketClient.ts
```diff
@@ -100,6 +103,9 @@ export class WebsocketClient {
         this.logger = jsLogger.get("netconnection");
         this.logger.setLevel(jsLogger.DEBUG);
 
+        // Use wss when connected to simularium-engine, use ws otherwise
+        this.secureConnection = !(useOctopus || localBackendServer);
```
toloudis, line 107: suggestion: add "secureConnection" boolean to NetConnectionParams instead of passing useOctopus and localBackendServer in to this class.

src/controller/index.ts
```diff
@@ -142,10 +151,16 @@ export default class SimulariumController {
...
                 netConnectionConfig,
+                this.useOctopus,
+                this.localBackendServer,
                 this.onError
             );
             this.remoteWebsocketClient = webSocketClient;
-            this.simulator = new RemoteSimulator(webSocketClient, this.onError);
+            this.simulator = new RemoteSimulator(
+                webSocketClient,
+                this.useOctopus,
```
toloudis, line 161: i wonder if useOctopus could be passed in to this createSimulatorConnection function (maybe even as part of NetConnectionParams) since it's not really needed for WebsocketClient, just for RemoteSimulator creation.

ascibisz: I just moved useOctopus and secureConnection to NetConnectionParams, it looks cleaner and they are both inherently parameters related to the net connection. Thanks for the suggestion!

examples/Viewer.tsx
```diff
@@ -155,6 +159,24 @@ class Viewer extends React.Component<{}, ViewerState> {
...
+                useOctopus: props.useOctopus,
+                secureConnection: false,
+            };
+        } else if (props.useOctopus) {
+            netConnectionSettings = {
+                serverIp: "18.223.108.15",
+                serverPort: 8765,
+                useOctopus: props.useOctopus,
+                secureConnection: false,
+            };
```
toloudis, line 176: maybe could put the other server settings here in a final else just to keep it all together in one place.

src/controller/index.ts
```diff
@@ -145,7 +145,13 @@ export default class SimulariumController {
                 this.onError
             );
             this.remoteWebsocketClient = webSocketClient;
-            this.simulator = new RemoteSimulator(webSocketClient, this.onError);
+            this.simulator = new RemoteSimulator(
+                webSocketClient,
+                netConnectionConfig.useOctopus
```
toloudis, line 150: I think what you are trying to do here is `!!netConnectionConfig.useOctopus` -- does that seem right?

toloudis (approved):
Looks good and I can't wait till we can switch over to octopus entirely.  If we start to incur more changes to the messaging protocols, I can imagine just having a whole separate `RemoteOctopusSimulator` class.

src/controller/index.ts
ascibisz: - [ ] change to !!netConnectionConfig.useOctopus

examples/Viewer.tsx
```diff
@@ -114,12 +107,17 @@ interface CustomType {
     };
 }
 
-const simulariumController = new SimulariumController({});
+interface InputParams {
+    localBackendServer: boolean;
+    useOctopus: boolean;
+}
+
+let simulariumController;
```
frasercl, line 115: Looks like `simulariumController` can also safely become a class property of `Viewer`? If that doesn't make sense for some reason I'm missing, I'd still give it a type here: `let simulariumController: SimulariumController;`

examples/Viewer.tsx
```diff
@@ -145,8 +143,9 @@ class Viewer extends React.Component<{}, ViewerState> {
     private viewerRef: React.RefObject<SimulariumViewer>;
     private panMode = false;
     private focusMode = true;
+    private netConnectionSettings: NetConnectionParams;
 
-    public constructor(props) {
+    public constructor(props: ViewerState & InputParams) {
```
frasercl, line 148: I think this can just be `InputParams`, as only those fields are used from `props`. This can also go in the first generic argument to `React.Component` above, which is for the props type: `class Viewer extends React.Component<InputParams, ViewerState> {`

README.md
```diff
@@ -118,7 +118,8 @@ This will run the example in `/examples/Viewer.tsx`, demonstrating the viewer's
 | transpileES       | run babel on `src` directory; _do not_ transpile `import/export` statements for an ES module compatible build (used by bundlers for tree-shaking) |
 | test              | run `jest`; searches for any files matching the pattern "src/\*_/_.test.js"                                                                      |
 | typeCheck         | run `tsc` in type-check only mode                                                                                                                 |
-| start             | runs an example app from `examples` for testing. Runs at `localhost:8080/public/`.                                                                |
+| start             | runs an example app from `examples` for testing. Runs at `localhost:8080/public/`. Use `--octopus` to connect to octopus backend and/or           |
+|                   | `--localserver` to run backend locally. With no flags, this script will default to using remote simularium-engine as backend                      |
```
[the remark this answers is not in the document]
ascibisz: - [ ] deal with markdown, as this should all be one table row

examples/Viewer.tsx
ascibisz: It appears you might be right, lemme take a look. I usually just try to change as little code as possible when dealing with TypeScript, but I'll investigate this one and change it to a class property unless I can figure out why that may not be possible. At the very least, I can certainly give it a type

- [ ] try making simulariumController a class variable, or at least give it a type

examples/Viewer.tsx
ascibisz: - [ ] I'll do that!

examples/Viewer.tsx
toloudis: I think that makes sense for this app but in general I think we wanted to emphasize that the SimulariumController could be initialized and live outside of the core Viewer (as in Model-View-Controller paradigm).  Also possibly mirroring (or being an example for) the initialization sequence of simularium-website?

ascibisz: Honestly, the original reason I had to move the SimulariumController initialization was from when I had been sending useOctopus and localBackendServer directly to them, so I'll just move it back to where it was originally.

@toloudis do you know why we're importing SimulariumController from `"../es"`? Cameron and I were looking at the code yesterday afternoon and when I was trying to set `simulariumController`'s type, and my editor wasn't thrilled about it because the SimulariumController we're importing is already transpiled into javascript.

toloudis: Yes, the point of importing from `/es` is to simulate as if we were using the npm-installed version of the library.  Some issues can come up with the building and bundling of the distribution that would arise sooner from here before having to publish it.

toloudis: We should definitely export any types needed by clients of the viewer.

ascibisz: Alright, that does make sense since it's an example viewer, good to test the whole. I'm still a bit novice when it comes to TypeScript, so just to make sure, does that mean it _doesn't_ make sense to try to type the simulariumController, like should I just leave if how it is?

frasercl (approved):

PR #307, opened 2023-04-17
Fix/init canvas size

meganrm:
Problem
=======
When trying to initialize the viewer in a jupyter notebook, the parent ele had no scrollHeight/Width, so the viewer wasn't being size correctly. We could get it to size correctly by calling `VisGeometry.resize` after mount, but it would be better to have it just have the correct height to begin with. 

Solution
========
Use html data instead of trying to grab the style in javascript. The height and width are computed in the DOM, while the data attributes are constants so we don't have to worry about wether the ele has mounted yet. 
with @toloudis 

## Type of change

Steps to Verify:
----------------
1. npm start
2. the viewer should look normal, and resize normally.

### DIFF
--- a/src/viewport/index.tsx
+++ b/src/viewport/index.tsx
@@ -637,6 +637,8 @@ class Viewport extends React.Component<
                     }}
                     ref={this.vdomRef}
                     tabIndex={0}
+                    data-height={height}
+                    data-width={width}
                 >
                     {showCameraControls && this.renderViewControls()}
                 </div>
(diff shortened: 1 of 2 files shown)


frasercl (approved):

toloudis (approved):
I wish there was a nicer way of receiving that earliest size change but it's always been an issue.   TODO: see if volume-viewer /website-3d-cell-viewer should do this too, in particular in nbvv.

Issue #303, opened 2023-04-12
ambient occlusion changes with zoom


toloudis:
## Description

When zooming in or out, agents get dimmer or brighter and it is not clear why.
We suspect the screen space AO sampling.

## Expected Behavior

Basically want things to stay the same brightness against zooming.

If AO is the problem, consider scaling the spacing of AO samples with depth value.

rheiland:
Agreed. Here's an example of what I'm seeing.

<img width="608" alt="Screen Shot 2023-04-21 at 5 45 21 AM" src="https://user-images.githubusercontent.com/1774917/233605426-7a1236bd-7a3c-4590-b7d3-9cb617d70c7f.png">
<img width="454" alt="Screen Shot 2023-04-21 at 5 44 47 AM" src="https://user-images.githubusercontent.com/1774917/233605433-68b94b11-2c43-49c1-b357-3b1aaa1e7dc3.png">

blairlyons:
@toloudis Randy's test trajectory is [here](https://simularium.allencell.org/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/physicell_ambient_occlusion_test.simularium)

blairlyons:
and [another test file](https://simularium.allencell.org/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/point_cloud_test.simularium) from Ritvik, this one is dim unless you zoom out

PR #312, opened 2023-05-05
Fix/threejs types

toloudis:
Problem
=======
Beginning to work on #303 I installed the type info package for threejs and discovered all kinds of issues.  So this is an intermediate PR to address consistent typing throughout the rendering code, and in particular the GeometryCache.

Solution
========
Added the @types/three package, ran npm run typeCheck, fixed all errors.

## Type of change


interim17:
I've been shy to comment on PR's due to inexperience, but in general this looks good to me. Some of the chained typecasting is unfortunate, but you have to make the compiler happy. It runs fine on my end with the same errors as main branch. Inclined to approve.

toloudis:
> I've been shy to comment on PR's due to inexperience, but in general this looks good to me. Some of the chained typecasting is unfortunate, but you have to make the compiler happy. It runs fine on my end with the same errors as main branch. Inclined to approve.

Yeah I can't stand the awful typecasts but they are still a thing we have to live with at times.  I always figure they will get better over time. 
The interesting test here is to go back to main, just add the @types/three devDependency, and then watch the errors appear.  The GeometryStore was really surprisingly inconsistent about some things.

interim17:
Yeah that's not a good look!

src/visGeometry/GeometryStore.ts
```diff
@@ -466,7 +474,8 @@ class GeometryStore {
                 // on updatescene, add instances
                 // on render, pass the group of marchingcubes objects
                 geometry = {
-                    mesh: null,
+                    // null is only acceptable for SphereGroup
+                    mesh: null as unknown as Object3D,
```
frasercl, line 478: To clarify: the comment is explaining that it's preferable to force-cast `null` to `Object3D` rather than update this type to allow `null` because `null` should be disallowed for all but one variant of this type?

frasercl (approved):

src/visGeometry/GeometryStore.ts
toloudis: Yes exactly- in fact it might be allowable to just create a `new Object3D()` here, which would be much cleaner, but I didn't try to confirm it.

PR #313, opened 2023-05-18
Fix/interface simularium file

meganrm:
Problem
=======
For the download option on the front end, the ISimulariumFile had no public way of getting the file contents back out. 

Solution
========
Created a `getAsBlob` function to get the file contents back out. We agreed this was a simple solution for the current implementation, but in the long run we want to create a writer class for getting the data back into a file format. 

with @toloudis and @interim17 

## Type of change

Steps to Verify:
----------------
1. `npm start`
2.  load a trajectory (select one and click load)
3. click download
4. drag and drop the downloaded file back on the viewer 
5. click download again 
6. (can also check a binary file)


src/simularium/BinaryFileReader.ts
```diff
@@ -263,7 +263,17 @@ export default class BinaryFileReader implements ISimulariumFile {
...
+    }
+
     getPlotData(): Plot[] {
         return this.plotData;
     }
+
+    getAsBlob(): Blob {
+        return new Blob([this.fileContents], {
+            type: "text/plain;charset=utf-8",
+        });
```
toloudis, line 277: I would have thought this should not be using "text/plain" with utf-8...?

src/simularium/ISimulariumFile.ts
```diff
@@ -1,9 +1,22 @@
...
+    VisDataMessage,
+} from "./types";
 
 export interface ISimulariumFile {
     getTrajectoryFileInfo(): TrajectoryFileInfo;
     getPlotData(): Plot[];
     getNumFrames(): number;
     getFrameIndexAtTime(time: number): number;
     getFrame(theFrameNumber: number): VisDataFrame | ArrayBuffer;
+    getSpatialData(): VisDataMessage | DataView;
```
toloudis, line 14: I'd say let's not add getSpatialData if no one is going to use it.

toloudis (approved):
I approve all except `getSpatialData` 👍

src/simularium/BinaryFileReader.ts
interim17: Maybe this should be "application/json;charset=utf-8" instead of "text/plain;charset=utf-8"". This came from my code originally, and you're right it's probably not the right MIME type although it seemed to generate valid output with this constructor set up.

toloudis: oh I was going completely the other way in thinking it should be a binary mime type for BinaryFileReader.  As long as it works and the file can be reloaded, i'm good!

examples/Viewer.tsx
```diff
@@ -524,9 +545,30 @@ class Viewer extends React.Component<{}, ViewerState> {
...
+        const downloadLink = document.createElement("a");
+        downloadLink.download = simulariumFile.name;
+        downloadLink.style.display = "none";
+        downloadLink.href = href;
+        document.body.appendChild(downloadLink);
+        downloadLink.click();
+        document.body.removeChild(downloadLink);
+        URL.revokeObjectURL(downloadLink.href);
+    }
+
```
interim17, line 568: Having this code in the viewer here will make the front end implementation much simpler. :)

meganrm: This is just the example viewer to test that it works. This doesn't get shipped with the viewer npm package

interim17 (approved):

PR #314, opened 2023-05-19
Feature/orthographic camera

### LINKED ISSUE #305: Implement orthographic camera view
## Use Case
As a user, it is easier to eyeball measurements if view is orthographic.

## Acceptance Criteria
User can switch between ortho and perspective, in test viewer. 

## Details
Do we already have fov-angle control?  if not, then add it to the secret gui.

frasercl:
Problem
=======
Resolves #305: Users would like the option to switch between perspective and orthographic camera views.

Solution
========
- Add `perspectiveCamera` and `orthographicCamera` properties to `VisGeometry`
- Add the public method `setCameraType` to switch which of these is active in the `camera` property
- Add some private methods for conveniently dealing with switching between cameras (see below)
- Update some other methods to consider orthographic cameras (`dolly`, `storeCamera`, `loadCamera`)
- Add fov to tweakpane and make camera saving/loading/importing/exporting from there more robust
- Do some formatting and cleanup along the way

Orthographic cameras render all objects in the scene at a consistent size regardless of distance, while the size of objects in perspective cameras decreases with size. Keeping sizes of objects looking reasonably consistent when switching views requires picking a distance at which both cameras will render the same object at the same size. After some testing, I found that picking the default distance from the camera to the origin kept things looking reasonable.

Additionally, `OrbitControls` handles zoom differently between perspective and orthographic cameras. To zoom a perspective camera, it physically moves the camera closer to or farther from the target. To zoom an orthographic camera, it doesn't touch the distance from the camera to the target and instead modifies the camera's `zoom` property. A lot of the technical challenge here was a result of having to reconcile these two zoom methods sharing one controller.

Given the notes above, I added a few private methods to `VisGeometry`:
- `getDefaultOrbitRadius`: Derives the default/initial distance between the camera and the target from `cameraDefault`. 98% of the time this ought to be equal to `DEFAULT_CAMERA_Z_POSITION` (120), but trajectories with non-standard units or camera positions will expect the app to notice and keep things consistent.
- `updateOrthographicFrustum`: The default "size" of the orthographic camera varies with the aspect of the frame, the FOV of the perspective camera (in other words, how fast objects get smaller as they get farther from the camera), and the default orbit radius. This method updates that if anything changes.
- `updateControlsZoomBounds`: `OrbitControls` bounds how much the camera can zoom in/out by distance for perspective cameras but by zoom level for orthographic cameras. These could all be set as constants, but the relationship between them will vary if the default orbit radius changes. So this method is here to keep things consistent should that change.

I also had to make some pretty significant changes to the `dolly` method to handle both ways of zooming a camera.

### Type of change

Steps to Verify:
----------------
1. `npm start`
2. Click new "Camera mode" button to switch between perspective and orthographic views
3. Move/zoom the view while switching to verify that position and zoom are preserved reasonably between modes
4. Verify zooming in/out with +/- buttons works in orthographic mode
5. Open tweakpane with Ctrl+Alt+1, verify that setting/saving/loading camera state from there works too

Screenshots:
-----------------------
image
image


src/simularium/types.ts
```diff
@@ -30,6 +30,9 @@ export interface CameraSpec {
     fovDegrees: number;
 }
 
+export type CameraSpecWithType = CameraSpec &
```
toloudis, line 33: can we just modify CameraSpec?

src/visGeometry/rendering/CompositePass.ts
```diff
@@ -251,7 +252,7 @@ class CompositePass {
 
     public render(
         renderer: WebGLRenderer,
-        camera: PerspectiveCamera,
+        camera: PerspectiveCamera | OrthographicCamera,
```
toloudis, line 255: I think threejs might have a base Camera type?

src/simularium/types.ts
frasercl: The regular `CameraSpec` type is an element of `TrajectoryFileInfoV2`, which is read in from JSON. (The `VisGeometry.cameraDefault` property is read in from that field, and has that type.) Existing metadata ought to still get handled fine - anything that checked `orthographic` would get `undefined` and continue treating the spec as perspective - but I didn't want to touch anything at the file interface, and I'm not sure anyways whether it makes sense to provide the option to set the default camera as orthographic in the metadata.

toloudis: I would rather have the CameraSpec at the implementation level be the "complete" one, and change the one in TrajectoryFileInfoV2 to be PerspectiveCameraSpec.  That feels a little cleaner to me, especially since the TrajectoryFileInfo will probably be improved later on.  What do you think of that?  I feel like it would keep visGeometry a little cleaner and there could be a one-time-only conversion from PerspectiveCameraSpec to CameraSpec when `cameraDefault` is being set up.

frasercl: Makes sense to me

src/visGeometry/rendering/CompositePass.ts
toloudis: actually it's really cool that none of the rendering code had to be modified

src/viewport/index.tsx
```diff
@@ -118,6 +118,8 @@ class Viewport extends React.Component<
 
         this.visGeometry = new VisGeometry(loggerLevel);
         this.props.simulariumController.visData.clearCache();
+        // TODO this is called in the `VisGeometry` constructor. Why is it called
+        //   again here? Removing it generates arcane WebGL warnings - why?
```
meganrm, line 122: I tried removing it in the visGeometry constructor and it doesn't give warnings, but I haven't dug into why

src/visGeometry/index.ts
```diff
@@ -157,9 +171,9 @@ class VisGeometry {
     private agentPdbsToDraw: PDBModel[];
 
     public constructor(loggerLevel: ILogLevel) {
-        this.cam1 = cloneDeep(DEFAULT_CAMERA_SPEC);
-        this.cam2 = cloneDeep(DEFAULT_CAMERA_SPEC);
-        this.cam3 = cloneDeep(DEFAULT_CAMERA_SPEC);
+        this.cam1 = { ...cloneDeep(DEFAULT_CAMERA_SPEC), orthographic: false };
+        this.cam2 = { ...cloneDeep(DEFAULT_CAMERA_SPEC), orthographic: false };
+        this.cam3 = { ...cloneDeep(DEFAULT_CAMERA_SPEC), orthographic: false };
```
meganrm: what are these three camera setups for? I only see them used in the setupGui code

frasercl: They're for saveable/reloadable camera states in the test gui:
image

toloudis (approved):

meganrm (approved):

PR #317, opened 2023-06-01
fix error

meganrm:
Problem
=======
In testing the current viewer on main, I get an undefined error on the heartbeatPing

Solution
========
I just added a check, but I don't understand why this would be undefined, so if someone has a better idea, let me know

## Type of change


toloudis:
I think this is actually the right fix:  https://github.com/simularium/simularium-viewer/pull/315/commits/368425fa91fa1cb6d5397093e323e83ccd4b441b

PR #315, opened 2023-05-25
Fix/ao with zoom

toloudis:
Problem
=======
Resolves #303 
Ambient Occlusion had poor behavior when zooming in close.  
This may address #306 too.

Because AO is a screen space algorithm here, we are not going to get a perfect result in all cases.   The things that affect the result are generally the range of very small geometries vs very large ones, and overall scene size (size of distance values).  A scene with a very large bounding box and lots of very small geometries can look different than a scene with a small bounding box and just a few objects filling it.

Solution
========
I chose to replace the AO algorithm completely.  The new one uses fewer samples, has more customizable parameters, and has a smoother result.  The quality is good enough to use with one single render pass as opposed to two, which we were using before.

The AO result is not identical to the previous.  Some care should be taken to verify.  I tried to use settings and variables to scale with scene size and distance such that it is reasonably close to our previous production trajectories.

## Type of change

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)

Screenshots (optional):
-----------------------
The screen shots were not taken at identical viewer sizes so in some cases the number of pixels may also affect the differences.

Before:
Simularium 🔊 2023-05-25 15-09-32
After
Simularium viewer 🔊 2023-05-25 15-09-56

Before:
Simularium 🔊 2023-05-25 15-06-12
After:
Simularium viewer 🔊 2023-05-25 15-08-24

Before:
Simularium 🔊 2023-05-25 15-05-29
After:
Simularium viewer 🔊 2023-05-25 15-08-55

PR #318, opened 2023-06-01
make all json message handlers into arrow functions

meganrm:
Problem
=======
There was a problem with the Remote simulator websocketClient being undefined

Solution
========
pulled this fix from @toloudis PR #315 

## Type of change


toloudis (approved):

frasercl (approved):
Another victim of the strange behavior of `this` in JS

PR #320, opened 2023-06-14
Fix bug with responding to heartbeat ping

ascibisz:
Problem
=======
Joe noticed that errors were showing up in the example viewer when a "ping" message was received from the simularium engine, and the viewer was failing to send a "pong" response. I didn't notice this before because it doesn't come up with Octopus.

I saw this a lot when I was digging in on on [this ticket](https://github.com/simularium/simularium-website/issues/421), and it was bothering me a lot so I dug in while I was there.

It's a tiny fix. Ugh. Still figuring out how to not shoot myself in the foot when I write JavaScript / TypeScript code, I suppose the only to improve is to keep practicing.


toloudis:
Is this the same as https://github.com/simularium/simularium-viewer/pull/318/ ?

Issue #148, opened 2021-07-14
Set agent color


blairlyons:
## Use Case
As a developer, I want to send new colors for agent types and have them update interactively in the viewer.

## Acceptance Criteria
A developer of an application embedding the viewer (esp the simularium website) can call a function to immediately change the color of agents of one or more types.

## Details
This will be required for a color picker on the simularium website to update the colors of agents (https://github.com/allen-cell-animated/simularium-website/issues/197).

Once we enable export of data, users may want to save the colors they've set through the UI, so it might be nice if the current values of these are accessible to developers for that. Although the embedding application could also handle that.

interim17:
Hey @mattakamatsu! Great to hear you're using the color feature! I'm looking into this, both for storing in file and in browser sessions.

Off hand would you say that you're likely to be accessing the trajectory in the same tab you were using before? Same window? Or neither? I'm trying to figure out what level of storage in the browser will be most appropriate for saving the color changes.

mattakamatsu:
First use case would be the same tab, i.e. change the colors, and then
return to that tab in an hour or two; or if the simularium file included
that color info, then I could just re-upload the file with my saved color
changes?

---
Matthew Akamatsu, PhD
Assistant Professor, University of Washington
Department of Biology
He/his
https://www.biology.washington.edu/people/profile/matthew-akamatsu

PR #325, opened 2023-08-17
Feature/set agent color

interim17:
Problem
=======
Closes Set Agent Color [#148](https://github.com/simularium/simularium-viewer/issues/148)

Solution
========
In brief: colorChanges are now part of SelectionStateInfo. Viewer watches for changes to selectionStateInfo.colorChanges and calls changeAgentsColor in response to changes. changeAgentsColor updates visGeometry and UiDisplayData to ensure color changes are represented both in the viewport and in embedding applications that reference the UiDisplayData.

Added a few helper functions, cleaned unused ColorMapping in visGeo and made small updates to existing functions to accommodate new behavior.

Selection Interface:
- Added property to the SelectionEntry type so that it now includes colorChanges
- previous function setAgentColors divided into setInitialAgentColors and updateAgentColors

Viewport:
- ComponentDidUpdate checks for new colorChangeAgents and calls changeAgentsColor which is the primary method for this PR

VisGeometry:
- add applyColorToAgents and indexOfColor, remove unused ColorMapping functions

Example Viewer/ColorPicker.tsx
- Added a new component file to prevent further growth of massive Viewer.tsx file, maybe a refactor of Viewer.tsx into different components would be wise at some point

Steps to Verify:
----------------
1. Try it out in the example viewer: select agents (and optionally subagents), select a color, apply color to agent. New colors can be provided via input field.

### DIFF
--- a/examples/ColorPicker.tsx
+++ b/examples/ColorPicker.tsx
@@ -49,10 +49,10 @@ const ColorPicker = ({
                     tags: subAgent,
                 },
             ];
-            setColorSelectionInfo({
+            setColorSelectionInfo([{
                 agents: entry,
                 color: selectedColor,
-            });
+            }]);
         }
     };
 
--- a/src/simularium/SelectionInterface.ts
+++ b/src/simularium/SelectionInterface.ts
@@ -25,7 +25,7 @@ export interface ColorChanges {
 export interface SelectionStateInfo {
     highlightedAgents: SelectionEntry[];
     hiddenAgents: SelectionEntry[];
-    colorChanges: ColorChanges;
+    colorChanges: ColorChanges[];
 }
 
 interface DisplayStateEntry {
@@ -277,6 +277,7 @@ class SelectionInterface {
     ): void {
         const newColor = convertColorNumberToString(color);
         const entry = this.entries[agentName];
+        // if no display state update parent color
         entry.forEach((displayState) => {
             if (idsToUpdate.includes(displayState.id)) {
                 displayState.color = newColor;
--- a/src/simularium/SelectionInterface.ts
+++ b/src/simularium/SelectionInterface.ts
@@ -18,6 +18,7 @@ export interface SelectionEntry {
 export interface SelectionStateInfo {
     highlightedAgents: SelectionEntry[];
     hiddenAgents: SelectionEntry[];
+    colorChangeAgents: SelectionEntry[];
 }
 
 interface DisplayStateEntry {
@@ -215,6 +216,18 @@ class SelectionInterface {
             indices = [...indices, ...this.getIds(name, tags)];
         });
 
+        return indices;
+    }
+
+    public getColorChangeAgentIds(info: SelectionStateInfo): number[] {
+        const requests = info.colorChangeAgents;
+        let indices: number[] = [];
+
+        requests.forEach((r) => {
+            const name = r.name;
+            const tags = r.tags;
+            indices = [...indices, ...this.getIds(name, tags)];
+        });
         return indices;
     }
 
--- a/src/simularium/SelectionInterface.ts
+++ b/src/simularium/SelectionInterface.ts
@@ -1,6 +1,7 @@
-import { filter, find, uniq } from "lodash";
+import { filter, find, set, uniq } from "lodash";
 import { EncodedTypeMapping } from "./types";
 import { convertColorNumberToString } from "../visGeometry/color-utils";
+import _ from "lodash";
 
 // An individual entry parsed from an encoded name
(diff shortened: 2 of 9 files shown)


src/visGeometry/index.ts
```diff
@@ -1146,11 +1153,17 @@ class VisGeometry {
...
+        //TODO: this is commented out because I'm unsure how best to handle this
+        // the purpose of this feature is to allow color mapping to be changed
+        // should we just remove this error handling? or should the color change methods
+        // clear the mapping, apply the color change then reset it?
+
+        // if (this.isIdColorMappingSet) {
+        //     throw new FrontEndError(
+        //         "Attempted to set agent-color after color mapping was finalized"
+        //     );
+        // }
```
interim17, line 1166: @meganrm thoughts on this? not sure the ideal way to approach this. If this is error handling feels necessary I'll add code needed to clear and reset color mapping, but I'm trying to keep things streamlined.

src/visGeometry/index.ts
```diff
@@ -1086,18 +1084,44 @@ class VisGeometry {
...
+
+    public addNewColor(color: number | string): number {
         const colorNumber = convertColorStringToNumber(color);
         const newColor = [
             ((colorNumber & 0x00ff0000) >> 16) / 255.0,
             ((colorNumber & 0x0000ff00) >> 8) / 255.0,
             ((colorNumber & 0x000000ff) >> 0) / 255.0,
             1.0,
         ];
+        const currentIndex = this.indexOfColor(newColor);
```
toloudis, line 1113: when does addNewColor get called?  Do you always know the agent type when you are calling it?

src/simularium/SelectionInterface.ts
```diff
@@ -1,8 +1,11 @@
-import { filter, find, uniq } from "lodash";
+import { filter, find, set, uniq } from "lodash";
 import { EncodedTypeMapping } from "./types";
 import { convertColorNumberToString } from "../visGeometry/color-utils";
+import _ from "lodash";
```
toloudis, line 4: can remove

src/simularium/SelectionInterface.ts
```diff
@@ -15,9 +18,15 @@ export interface SelectionEntry {
     tags: string[];
 }
 
+export interface ColorChanges {
+    agents: SelectionEntry[];
```
toloudis, line 22: this would be better named as agentTypes.

toloudis: on second thought, I realize we use "agents" in other SelectionEntry names as well. so at least it's consistent.   Slightly confusing because we don't set colors on individual agents.

src/simularium/SelectionInterface.ts
interim17: oops, fixed this and the two other things the linter caught

src/visGeometry/index.ts
meganrm: it's just in the case that the user changed the color of an agent to a color already in the data array

interim17: These get called in changeAgentsColor in viewport, which takes agent, tags, and color in the form of colorChanges.  The function call that needs the IDs returned from addNewColor is setColorForIds in visGeo.

toloudis (approved):
LGTM!

meganrm (approved):
looks good, there's just one comment I left to myself that is no longer relevant

meganrm:
todo: make sure `ColorChanges` and `SelectionEntry` are exported from `type-declarations` before you merge this

examples/ColorPicker.tsx
```diff
@@ -0,0 +1,123 @@
+import React, { useState } from "react";
+import { SelectionEntry } from "../type-declarations/simularium/SelectionInterface";
+
+const ColorPicker = ({
+    uiDisplayData,
+    particleTypeNames,
+    agentColors,
+    setColorSelectionInfo,
+    updateAgentColorArray,
```
ShrimpCryptid, line 9: Just peeking in, should these props have type annotations?

toloudis: I am guessing that the type checking rules are not set up to enforce this in our typecheck/linter. It would be good to turn those rules on in a separate PR!

interim17: The example Viewer code can be a bit fast and loose compared the to the core Viewer code. I like the idea of typing it better, or even refactoring it into component files someday when there is bandwidth.

toloudis: I didn't realize this was the viewer test-app code.  Totally agree that the bar is lower for those source files, although it doesn't need to be.

Issue #326, opened 2023-08-19
right-click drag panning behavior changed


toloudis:
## Description

(In current main branch, not on live production server)
When I right-click drag to pan the view, instead there is some combined zoom and rotation behavior.

## Expected Behavior

Right-click drag should pan the view (translate both view position and view target in the plane of the screen)

## Reproduction

Must use a recent git branch to see this behavior.  It is very likely due to updates in three.js's OrbitControls that we need to work better with.

## Environment

Observed in both Chrome and Firefox.

toloudis:
fixed by #327

PR #327, opened 2023-08-23
delay webgl init until after component mounted

toloudis:
Problem
=======
#326 the zoom/pan/rotate controls were not working as expected, due to changes from #319 .

Solution
========
Delay initialization of threejs WebGL until after component mount.  Then create OrbitControls and attach to the WebGL canvas element.   This has restored the correct functioning.  We were previously creating and re-creating webgl renderers and the canvases that go along with them.  

## Type of change

Change summary:
---------------
* Need to re-test within simularium-website.  Also I noticed that resetCamera didn't seem to nicely bound every scene but I wasn't sure if it was a problem with the data or the code.  Will do more testing within context of the website app.

### DIFF
--- a/src/visGeometry/index.ts
+++ b/src/visGeometry/index.ts
@@ -250,7 +250,6 @@ class VisGeometry {
         this.camera.position.z = DEFAULT_CAMERA_Z_POSITION;
         this.initCameraPosition = this.camera.position.clone();
 
-        //this.controls = this.setupControls();
         this.focusMode = true;
 
         this.tickIntervalLength = 0;
--- a/src/visGeometry/index.ts
+++ b/src/visGeometry/index.ts
@@ -120,13 +120,13 @@ class VisGeometry {
     public agentPaths: Map<number, AgentPath>;
     public mlogger: ILogger;
     // this is the threejs object that issues all the webgl calls
-    public threejsrenderer: WebGLRenderer;
+    public threejsrenderer!: WebGLRenderer;
     public scene: Scene;
 
     public perspectiveCamera: PerspectiveCamera;
     public orthographicCamera: OrthographicCamera;
     public camera: PerspectiveCamera | OrthographicCamera;
-    public controls: OrbitControls;
+    public controls!: OrbitControls;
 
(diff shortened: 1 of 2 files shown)


interim17:
I just put a build of this branch in the website locally and the camera behavior seems correct. I'm can see the difference between example viewer running main branch, and website running the fixed build.

src/visGeometry/index.ts
```diff
@@ -120,13 +120,13 @@ class VisGeometry {
     public agentPaths: Map<number, AgentPath>;
     public mlogger: ILogger;
     // this is the threejs object that issues all the webgl calls
-    public threejsrenderer: WebGLRenderer;
+    public threejsrenderer!: WebGLRenderer;
```
interim17, line 123: Just because I don't know: is using this operator to delay assignment considered best practice if you're confident that its going to be assigned, or is it hacky as it circumvents a kind of check that typescript does? It's fine with me, just wondering.

interim17:
LGTM, works in example viewer and on website.

interim17 (approved):

src/visGeometry/index.ts
toloudis: it's kinda both.  In this case I could have done a typescript definition like `WebGLRenderer | undefined` but the point here is that any usage of that variable before component mount time (before reparent is called) is really an error and totally unexpected, and should have an explicit check.

toloudis:
> LGTM, works in example viewer and on website.

I'm curious, when you integrated it with website, did you test changing trajectory and doing things like camera reset?

interim17:
> 

Yes, but I also wasn't super familiar with the buggy behavior... Happy to huddle and share screen and show ya if you want!

meganrm (approved):

src/visGeometry/index.ts
toloudis: another possible way to handle it is to do `WebGLRenderer | null` and explicitly init it to null in the constructor.  That way it is at least clear that we intentionally did this.  Arguably that is a more robust / readable way to do it.

PR #328, opened 2023-08-23
Feature/set agent color edits

meganrm:
Problem
=======
1. There were some existing functions in visGeometry we could make use of, 
2. I also noticed that we were actually storing the new colors permanently in the selection interface
3. There were too many things being added to the example viewer state that made things confusing 

Solution
========
See comments in the code

### DIFF
--- a/examples/ColorPicker.tsx
+++ b/examples/ColorPicker.tsx
@@ -12,6 +12,7 @@ const ColorPicker = ({
     const [selectedAgent, setSelectedAgent] = useState("");
     const [selectedColor, setSelectedColor] = useState("");
     const [selectedSubagent, setSelectedSubAgent] = useState("");
+    const [colorToAppend, setColorToAppend] = useState("");
 
     const getSubAgentsforAgent = (agentName: string) => {
         const agent = uiDisplayData.find(
@@ -55,9 +56,7 @@ const ColorPicker = ({
         }
     };
 
-    const addColorToColorArray = (event) => {
-        const customColor = event.target.value;
-
+    const addColorToColorArray = (customColor: string) => {
         const hexColorCodeRegex = /^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/;
         const color = customColor;
         if (hexColorCodeRegex.test(color)) {
@@ -110,9 +109,11 @@ const ColorPicker = ({
                 id="colorAddition"
                 type="text"
                 placeholder="add Hex Color"
-                onChange={addColorToColorArray}
+                onChange={(event) => {
+                    console.log(event.target.value)
+                    setColorToAppend(event.target.value)}}
             ></input>
-            <button onClick={addColorToColorArray}>
+            <button onClick={() => addColorToColorArray(colorToAppend)}>
                 Add color to color array
             </button>
         </>
--- a/examples/ColorPicker.tsx
+++ b/examples/ColorPicker.tsx
@@ -1,0 +1,122 @@
+import React, { useState } from "react";
+import { SelectionEntry } from "../type-declarations/simularium/SelectionInterface";
+
+const ColorPicker = ({
+    uiDisplayData,
+    particleTypeNames,
+    agentColors,
+    setColorSelectionInfo,
+    updateAgentColorArray,
+}) => {
+    const [subAgents, setSubAgents] = useState([{ name: "", id: "" }]);
+    const [selectedAgent, setSelectedAgent] = useState("");
+    const [selectedColor, setSelectedColor] = useState("");
+    const [selectedSubagent, setSelectedSubAgent] = useState("");
+
+    const getSubAgentsforAgent = (agentName: string) => {
+        const agent = uiDisplayData.find(
+            (element) => element.name === agentName
+        );
+        if (!agent) {
+            throw new Error("No agent found");
+        }
+        setSubAgents(agent.displayStates);
+    };
+
+    const handleAgentSelection = (event) => {
+        const value = event.target.value;
+        getSubAgentsforAgent(value);
+        setSelectedAgent(value);
+    };
+
+    const assignColorToAgent = () => {
+        if (!selectedAgent) {
+            throw new Error("No agent selected");
+        } else if (!selectedColor) {
+            throw new Error("No color selected");
+        } else {
+            let subAgent: string[] = selectedSubagent ? [selectedSubagent] : [];
+            // hooks doesn't save an empty string
+            // but an empty string is a possible tag
+            // that represents the unmodified state
+            if (selectedSubagent === "<unmodified>") {
+                subAgent = [""];
+            }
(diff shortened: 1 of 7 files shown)


examples/ColorPicker.tsx
meganrm, line 1: we did this together, the main reason to do this was to try not to clutter up the viewer state with a lot of more temporary selections

examples/ColorPicker.tsx
```diff
@@ -0,0 +1,123 @@
...
+        } else {
+            alert("Please enter a valid hex color code");
+        }
+    };
+
+    return (
+        <>
+            <span>Color change agent selections:</span>
+            <select id="agentSelect" onChange={handleAgentSelection}>
+                <option value=""> Select Agent</option>
```
meganrm, line 73: if you are going to include a placeholder option, make sure the value is set to empty string, otherwise you get "Select Agent" as your selection

examples/ColorPicker.tsx
```diff
@@ -0,0 +1,123 @@
...
+            </select>
+            <select
+                id="subAgentSelect"
+                onChange={(event) => setSelectedSubAgent(event.target.value)}
+            >
+                <option value="">Select Sub-Agent</option>
+                {subAgents.map((subAgent) => (
+                    <option
+                        key={subAgent.name}
+                        value={subAgent.id || subAgent.name}
```
meganrm, line 88: we have subAgents with and ID of an empty string. This is the unmodified agent. it's a little convoluted but because hooks doesn't store the empty string value I had to do this here

examples/ColorPicker.tsx
```diff
@@ -0,0 +1,123 @@
...
+                ))}
+            </select>
+            <button onClick={assignColorToAgent}> Apply Color to Agent</button>
+            <input
+                id="colorAddition"
+                type="text"
+                placeholder="add Hex Color"
+                onChange={(event) => {
+                    console.log(event.target.value)
+                    setColorToAppend(event.target.value)}}
```
meganrm, line 114: before you were also calling the addColorToColorArray here, which meant you couldn't type in the input field, because it would keep checking to see if you had a hex value yet. I imagine you were only pasting into this field so you didn't run into this.

examples/Viewer.tsx
```diff
@@ -146,15 +140,10 @@ const initialState: ViewerState = {
     selectionStateInfo: {
         highlightedAgents: [],
         hiddenAgents: [],
-        colorChangeAgents: [],
+        colorChanges: { agents: [], color: "" },
```
meganrm, line 143: I thought it was better to keep these data fields together since it's all connected rater than have the new color separated out

src/simularium/SelectionInterface.ts
```diff
@@ -186,14 +197,9 @@ class SelectionInterface {
...
     }
-    /*
-     * If an entity has both a name and all the tags specified in the
-     * selection state info, it will be considered highlighted
-     */
-    public getHighlightedIds(info: SelectionStateInfo): number[] {
-        const requests = info.highlightedAgents;
-        let indices: number[] = [];
 
+    public getAgentIdsByNamesAndTags(requests: SelectionEntry[]): number[] {
```
meganrm, line 201: we had this same code 3 times so I separated it out

examples/ColorPicker.tsx
interim17: you are correct

src/simularium/SelectionInterface.ts
```diff
@@ -276,19 +271,39 @@ class SelectionInterface {
...
-                displayState.color = convertColorNumberToString(color);
+    ): void {
+        const newColor = convertColorNumberToString(color);
+        const entry = this.entries[agentName];
+        entry.forEach((displayState) => {
+            if (idsToUpdate.includes(displayState.id)) {
+                displayState.color = newColor;
+            }
+        });
+    }
```
meganrm, line 285: This might necessitate a change on the front end. As it was, it was modified the uiDisplay data, which is derived from the selectionInterface. But it's not being stored. Now when you adjust the color it's being stored in the selection interface data

src/test/SelectionInterface.test.ts
```diff
@@ -331,70 +335,14 @@ describe("SelectionInterface module", () => {
...
-            si.parse(idMapping);
-            const selectionStateColor = {
-                highlightedAgents: [],
-                hiddenAgents: [],
-                colorChangeAgents: [
-                    { name: "A", tags: [] },
-                    { name: "C", tags: [] },
-                ],
-            };
-            const ids = si.getColorChangeAgentIds(
```
meganrm, line 354: this was really just testing the same code we already had tested now that I consolidated it into one function

src/viewport/index.tsx
```diff
@@ -547,66 +543,36 @@ class Viewport extends React.Component<
...
-            color,
-            this.props.agentColors as string[]
-        );
-        this.applyColorToAgents(agentIds, colorId);
+    private getColorId(color: string): number {
+        /**
+         * Check if the new color is in our current array of color options, if not,
+         * add it before returning the index
+         */
+        return this.visGeometry.addNewColor(color);
```
meganrm, line 551: there was already a function in visGeometry for adding a color to the color data

src/visGeometry/index.ts
```diff
@@ -726,11 +724,6 @@ class VisGeometry {
         this.updateScene(this.currentSceneAgents);
     }
 
-    public setColorChangeByIds(ids: number[]): void {
-        this.colorChangeIds = ids;
-        this.updateScene(this.currentSceneAgents);
```
meganrm, line 731: this updateScene call is needed to trigger the canvas to re-render but we didn't need to store the ids of the agents that were just changed color, we have that information elsewhere

src/viewport/index.tsx
```diff
@@ -547,66 +543,36 @@ class Viewport extends React.Component<
...
-            this.setColors(updatedColors);
-        }
-        const colorId = typeCastAgentColors.indexOf(color);
-        return colorId;
     }
 
     public setColors(agentColors: string[] | number[]): void {
-         this.visGeometry.clearColorMapping();
-         const uiDisplayData = this.selectionInterface.getUIDisplayData();
-         const updatedColors = this.selectionInterface.setAgentColors(
```
meganrm, line 596: this really updates all the colors at once, which we don't need to do when we're just changing one agents color. Plus this is relying on a prop staying constant (the agent colors), which is not something this component has control over, because the parent component could just change that whole list. In which case we would want to trigger re-mapping of all the colors

interim17 (approved):

examples/ColorPicker.tsx
interim17: the one behavior i'm seeing now that we might not want, is that when a color is added to the array it is immediately applied to some agents, even when no agents have been selected.

PR #334, opened 2023-10-10
fix type exports

interim17:
I merged changes for the Set Agent Colors feature that caused errors. 

ColorChanges and SelectionEntry types were not being exported properly,  I think this should fix it.

To check:
It was giving errors when running `npm run build` prior to fix, now I can build without seeing those errors, and the type check here in GitHub seems happy. Any other ways we should confirm?

Apologies for the slowdown this caused!

### DIFF
--- a/src/simularium/index.ts
+++ b/src/simularium/index.ts
@@ -5,6 +5,8 @@ export type {
     TrajectoryFileInfo,
     EncodedTypeMapping,
     SimulariumFileFormat,
+    SelectionEntry,
+    ColorChanges,
 } from "./types";
 export type { SelectionStateInfo, UIDisplayData } from "./SelectionInterface";
 export { ErrorLevel, FrontEndError } from "./FrontEndError";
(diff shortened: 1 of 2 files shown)


ShrimpCryptid (approved):
Looks like this builds fine for me with `npm run build`! LGTM 👍

src/simularium/index.ts
```diff
@@ -5,6 +5,8 @@ export type {
     TrajectoryFileInfo,
     EncodedTypeMapping,
     SimulariumFileFormat,
+    SelectionEntry,
+    ColorChanges,
```
toloudis, line 9: I think the types are already defined in SelectionInterface (the import below this line).  Could you just get it from there and skip adding it to types.ts?

interim17: Yes I think so!

The comments from Megan that I was trying to address when I created the error, had to do with getting these types exported from the right places. I'm honestly not sure what best practice would be... declaring them in SelectionInterface or types.ts?

Seems like we can resolve the errors with either approach.

toloudis: so right now with this PR they are declared in two places. It would be better if they are defined only once.

interim17: Ok, best to keep it in SelectionInterface I think.

toloudis (approved):
ship it!

PR #335, opened 2023-10-11
change dropbox links

meganrm:
Problem
=======
Dropbox links to OBJ files started having a CORS error

Solution
========
I replaced the `www.dropbox.com` with `dl.dropboxusercontent.com`
[based on this post](https://www.dropboxforum.com/t5/View-download-and-export/Shared-Link-quot-scl-quot-to-quot-s-quot/m-p/695266/highlight/true#M49151)
They do note in the post that `dl.dropboxusercontent.com` is an old style that they still support "for now" so we'll have to keep an eye on it. I did ask in the forum if they plan to address the CORS issue with the new link style, so we'll see what they say. 

## Type of change

Steps to Verify:
----------------
1. `npm start`
1. open test bed viewer 
2. drag and drop [this file](https://allencellscience.slack.com/archives/CMQFRJPPF/p1696971385939479?thread_ts=1696965553.287399&cid=CMQFRJPPF)
3. should see structures

Screenshots (optional):
-----------------------
Screenshot 2023-10-11 at 10 16 41 AM

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -37,7 +37,11 @@ export const checkAndSanitizePath = (pathOrUrl: string): string => {
     const isUrlRegEX =
         /(https?:\/\/)([\w\-])+\.{1}([a-zA-Z]{2,63})([\/\w-]*)*\/?\??([^#\n\r]*)?#?([^\n\r]*)/g;
     if (isUrlRegEX.test(pathOrUrl)) {
-        return pathOrUrl;
+        let url = pathOrUrl;
+        if (url.includes("dropbox")) {
+            url = url.replace("www.dropbox.com", "dl.dropboxusercontent.com");
+        }
+        return url;
     } else if (/\B\//g.test(pathOrUrl)) {
         return pathOrUrl;
     }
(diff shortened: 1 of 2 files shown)


toloudis (approved):

interim17 (approved):

src/util.ts
```diff
@@ -37,7 +37,11 @@ export const checkAndSanitizePath = (pathOrUrl: string): string => {
     const isUrlRegEX =
         /(https?:\/\/)([\w\-])+\.{1}([a-zA-Z]{2,63})([\/\w-]*)*\/?\??([^#\n\r]*)?#?([^\n\r]*)/g;
     if (isUrlRegEX.test(pathOrUrl)) {
-        return pathOrUrl;
+        let url = pathOrUrl;
+        if (url.includes("dropbox")) {
+            url = url.replace("www.dropbox.com", "dl.dropboxusercontent.com");
```
ShrimpCryptid, line 42: Would it make sense to just replace `dropbox.com` and leave out the `www.`? That seems like an optional part of the link and could be missing in some circumstances

PR #341, opened 2023-11-07
fix drag n drop in test viewer

toloudis:
Problem
=======
Drag and drop with multiple geometry assets was not working in the testbed viewer.

Solution
========
The test viewer uses a slightly different code path than the simularium-website viewer for loading from drag and drop.  It was checking a data structure in the wrong way, assuming it was an array instead of a keyed object.  This resulted in it ignoring the provided geometry files.
(As far as I can tell, this code path is unused by the website client so it should be totally safe.  It was needed to debug a different issue with meshes)

Steps to Verify:
----------------
1. Drag and drop a .simularium file accompanied by its necessary .obj meshes.
2. In the bug case, the meshes totally fail to load.


interim17:
LGTM, tested with the examples @mogres provided in the slack thread.

interim17 (approved):

meganrm (approved):

PR #343, opened 2023-11-16
Fix/viewer errors

meganrm:
Problem
=======

The test bed viewer had a lot of errors, and it loaded files different than the web app, ie you had to click "load". 

Solution
========
I loaded everything from the local files, and changed the "on change" function to just load the file immediately. 

## Type of change

### DIFF
--- a/examples/Viewer.tsx
+++ b/examples/Viewer.tsx
@@ -5,17 +5,18 @@ import type {
     ISimulariumFile,
     UIDisplayData,
     SelectionStateInfo,
+    SelectionEntry,
 } from "../type-declarations";
 import SimulariumViewer, {
     SimulariumController,
     RenderStyle,
-    ErrorLevel,
-    FrontEndError,
     loadSimulariumFile,
-} from "../es";
+    FrontEndError,
+    ErrorLevel,
+} from "../src/index";
 import "../style/style.css";
-
 import PointSimulator from "./PointSimulator";
+import BindingSimulator from "./BindingSimulatorLive";
 import PointSimulatorLive from "./PointSimulatorLive";
 import PdbSimulator from "./PdbSimulator";
 import SinglePdbSimulator from "./SinglePdbSimulator";
@@ -118,7 +119,9 @@ interface InputParams {
     useOctopus: boolean;
 }
 
-const simulariumController = new SimulariumController({});
+const simulariumController = new SimulariumController(
+    {}
+);
 
 let currentFrame = 0;
 let currentTime = 0;
@@ -146,8 +149,6 @@ const initialState: ViewerState = {
     filePending: null,
     simulariumFile: null,
 };
-
-type FrontEndError = typeof FrontEndError;
 
 class Viewer extends React.Component<InputParams, ViewerState> {
     private viewerRef: React.RefObject<SimulariumViewer>;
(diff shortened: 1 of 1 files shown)


toloudis:
some of this is already in #341

examples/Viewer.tsx
```diff
@@ -5,16 +5,16 @@ import type {
...
 import SimulariumViewer, {
     SimulariumController,
     RenderStyle,
-    ErrorLevel,
-    FrontEndError,
     loadSimulariumFile,
-} from "../es";
+    FrontEndError,
+    ErrorLevel,
+} from "../src/index";
```
toloudis, line 16: The original `from "../es"` was there intentionally to help simulate an environment more like simularium-website where we are loading from the packaged build.  This was relevant because of differences in the way the web worker and style sheets are loaded, for example.  So if changing this, we should be intentional about it

meganrm: ok, it seemed like this was what was causing a lot of the typing errors and it means you don't have to rebuild to make sure all your changes are reflected in the app. I don't see any differences after changes these to the way the app behaves. do you know of something I could test?

meganrm: I had to change the server to production because the staging one is down, but I was able to load COVIDLUNG and it seemed to behave the same in both cases (changing the imports)

toloudis: This is expected to behave the same if you : 
* use `from ../es` and run `npm build` before `npm start`
* compared to use `from ../src` and don't build before start.
Main thing to test is loading a trajectory with pdbs. 
But the ../es case is more important to make sure things are working for production. 
Being intentional probably means just keeping it easy to switch between the two (keep all the imports in this one statement if possible) and maybe put something in the readme, and a comment at the import to say what to do.

meganrm: no I know, I meant is there any behavior you know about that does behave differently in the two states? Because either way we're running the same commands, ie, `npm start` calls `npm build` and the dev server setup is supposed to serve the built artifacts just while watching for changes. I'm in favor of using the `/src` because it removes the typescript errors so it's more obvious when we do have an actual error in the test bed viewer. But I can add a commented out import statement? Is that what you're suggesting?

toloudis: There should be no actual differences in runtime behavior between the two cases of importing from src vs es. In the past, however, there HAVE been differences which had to be fixed.  Loading it here lets you find the error before you test in simularium-website.  This is the reason why this demo app is loading from `es/`.

There definitely shouldn't be any typescript errors when we import from the es/ dir, so if there are, maybe that means we need to make sure that the test app can see the `type-declarations` directory.

toloudis: So I would like to keep the ability to run the test app with the post-build artifacts.

In order to support both situations (import from es and requiring a npm build, versus importing from src and not requiring a npm build) I would suggest a separate `start` script in package.json and some kind of dev-or-production switch in the code.  The alternative is to have a commented-out version of both like this:

```
import foo, bar, baz from 'src';
// import foo, bar, baz from 'es';
```
As long as there is only one place to do this, then you have to just change the comment to make it use the right thing. And then every PR from now on will have this comment battle... 🤣

ShrimpCryptid (approved):
Looks good!

toloudis (approved):

PR #344, opened 2023-11-16
Feature/mp4 download

interim17:
### Problem
Users of Simularium want to download mp4 video. 
It's already marked closed, but this addresses #240

### Approach
`FrameRecorder` class handles the frame by frame recording of the `canvas`. Aimed for a good separation of concerns: using `recorder` to just record; front end handles downloading, etc. 

`Viewport.tsx` makes an instance of `FrameRecorder` and watches for a boolean flag, `recording`.
`recorder.onFrame()` is called in the `animate()` loop to keep frame capture tightly coupled to scene updates.

`Viewer.tsx` stores `recording` in state, and defines `onRecordedMovie` for handling recorded the output.

`RecordMovieComponent` buttons and a print of recording duration, this can get fleshed out more if desired?

`onCompletedRecording` could also send other data if wanted, number of frames, total duration of frames, etc.

There was a bug that I discussed with @toloudis that I still don't fully understand, where the `canvas` reference would go stale and return blank frames after changing Simularium files. Instantiating a new `VideoEncoder` on each call of `setup()` fixed the problem, but I'm still curious why the old `encoder` went stale, as it doesn't receive a canvas reference at all...

### To confirm:
- open a trajectory
- start recording, and alter playback in various ways (starting, stopping, using camera, zooming, changing colors)
- stop recording
- open a new trajectory and repeat steps
- see if the resulting videos download, match your session, and look good


package.json
```diff
@@ -102,8 +103,12 @@
         "@fortawesome/react-fontawesome": "^0.1.14",
         "@tweakpane/core": "^1.0.8",
         "@tweakpane/plugin-essentials": "^0.1.4",
+        "@types/dom-mediacapture-record": "^1.0.19",
+        "@types/dom-webcodecs": "^0.1.8",
+        "2.2.2": "^2.2.2",
```
meganrm, line 108: this looks like a mistake

src/simularium/FrameRecorder.ts
```diff
@@ -0,0 +1,143 @@
...
+        getCanvas: () => HTMLCanvasElement | null,
+        handleFile: (videoBlob: Blob) => void
+    ) {
+        this.getCanvas = getCanvas;
+        this.handleFile = handleFile;
+        this.encoder = null;
+        this.isRecording = false;
+        this.frameIndex = 0;
+        this.supportedBrowser = !/firefox|fxios/i.test(navigator.userAgent);
+    }
```
interim17, line 32: was originally going to handle all browser checking in the website/front end, but @meganrm suggested we check for firefox further down and calling `stop()` on firefox was leading to errors. you could circumvent this on the front end, but this prevent it in the first place.

toloudis: I wonder if there is a better way to init `supportedBrowser` by actually checking for the api that will be used?

examples/src/RecordMovieComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setIsRecording: (isRecording: boolean) => void;
+}
+
+const RecordMovieComponent = ({
+    isRecording,
+    setIsRecording,
+}: RecordMovieComponentProps) => {
+    let [recordingDuration, setRecordingDuration] = useState<number>(0);
+    let [recordingStatus, setRecordingStatus] =
+        useState<string>("Not recording");
```
ShrimpCryptid, line 14: Does this state need to exist? It seems like you could remove this, and just have the text switch between "Not recording" and "Recording..." based on the `isRecording` prop.

examples/src/RecordMovieComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    return (
+        <div>
+            <button onClick={startRecording} disabled={isRecording}>
+                Start Recording
+            </button>
+            <button onClick={stopRecording} disabled={!isRecording}>
+                Stop Recording
+            </button>
+            <div>{recordingStatus}</div>
```
ShrimpCryptid, line 51:
```suggestion
            <div>{isRecording ? "Recording..." : "Not recording"}</div>
```

examples/src/RecordMovieComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const RecordMovieComponent = ({
+    isRecording,
+    setIsRecording,
+}: RecordMovieComponentProps) => {
+    let [recordingDuration, setRecordingDuration] = useState<number>(0);
+    let [recordingStatus, setRecordingStatus] =
+        useState<string>("Not recording");
+    let [outputStatus, setOutputStatus] = useState<string>("");
```
ShrimpCryptid, line 15: Same here, this could also be a bool/enum

src/simularium/FrameRecorder.ts
```diff
@@ -0,0 +1,144 @@
...
+                // making a new one during setup() prevents a bug where
+                // frames returned blank (stale canvas reference?)
+                this.encoder = new VideoEncoder({
+                    output: (chunk, meta) => {
+                        if (this.isRecording && this.muxer) {
+                            this.muxer?.addVideoChunk(chunk, meta);
+                        }
+                    },
+                    error: (error) => {
+                        console.error("Encoder error:", error);
```
toloudis, line 52: should this be a `throw new Error`?  Are you able to cause it to happen?

src/simularium/FrameRecorder.ts
```diff
@@ -0,0 +1,144 @@
...
+                    codec: "avc1.420028",
+                    width: canvas.width,
+                    height: canvas.height,
+                };
+                const { supported, config: supportedConfig } =
+                    await VideoEncoder.isConfigSupported(config);
+                if (supported && supportedConfig) {
+                    this.encoder.configure(config);
+                } else {
+                    throw new Error("Unsupported video encoder configuration");
```
toloudis, line 65: Same question here, have you seen this error happen and when?

src/simularium/FrameRecorder.ts
```diff
@@ -0,0 +1,144 @@
...
+                    target: new ArrayBufferTarget(),
+                    video: {
+                        codec: "avc",
+                        width: canvas.width,
+                        height: canvas.height,
+                    },
+                });
+                this.frameIndex = 0;
+            } catch (error) {
+                throw new Error("Error setting up video encoder: " + error);
```
toloudis, line 78: this one might be related to the supportedBrowser=false case, but could also include other conditions.

src/simularium/FrameRecorder.ts
```diff
@@ -0,0 +1,144 @@
...
+            if (this.encoder.encodeQueueSize > 2) {
+                console.log("Dropping frame, too many frames in flight");
+                return;
+            }
+            const canvas = this.getCanvas();
+            if (canvas) {
+                // Add a keyframe every second: https://en.wikipedia.org/wiki/Key_frame
+                const keyFrame = this.frameIndex % this.frameRate === 0;
+                const timestampMicroseconds =
+                    (this.frameIndex / this.frameRate) * 1_000_000;
```
toloudis, line 112: Nit: I'm not sure I get the math here.  Because the calling code is already controlling the frame rate, could this be something more like Date.now() - startTime ?  I don't know that anything needs to change because maybe what I'm suggesting is really equivalent.  Are the timestamps supposed to be perfectly regular or can they be the true wall clock time?

toloudis (approved):
LGTM , just a couple of minor questions

src/simularium/FrameRecorder.ts
ShrimpCryptid: We do the check in Colorizer like this: 
```
  public static isSupported(): boolean {
    // Check if WebCodecs API exists
    return typeof VideoEncoder === "function";
  }
```

https://github.com/allen-cell-animated/nucmorph-colorizer/blob/main/src/components/Export.tsx#L326

src/simularium/FrameRecorder.ts
```diff
@@ -0,0 +1,144 @@
...
+                    },
+                    error: (error) => {
+                        console.error("Encoder error:", error);
+                    },
+                });
+                const config: VideoEncoderConfig = {
+                    codec: "avc1.420028",
+                    width: canvas.width,
+                    height: canvas.height,
+                };
```
ShrimpCryptid, line 59: May be a good idea to check for hardware acceleration and to define stuff like the framerate of the resulting video here.
Using the avc codec makes sense and should be broadly supported, but might be worth doing a check for official documentation on that.

https://github.com/allen-cell-animated/nucmorph-colorizer/blob/main/src/colorizer/recorders/Mp4VideoRecorder.ts#L43-L91

src/simularium/FrameRecorder.ts
interim17: Oh interesting @ShrimpCryptid Just came back here to reply to Dan's comment because
 `this.supportedBrowser = "VideoEncoder" in window;` 
 seemed to be working. I think they are more or less equivalent?

examples/src/RecordMovieComponent.tsx
interim17: You're right this is a stow away from the previous implementation.

examples/src/RecordMovieComponent.tsx
interim17: Except that I want it to be a `Finished recording` message on `stop`, but an empty string on page load which is also `isRecording = false`

src/simularium/FrameRecorder.ts
interim17: I can make it happen by corrupting the setup code, and it does seem like it should throw an error. I haven't made this happen in the normal course of playing with the code.

src/simularium/FrameRecorder.ts
interim17: In regard to this and above, tbh I was including error handling where the docs/examples seemed to have it. Caniuse says `avc` is supported almost everywhere. Except Opera mini and some very old browser versions

src/simularium/FrameRecorder.ts
interim17: The `encoder` requires that the time stamps start at 0 and go up from there. If it weren't for that then you could probably use `Date.now() * 1000 - startTime`

src/simularium/FrameRecorder.ts
interim17: Defining frame rate definitely seems sensible, but I feel like when I do I can see a subtle reduction in output quality, more blur during camera motions. It's subtle so it could be in my head, maybe I can show you tomorrow in the office? Can you think why that would be?

As for hardware acceleration, it sounds like a good thing, but I find the docs confusing. [These docs](https://www.w3.org/TR/webcodecs/#hardware-acceleration) say `Setting a value of [prefer-hardware] or [prefer-software] can significantly restrict what configurations are supported.`

I think the fall back for that failed config is just to prefer software? But if there is a risk of a configuration error is the hardware acceleration worth it? I find all my videos download really fast while leaving this unconfigured.

examples/src/RecordMovieComponent.tsx
ShrimpCryptid: Okay, keeping it as a string is fine for now! If this needs to capture more states (or if the same message string is being reused) then it's probably worth turning into an enum in the future.

examples/src/RecordMovieComponent.tsx
```diff
@@ -9,17 +9,15 @@ const RecordMovieComponent = ({
...
+    let [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
     let [outputStatus, setOutputStatus] = useState<string>("");
 
     // this useEffect is a timer that updates the recording duration
     useEffect(() => {
         let intervalId;
         if (isRecording) {
             intervalId = setInterval(() => {
-                setRecordingDuration((prevTimer) => prevTimer + 1);
+                setRecordingTimeElapsed((prevTimer) => prevTimer + 1);
```
ShrimpCryptid, line 20: One more nit on the name, if it has units (looks like seconds here?) it's a good idea to document it through a comment or the variable name itself.

examples/src/RecordMovieComponent.tsx
```diff
@@ -28,16 +26,16 @@ const RecordMovieComponent = ({
     }, [isRecording]);
 
     const startRecording = async () => {
-        setRecordingStatus("Recording...");
+        if ("VideoEncoder" in window) {
         setOutputStatus("");
-        setIsRecording(true);
+        setIsRecording(true);}
```
ShrimpCryptid, line 31: Nit: weird indentation here?

examples/src/RecordMovieComponent.tsx
```diff
@@ -28,16 +26,16 @@ const RecordMovieComponent = ({
...
+        else alert("VideoEncoder not supported in this browser")
```
ShrimpCryptid, line 32: Is an alert the standard way in the Simularium viewer to let users know about an issue?

Would it be better to just disable the recording buttons and add a tooltip letting the user know that their browser doesn't support video recording?

interim17: Ok, I'll add these! Was just being quick about these changes since this is just the test bed viewer, not a public facing front end implementation. I'm operating on the assumption that a dev using Firefox would have context for the limitations of this feature.

examples/src/RecordMovieComponent.tsx
interim17: Ok! Adding a comment.

examples/src/RecordMovieComponent.tsx
interim17: removing this anyway and adding `broswerSupported` state to disable buttons with

examples/src/RecordMovieComponent.tsx
interim17: Just added a printout instead of tooltip.

examples/src/RecordMovieComponent.tsx
```diff
@@ -9,8 +9,17 @@ const RecordMovieComponent = ({
...
     let [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
     let [outputStatus, setOutputStatus] = useState<string>("");
+    let [browserSupported, setBrowserSupported] = useState<boolean>(false);
+
+    useEffect(() => {
+        if ("VideoEncoder" in window) {
+            setBrowserSupported(true);
+        }
+    }
+    , []);
```
ShrimpCryptid, line 22: Does this need to be a state variable? Can it just be a one-line `const browserSupported = "VideoEncoder" in window`?

ShrimpCryptid (approved):
Haven't tested locally but the code changes look good. Thank you for making edits!

interim17:
> Haven't tested locally but the code changes look good. Thank you for making edits!

Thanks for review! I do want to clean up the test-bed viewer a bit so I appreciate the push to do that right.

PR #346, opened 2023-11-22
Fix/color change bug

meganrm:
Problem
=======
in working on the color picker @interim17 and @ShrimpCryptid noticed a bug with wrong color being applied

Solution
========
The fix was to check if the color changes coming into the app had been reset to defaults. But in debugging we found that the color handling was confusing. There are two different color arrays: an array where each item is a single color, and `colorsData` where each color is represented by 4 numbers, so it's 4 times as long as the actual number of colors. We identify colors by their index in the first array, but sometimes need to switch between the two different indexes, and in each case they were labeled something like `colorIndex` or just `index`. Here I renamed everything as either a `colorId` :index into the shorter array where each item is a color and `colorDataIndex`: index into the 4 timex longer array. Hopefully this will reduce future confusion. 

So the majority of this changeset is documentation and tests. 

with @interim17 

## Type of change

Steps to Verify:
----------------
1. `npm start`
2. select a model 
3. change colors, add new colors, etc, should be stable.


src/viewport/index.tsx
```diff
@@ -321,6 +329,10 @@ class Viewport extends React.Component<
                 !isEqual(
                     selectionStateInfo.colorChanges,
                     prevProps.selectionStateInfo.colorChanges
+                ) &&
+                !isEqual(
+                    selectionStateInfo.colorChanges,
+                    defaultProps.selectionStateInfo.colorChanges
```
meganrm, line 335: this was the change that fixed the bug

src/visGeometry/index.ts
```diff
@@ -1086,11 +1108,28 @@ class VisGeometry {
...
+
+    private getColorDataIndex(color: number[]): number {
+        /**
+         * returns the index into the colorsData array
+         */
         const colorArray = this.colorsData;
         const colorToCheck = map(color, (num) => round(num, 6));
         for (let i = 0; i < colorArray.length - 3; i += 4) {
-            const index = i / 4;
+            const index = i;
```
meganrm, line 1132: this was also a bug

src/visGeometry/index.ts
```diff
@@ -1136,54 +1213,39 @@ class VisGeometry {
...
-         * @param id agent id
-         * @param colorId index into the color array
-         */
-        this.idColorMapping.set(id, colorId);
-
-        // if we don't have a mesh for this, add a sphere instance to mesh registry?
-        if (!this.visGeomMap.has(id)) {
-            this.visGeomMap.set(id, DEFAULT_MESH_NAME);
-        }
-    }
```
meganrm, line 1166: these are just moved up, I grouped all the private methods for handling color together, followed by all the public methods

src/visGeometry/index.ts
```diff
@@ -1136,54 +1213,39 @@ class VisGeometry {
...
+        /**
+         * Sets one color for a set of ids
+         */
         this.setColorForIds(agentIds, colorId);
         this.updateScene(this.currentSceneAgents);
     }
 
-    public getColorForIndex(index: number): Color {
+    public getColorForColorId(colorId: number): Color {
+        if (colorId < 0) {
```
meganrm, line 1236: I think these cases will be impossible to hit. I wrote them as I was thinking of edge cases writing tests.

src/viewport/index.tsx
toloudis: I wonder if this function would be more performant if we updated to newer react and use useEffect.  I always wonder if the number of checks here could bog things down.

src/visGeometry/index.ts
```diff
@@ -1060,21 +1060,43 @@ class VisGeometry {
         }
     }
 
+    /**
+     * Agent color handling
+     * @property this.colorsData is an array of floats, each 4 floats is a color
+     * @param dataColorIndex is an index into the colorsData array, so it is a multiple of 4
+     * @param colorId is an index into the numberOfColors, so it is a number between 0
```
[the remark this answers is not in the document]
meganrm: yeah this comment is for the whole section of "agent color handling". I do agree it should be its own module (in addition to a lot of other things in visGeometry).  I was consistent in all the following functions when using these two different names for parameters instead of generic things like "index" or "colorIndex" that we had before. I thought it would be clearer here than having the same comments for every function that used one of these

src/viewport/index.tsx
meganrm: it would be clearer to read, but I imagine at some level hooks are doing all these checks anyway to know if they should run. My understanding is class components are equally performant to function/hook based components in a one to one comparison. you can get increases in performance if you manage to make your overall tree structure less nested which is one of the benefits

src/visGeometry/index.ts
```diff
@@ -1136,54 +1213,39 @@ class VisGeometry {
...
-        ids.forEach((id) => this.setColorForId(id, colorId));
+        ids.forEach((id) => {
+            this.setColorForId(id, colorId);
+        });
     }
 
     public applyColorToAgents(agentIds: number[], colorId: number): void {
+        /**
+         * Sets one color for a set of ids
+         */
```
toloudis, line 1230: jsdoc style comments should go above the function name and not below it.

src/visGeometry/index.ts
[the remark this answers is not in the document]
toloudis: It's the `@param` tags that are not helping.  They look like jsdoc tags for autogenerated documentation and you can't tell what they are parameters for.  I think it would be clearer to have each function get its own jsdoc if you want to leave the code here in the middle of this module.

src/visGeometry/index.ts
```diff
@@ -1060,21 +1060,43 @@ class VisGeometry {
...
+     * long as the input colors. They should only know about colorId
+     * Therefore all the public color methods should use colorId, and the private methods
+     * can convert between colorId and dataColorIndex
+     */
+
+    /**
+     * Agent color handling: private methods
+     */
+
+    private get numberOfColors(): number {
```
toloudis, line 1081: Nit: I feel this would be more readable when it is used later in the code, if it were a simple function.  It always messes me up when I think it's a simple variable and then wonder, who is updating its value when we add a new color??

src/test/VisGeometry.test.ts
```diff
@@ -0,0 +1,125 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* NOTE: this file is testing many private
+    methods and properties of VisGeometry.ts
+    which we do by using (visGeometry as any) 
+    so we need to disable the eslint rule here
+*/
```
ShrimpCryptid, line 6: I didn't know you could do this! Is this the preferred way to do testing of private members?

src/visGeometry/index.ts
```diff
@@ -1060,21 +1060,44 @@ class VisGeometry {
         }
     }
 
+    /**
+     * AGENT COLOR HANDLING (TODO: move to separate file)
+     * General notes about data being used to map color to agents:
+     * @property this.colorsData is an array of floats, each 4 floats is a color
+     *  `dataColorIndex` is always an index into the colorsData array, so it is a multiple of 4
+     * `colorId` is always an index into the numberOfColors, so it is a number between 0
```
ShrimpCryptid, line 1068: Nit: The phrase `index into the numberOfColors` seems wrong here, since it suggests that `numberOfColors` is an array?

src/visGeometry/index.ts
```diff
@@ -1104,6 +1144,44 @@ class VisGeometry {
         return -1;
     }
 
+    private getColorIdForTypeId(typeId: number): number {
+        /**
+         * returns the index into numberOfColors (colorId). No conversion
```
ShrimpCryptid, line 1149:
```suggestion
         * returns the index in terms of numberOfColors (colorId). No conversion
```

ShrimpCryptid: or index into `idColorMapping`?

src/visGeometry/index.ts
```diff
@@ -1104,6 +1144,44 @@ class VisGeometry {
...
+         * is necessary because idColorMapping is also using this index
+         */
+        const index = this.idColorMapping.get(typeId);
+        if (index === undefined) {
+            this.logger.error(
+                "getColorIndexForTypeId could not find " + typeId
```
ShrimpCryptid, line 1155:
```suggestion
                "getColorIdForTypeId could not find " + typeId
```

src/visGeometry/index.ts
```diff
@@ -1112,18 +1190,20 @@ class VisGeometry {
...
+        // the color isn't in colorsData, so add it and return the colorId
+        const newColorDataIndex = this.colorsData.length;
         const newArray = [...this.colorsData, ...newColor];
         const newColorData = new Float32Array(newArray.length);
         newColorData.set(newArray);
         this.colorsData = newColorData;
-        this.renderer.updateColors(this.colorsData.length / 4, this.colorsData);
-        return newIndex;
+        this.renderer.updateColors(this.numberOfColors, this.colorsData);
+        return this.convertDataColorIndexToId(newColorDataIndex);
```
ShrimpCryptid, line 1206: Are we worried about performance issues here?

Like, if I keep setting different colors for agents and then changing them, all of the intermediate colors would get added to `this.colorsData`. I don't think the array itself would cause issues (since it would take up like, 67 MB max if ignoring alpha values), but searching through it to find color indices later might be slower since it could potentially have up to 16 million entries.

src/visGeometry/index.ts
```diff
@@ -1136,54 +1216,39 @@ class VisGeometry {
...
+            colorId = colorId % this.numberOfColors;
+        }
+
         return new Color(
-            this.colorsData[index * 4],
-            this.colorsData[index * 4 + 1],
-            this.colorsData[index * 4 + 2]
+            this.colorsData[colorId * 4],
+            this.colorsData[colorId * 4 + 1],
+            this.colorsData[colorId * 4 + 2]
```
ShrimpCryptid, line 1251: Should this return the alpha channel, since we're already storing it anyways?

src/test/VisGeometry.test.ts
meganrm: This was just a stack overflow suggestion, I'm not sure if we have a preferred way of testing private methods.

toloudis: I don't know if we have standardized on a way to do this across many projects.  IMO this is very acceptable but a tradeoff:
(1) It's fair game to do this to simply unit test functions that are supposed to be private/internal.   It lets you get at those functions without having to modify the actual code.  Idk if typescript has other clean ways to do it (that would be more lintable).
(2) if you are trying to test the public api then this could just confuse things about what is a class's intended contract to the outside world

src/visGeometry/index.ts
meganrm: yeah it's a good question. I don't think there is any real issue with just adding a new color every time we change color even if it already exists. My thinking was around keeping the data consistent ie, what if we wanted to select all agents that have the same color? But we don't currently have a use case for that. I could go either way on this

toloudis (approved):
The code changes look good.  

There are still a bunch of jsdoc function comments in here which are below the function declaration, which is more python-style than js style and not what jsdoc recommends.

We need a new ticket to move this color handling code out to its own module.

src/visGeometry/index.ts
meganrm: this `Color` from THREE js doesn't accept 4th term. but it is a question for when we're actually supporting transparency

src/test/VisGeometry.test.ts
meganrm: regarding #2: should I separate out a file that is only the private methods? I think there is only one public method I test here

src/visGeometry/index.ts
toloudis: I would expect the array of color data that gets passed to the shader, to be the same length as the number of agent types...keeping that small and efficient is good for the shader that uses it.  But it will only be "bad" in the case Peyton describes where people start putting a lot of unused colors in the array and it gets to like more than 2-3x the number of agents and we get extra cache misses on lookups and just have dead data lying around for no reason.

src/test/VisGeometry.test.ts
toloudis: If you feel like it.  I am not worried about it for this code.

meganrm:
> The code changes look good.
> 
> There are still a bunch of jsdoc function comments in here which are below the function declaration, which is more python-style than js style and not what jsdoc recommends.
> 
> We need a new ticket to move this color handling code out to its own module.

I made one: #347

src/visGeometry/index.ts
```diff
@@ -1136,54 +1216,39 @@ class VisGeometry {
...
-            this.visGeomMap.set(id, DEFAULT_MESH_NAME);
-        }
-    }
-
     public setColorForIds(ids: number[], colorId: number): void {
         /**
          * Sets one color for a set of ids, using an index into a color array
          * @param ids agent ids that should all have the same color
-         * @param colorId index into the color array
+         * @param colorId index into the numberOfColors
```
ShrimpCryptid, line 1223: Same thing here for indexing into

ShrimpCryptid, line 1223:
```suggestion
         * @param colorId index in terms of numberOfColors
```

ShrimpCryptid (approved):
Looks good! Thanks for the discussion on the points I brought up.

PR #349, opened 2023-12-01
Feature/server health check

ascibisz:
Problem
=======
The server health check has existed in Octopus for a while now, time to take it in to the viewer!

One part of this [ticket](https://github.com/simularium/simularium-website/issues/389). The other piece of that ticket will include adding a health check call, similar to the one implemented here in the example viewer, to the simualrium-website in the early stages of the autoconversion form.

Also added in support for handling error messages octopus sends now, which is another new message type that has existed in Octopus for a bit.

Solution
========
Paired on this with with @interim17 

## Type of change

Change summary:
---------------
* Added handling for new message types that come from octopus: `ID_ERROR_MSG = 21` and `ID_SERVER_HEALTHY_RESPONSE = 23`
* Added ability to request health check, using new message type `ID_CHECK_HEALTH_REQUEST = 22`, which octopus already knows how to handle
* Added an example of using the health check request / response to the example viewer. Now, when you hit the "Load a smoldyn trajectory" button, a websocket connection will be initiated and health check request will be sent to octopus. The "submit" button will remain disabled until a server healthy response comes back
* As part of the changes to the example, I updated the autoconversion request call to only initiate a new websocket connection if there isn't already one there. Previously, we just automatically created a new websocket connection, but in the case that one may already exist from the server health check, we should just reuse that one

### DIFF
--- a/src/controller/index.ts
+++ b/src/controller/index.ts
@@ -235,9 +235,7 @@ export default class SimulariumController {
     ): Promise<void> {
         try {
             if (
-                !this.simulator ||
-                !(this.simulator instanceof RemoteSimulator) ||
-                !this.simulator.socketIsValid()
+                !(this.simulator && this.simulator.isConnectedToRemoteServer())
             ) {
                 // Only configure network if we aren't already connected to the remote server
                 this.configureNetwork(netConnectionConfig);
@@ -423,15 +421,11 @@ export default class SimulariumController {
         handler: () => void,
         netConnectionConfig: NetConnectionParams
     ): void {
-        if (
-            !this.simulator ||
-            !(this.simulator instanceof RemoteSimulator) ||
-            !this.simulator.socketIsValid()
-        ) {
+        if (!(this.simulator && this.simulator.isConnectedToRemoteServer())) {
             // Only configure network if we aren't already connected to the remote server
             this.configureNetwork(netConnectionConfig);
         }
-        if (this.simulator && this.simulator instanceof RemoteSimulator) {
+        if (this.simulator instanceof RemoteSimulator) {
             this.simulator.setHealthCheckHandler(handler);
             this.simulator.checkServerHealth();
         }
--- a/src/controller/index.ts
+++ b/src/controller/index.ts
@@ -234,7 +234,14 @@ export default class SimulariumController {
         fileType: TrajectoryType
     ): Promise<void> {
         try {
-            this.configureNetwork(netConnectionConfig);
+            if (
+                !this.simulator ||
+                !(this.simulator instanceof RemoteSimulator) ||
+                !this.simulator.socketIsValid()
+            ) {
+                // Only configure network if we aren't already connected to the remote server
+                this.configureNetwork(netConnectionConfig);
+            }
             if (!(this.simulator instanceof RemoteSimulator)) {
                 throw new Error("Autoconversion requires a RemoteSimulator");
             }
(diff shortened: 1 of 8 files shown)


src/simularium/WebsocketClient.ts
```diff
@@ -313,7 +316,7 @@ export class WebsocketClient {
         jsonData: Record<string, unknown>,
         requestDescription: string
     ): void {
-        if (this.socketIsValid()) {
+        if (this.socketIsConnected()) {
```
ascibisz, line 319: I ran into this issue when testing where the socket can be "valid" (which means the websocket's state isn't CLOSED), but not "connected" (which means the websocket's state is OPEN) if the state is PENDING, in which case sending a message throws an error. I feel like this check is trying to prevent that error, so might as well cover the PENDING case too?

src/controller/index.ts
```diff
@@ -412,6 +419,24 @@ export default class SimulariumController {
...
+        handler: () => void,
+        netConnectionConfig: NetConnectionParams
+    ): void {
+        if (
+            !this.simulator ||
+            !(this.simulator instanceof RemoteSimulator) ||
+            !this.simulator.socketIsValid()
+        ) {
+            // Only configure network if we aren't already connected to the remote server
+            this.configureNetwork(netConnectionConfig);
```
interim17, line 432: It feels a bit funny to have to pass the config in every time we make a check, it looks redundant on the front-end, but it does ensure we aren't querying an un-configured controller that won't respond.

interim17 (approved):

src/controller/index.ts
toloudis: Yeah it does sort of make sense that you might want to check health before a connection was made.

src/controller/index.ts
```diff
@@ -412,6 +419,24 @@ export default class SimulariumController {
...
+        }
+        if (this.simulator && this.simulator instanceof RemoteSimulator) {
+            this.simulator.setHealthCheckHandler(handler);
+            this.simulator.checkServerHealth();
```
toloudis, line 436: should checkServerHealth accept a callback as an arg instead of having a separate setter?  Are they always called together like this?

src/simularium/RemoteSimulator.ts
```diff
@@ -134,6 +142,17 @@ export class RemoteSimulator implements ISimulator {
         // TODO: implement callback
     }
 
+    public onErrorMsg(msg): void {
```
toloudis, line 145: linter wants msg to have a type

src/controller/index.ts
toloudis: I see why they are separate now, because of the message passing and receiving.

src/simularium/WebsocketClient.ts
```diff
@@ -47,6 +47,9 @@ export const enum NetMessageEnum {
     ID_AVAILABLE_METRICS_RESPONSE = 18,
     ID_PLOT_DATA_REQUEST = 19,
     ID_PLOT_DATA_RESPONSE = 20,
+    ID_ERROR_MSG = 21,
+    ID_CHECK_HEALTH_REQUEST = 22,
+    ID_SERVER_HEALTHY_RESPONSE = 23,
```
toloudis, line 52: is there any documentation in this repo for what these messages are supposed to mean, or their payload schema?  (One thing that is nice to know is whether it's intended to go from client to server or vice versa)

toloudis (approved):
I left a bunch of comments but basically this seems ok.  I would love to clean up the connection logic but probably out of scope for this.

src/simularium/WebsocketClient.ts
ascibisz: There isn't... I do have it fairly well documentation in the Octopus repo though, so I should probably copy that over

ascibisz: Yeah and that's true of the message types themselves. We're already in a slightly weird situation, may as well have it be a better documented weird situation!

src/simularium/RemoteSimulator.ts
```diff
@@ -412,4 +446,19 @@ export class RemoteSimulator implements ISimulator {
...
+                this.webSocketClient.sendWebSocketRequest(
+                    {
+                        msgType: NetMessageEnum.ID_CHECK_HEALTH_REQUEST,
+                    },
+                    "Request server health check"
+                );
+            })
+            .catch((e) => {
+                throw new FrontEndError(e.message, ErrorLevel.ERROR);
+            });
```
interim17, line 462: Questions about error handling.

UX request is that if "server is down" we throw a styled modal.

But if the viewer can't connect this error also generates up a full screen React runtime error.

Removing this error handling yields the behavior UX asked for (just a modal), but is that not proper error handling?

PR #351, opened 2023-12-02
install @types/babel__traverse@7.18.3

interim17:
Problem
=======
I found that I could not build the viewer from main because our `typescript` version was out of sync with `@types/babel__traverse` Found some other folks in github threads facing a similar problem

Installing `@types/babel__traverse@7.18.3` fixes it.

Here's the error text I was seeing if you would like to compare to your experience:
```
node_modules/@types/babel__traverse/index.d.ts:314:10 - error TS1023: An index signature parameter type must be either 'string' or 'number'.

314         [k: `${string}|${string}`]: VisitNode<S, Node>;
             ~


Found 1 error.

ERROR: "generateTypes" exited with 2.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
         "@fortawesome/react-fontawesome": "^0.1.14",
         "@tweakpane/core": "^1.0.8",
         "@tweakpane/plugin-essentials": "^0.1.4",
+        "@types/babel__traverse": "^7.18.3",
         "comlink": "^4.3.1",
         "js-logger": "^1.6.1",
         "parse-pdb": "^1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -102,6 +102,7 @@
         "@fortawesome/react-fontawesome": "^0.1.14",
         "@tweakpane/core": "^1.0.8",
         "@tweakpane/plugin-essentials": "^0.1.4",
+        "@types/babel__traverse": "^7.18.3",
```
meganrm, line 105: this should be in dev dependancies

meganrm:
why is it babel_traverse and not [@babel/traverse](https://github.com/babel/babel/tree/HEAD/packages/babel-traverse)?

toloudis:
I'd love to see us just update all dev dependencies wholesale to the latest major.minor versions (except for react, antd and whatever else might incur major code changes).  This feels like a bandaid somewhat but it also seems pretty harmless!

toloudis:
@interim17 can you see if #354 fixed this build issue?

interim17:
@toloudis yes it fixes it! appreciate the discussion it's useful for me.

interim17:
#354 fixed this!

Issue #347, opened 2023-11-28
Move color handling into it's own module 


meganrm:
## What needs to happen?
Make a module that handles assigning and storing color data 

## Why should we do this?
It's all in visGeometry currently and the file is bloated

PR #352, opened 2023-12-04
Feature/color module

meganrm:
Problem
=======
#347

Solution
========
I moved all the functions in visGeometry into a colorHandling class. In the process I also simplified the `colorChanges` which was an array, but we're only ever sending in one `colorChange` at a time. I updated https://github.com/simularium/simularium-website/pull/432 for this change too. 

VisGeometry still has some color handling functions to unite the color with the visAgent, but it now only deals with colors by their names (no indices). 

## Type of change

Steps to Verify:
----------------
1. npm start
2. change an agent color using the drop downs.


src/visGeometry/ColorHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+    return "#" + new Color(color).getHexString();
+}
+
+export const checkHexColor = (color: string): string => {
+    const hexColorCodeRegex = /^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/;
+    if (!hexColorCodeRegex.test(color)) {
+        throw new Error("Invalid color code");
+    }
+    return color;
+};
```
meganrm, line 25: These functions were previously in `color-utils`. I thought they made sense here, but I didn't make them methods on the class

src/visGeometry/ColorHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+
+class ColorHandler {
+    public idColorMapping: Map<number, number>;
+    private colorsData: Float32Array;
+    private renderer: SimulariumRenderer;
+
+    constructor(colors: (number | string)[], renderer: SimulariumRenderer) {
+        this.idColorMapping = new Map<number, number>();
+        this.colorsData = this.setColorArray(colors);
+        this.renderer = renderer;
```
meganrm, line 35: I wasn't really sure about storing the renderer. It does make referencing it easier, but I could also just send in the renderer to each method when their called. I tried moving everything that used the renderer into here, but there was one function that made that difficult. Let me know what you think

toloudis:
About the change to only accept a single color at a time, I think we used to have in mind bulk color changes like "palettes".  That's the only reason I can remember ever needing to change multiple colors in one function call in an optimized way.

meganrm:
> About the change to only accept a single color at a time, I think we used to have in mind bulk color changes like "palettes". That's the only reason I can remember ever needing to change multiple colors in one function call in an optimized way. (And of course we never did it)

Yeah, I think in that case we would wipe everything out, ie, re-init the colorHandler with the new pallet, and then just call 'setInitialColors' again.

src/visGeometry/ColorHandler.ts
toloudis: I would say this class should know nothing about the renderer.  I would not have this class call any update functions on the renderer - it should just provide an accessor function to the raw data array and there are only one or two places in visGeometry/index.ts where you need to ask for that array and pass it to the renderer method.

src/visGeometry/VisAgent.ts
```diff
@@ -65,9 +68,9 @@ export default class VisAgent {
         };
     }
 
-    public setColor(color: Color, colorIndex: number): void {
-        this.color = color;
-        this.colorIndex = colorIndex;
+    public setColor(colorInfo: { color: Color; colorId: number }): void {
```
toloudis, line 71: what is this change doing? Is this an object that you pass around in other places?  I will read on...

src/visGeometry/index.ts
```diff
@@ -718,8 +717,9 @@ class VisGeometry {
             const visAgent = this.visAgents[i];
             if (typeIds.includes(visAgent.agentData.type)) {
                 visAgent.setColor(
-                    this.getColorForTypeId(visAgent.agentData.type),
-                    this.getColorIdForTypeId(visAgent.agentData.type)
+                    this.colorHandler.getColorInfoForAgentType(
+                        visAgent.agentData.type
+                    )
```
toloudis, line 722: I see what you did here. Maybe this deserves a named type like AgentColorInfo { color: Color, id: number}

src/visGeometry/VisAgent.ts
meganrm: I made this change so that I could call one function in the `colorHandler` that returns both things needed for this function. Before calls to `setColor` were calling two functions that each returned one of the two arguments. I know I could have spread out the return into arguments so I didn't need to change this function, but I think that looks more confusing

src/visGeometry/index.ts
```diff
@@ -1066,198 +1066,37 @@ class VisGeometry {
...
-            const index = i;
-            const currentColor = [
-                round(this.colorsData[i], 6),
-                round(this.colorsData[i + 1], 6),
-                round(this.colorsData[i + 2], 6),
-                this.colorsData[i + 3],
-            ];
-            if (isEqual(currentColor, colorToCheck)) {
-                return index;
+    public setGeom(ids: number[]): void {
```
toloudis, line 1077: should this be private?

toloudis: I think this needs to be clearer as to why anyone would ever call it.  Geometry should have been initialized in a better way elsewhere.

src/visGeometry/index.ts
```diff
@@ -1066,198 +1066,37 @@ class VisGeometry {
...
-
-    private get numberOfColors(): number {
-        return this.colorsData.length / 4;
-    }
-
     private setAgentColors(): void {
         this.visAgents.forEach((agent) => {
             agent.setColor(
-                this.getColorForTypeId(agent.agentData.type),
-                this.getColorIdForTypeId(agent.agentData.type)
```
meganrm, line 1096: @toloudis this is the result of changing the interface to `setColor`. Since each of these functions are receiving the same data it made sense to me to make a single function

toloudis: Yup I think the usage pattern makes sense but I would prefer a named type for this bundle of color info.

src/visGeometry/index.ts
meganrm: yeah this was previously part of : 
```  private setColorForId(id: number, colorId: number): void {
         /**
          * @param id agent id
          * @param colorId index into the numberOfColors
          */
         this.idColorMapping.set(id, colorId);

         // if we don't have a mesh for this, add a sphere instance to mesh registry?
         if (!this.visGeomMap.has(id)) {
             this.visGeomMap.set(id, DEFAULT_MESH_NAME);
         }
     }
```
I factored out the the part that sets the colorMapping, but I wasn't sure about this other part. You're right it should be private, but it also doesn't seem necessary, (we default things to spheres in the geometryStore). It might have gotten left behind accidentally during a refactor

toloudis: I see, yeah- even better if we can get away with removing it.

meganrm: I removed it

src/visGeometry/ColorHandler.ts
```diff
@@ -0,0 +1,206 @@
...
+        };
+    }
+
+    public getColorForAgentType(agentType: number): Color {
+        const index = this.getColorIdForAgentType(agentType);
+        return this.getColorById(index);
+    }
+
+    public getColorInfoForAgentType(agentType: number): AgentColorInfo {
+        const color = this.getColorForAgentType(agentType);
```
toloudis, line 200: Nit: getColorForAgentType also calls getColorIdForAgentType, and then you call it again on the next line.. so maybe there is a small refactor/cleanup to remove this redundancy. Not sure tho.

meganrm: yeah, they were essentially the same function with one returning less data. I changed the places were the parent module only wanted the color (and not the colorId) to just unpack the one variable from `AgentColorInfo`

src/visGeometry/ColorHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+            colorsData[i * 4 + 0] =
+                ((colorNumbers[i] & 0x00ff0000) >> 16) / 255.0;
+            colorsData[i * 4 + 1] =
+                ((colorNumbers[i] & 0x0000ff00) >> 8) / 255.0;
+            colorsData[i * 4 + 2] =
+                ((colorNumbers[i] & 0x000000ff) >> 0) / 255.0;
+            colorsData[i * 4 + 3] = 1.0;
+        }
+        return colorsData;
+    }
```
ShrimpCryptid, line 58: Should this be `makeColorDataArray`? `set` implies it's setting an internal value, which it doesn't seem like it's doing.

src/visGeometry/ColorHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+export const checkHexColor = (color: string): string => {
+    const hexColorCodeRegex = /^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/;
+    if (!hexColorCodeRegex.test(color)) {
+        throw new Error("Invalid color code");
+    }
+    return color;
+};
+
+class ColorHandler {
+    public idColorMapping: Map<number, number>;
```
ShrimpCryptid, line 28: I feel like this could either use a comment or a more descriptive rename, like `agentIdToColorId` or something.

ShrimpCryptid (approved):
Looks good, left a few notes about improving code readability 👍

interim17 (approved):

PR #355, opened 2024-01-08
Binding simulation

meganrm:
Problem
=======
I built a prototype for the education module. First here, and then in a new repo. In order to install the viewer and make use of the front end simulation interface, I had to make a number of changes. 
#### Issues
1. the other repo isn't using babel
2. If you remount the viewer component, it parents a second canvas (only known because of `React.StrictMode`)
3. A number of types/interfaces were needed by the installing app but weren't exposed 
4. The other app couldn't resolve the urls for the web workers  

Solution
========
1. moved some runtime babel deps from dev deps into regular 
2. Added check to see of the canvas had already been added to the parent div
3. Exported interfaces 
4. added `{ type: "module" }` for the URL declarations 

## Type of change

Steps to Verify:
----------------
Since these changes are largely to enable installation by another app, to truly test it you need to also download the [other repo](https://github.com/simularium/binding-sim-edu) and install this branch as the viewer. Short of that, everything else in the test bed app should behave normally.


package.json
```diff
@@ -92,17 +91,21 @@
         "webpack-dev-server": "^4.5.0"
     },
     "peerDependencies": {
-        "react": "^16.9.0",
-        "react-dom": "^16.9.0"
+        "react": "^18.2.0",
+        "react-dom": "^18.2.0"
```
toloudis, line 95: how does bumping the react version affect simularium-website?

package.json
```diff
@@ -92,17 +91,21 @@
...
+        "@babel/runtime": "^7.23.6",
         "@fortawesome/fontawesome-free": "^5.15.2",
         "@fortawesome/fontawesome-svg-core": "~1.2.34",
         "@fortawesome/free-solid-svg-icons": "^5.15.2",
         "@fortawesome/react-fontawesome": "^0.1.14",
         "@tweakpane/core": "^1.0.8",
         "@tweakpane/plugin-essentials": "^0.1.4",
+        "@types/react-dom": "^18.2.18",
         "comlink": "^4.3.1",
+        "detect-collisions": "^9.2.2",
```
toloudis, line 108: this probably should not be a "real" dependency since it is only used by a simulator class that lives in examples, and is not a dependency of the actual simularium-viewer.  (Do we need a separate package.json for the test app to capture that dependency?)

package.json
```diff
@@ -92,17 +91,21 @@
...
     "dependencies": {
+        "@babel/plugin-transform-runtime": "^7.23.6",
+        "@babel/runtime": "^7.23.6",
         "@fortawesome/fontawesome-free": "^5.15.2",
         "@fortawesome/fontawesome-svg-core": "~1.2.34",
         "@fortawesome/free-solid-svg-icons": "^5.15.2",
         "@fortawesome/react-fontawesome": "^0.1.14",
         "@tweakpane/core": "^1.0.8",
         "@tweakpane/plugin-essentials": "^0.1.4",
+        "@types/react-dom": "^18.2.18",
```
toloudis, line 106: are types always a devDependency?

src/simularium/VisData.ts
```diff
@@ -237,7 +237,8 @@ class VisData {
 
     private setupWebWorker() {
         this.webWorker = new Worker(
-            new URL("../visGeometry/workers/visDataWorker", import.meta.url)
+            new URL("../visGeometry/workers/visDataWorker", import.meta.url),
+            { type: "module" }
```
toloudis, line 241: Slight concern that this modification can affect other clients like simularium-website etc. Does it work across-the-board?

src/visGeometry/index.ts
```diff
@@ -845,9 +845,12 @@ class VisGeometry {
         if (parent === undefined || parent == null) {
             return;
         }
-
+        if (parent["data-has-simularium-viewer-canvas"]) {
+            return;
+        }
         this.threejsrenderer = this.createWebGL();
         parent.appendChild(this.threejsrenderer.domElement);
+        parent["data-has-simularium-viewer-canvas"] = true;
```
toloudis, line 853: I like this, this is a nice fix.  I still kinda think it could do more intelligent checking (parent!==canvas.parent) and other stuff but it gets more complicated quickly.

package.json
meganrm: everything seems to work fine, but it will necessitate us bumping the react version on the front end too. I was able to do it pretty quickly on the front end with a local installation. The only other dependency that had to be changed was a "react-waypoints" dep.

package.json
meganrm: that's my understanding since typescript isn't runtime

toloudis: then this line can move up out of this section ⬆️

package.json
toloudis: if you make this look too easy, you will remove any excuses we have for not updating other codebases

package.json
meganrm: OOH yeah, I missed that, thanks

src/simularium/VisData.ts
meganrm: I've tested the website and it seems fine. nbsv actually also had an issue on build that couldn't find the web workers. I'm not sure if this change would resolve it or not

package.json
toloudis: ideally yes.  I don't know if it's worth it but yeah we are building and bundling two projects really.  One for the library that is the viewer and the other for the test app.

meganrm: ok, I moved the scripts, the webpack file, the dev deps all into examples. I also created a `src` in examples to add a little organization, so it ends up looking like a lot of file changes

examples/src/Viewer.tsx
```diff
@@ -45,10 +46,10 @@ import {
     UI_TEMPLATE_URL_ROOT,
 } from "./api-settings";
 import ConversionForm from "./ConversionForm";
-import MetaballSimulator from "./MetaballSimulator";
-import { TrajectoryType } from "../src/constants";
+import MetaballSimulator from "./simulators/MetaballSimulator";
+import { TrajectoryType } from "../../src/constants";
```
toloudis, line 50: nit/cleanup: I suggest moving this line up closer to the other imports from ../../src

examples/src/index.tsx
```diff
@@ -3,6 +3,9 @@ import React from "react";
 
 import Viewer from './Viewer';
 
+declare const SIMULARIUM_USE_OCTOPUS: boolean;
+declare const SIMULARIUM_USE_LOCAL_BACKEND: boolean;
```
toloudis, line 7: Is this a way to tell the linters etc that these vars will be injected by external means?

examples/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "testbed_viewer",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "serve-example": "webpack serve --config webpack.dev.js",
```
toloudis, line 7: because it's such a standard convention, should we have a `start` script in there too?

package.json
```diff
@@ -60,28 +59,24 @@
...
         "npm-run-all": "^4.1.5",
         "nyc": "~15.0",
         "postcss": "^8.2.10",
         "postcss-cli": "^8.3.1",
         "postcss-preset-env": "^6.7.0",
         "prettier": "^2.2.1",
-        "react": "^16.9.0",
-        "react-dom": "^16.9.0",
+        "react": "^18.2.0",
+        "react-dom": "^18.2.0",
```
toloudis, line 79: So theoretically you can remove all traces of webpack from this package.json now that only the example app needs it

toloudis (approved):
This is looking great, thank you for the extra reorganization

interim17:
LGTM, ran the binding sim in the test bed and it looked good, installed and ran `binding-sim-edu` fine after fussing with `bunx` and `link` a bit

interim17 (approved):

package.json
meganrm: oh you're right, I thought we were using that `webpack.config.js` for something, so I didn't remove it but I guess not, I'll make another ticket to do that.

examples/package.json
meganrm: you have to run the build process in the root directory first before you can serve this. My thinking was that a developer would always be running things from the root directory to avoid confusion of moving up and down directories

examples/src/index.tsx
meganrm: yes

PR #363, opened 2024-02-07
Feature/freeze camera

meganrm:
Problem
=======
The education module will not need camera controls because it's a 2d simulation 

Solution
========
I added a prop called `stationaryViewport` that defaults to `false`. 

NOTES/QUESTIONS:

1. I made the assumption here that this setting is unlikely to change for a simulation. It's set when the canvas is mounted and nothing would happen if the prop changed but the canvas isn't remounted. I can add the ability to have it react to the prop change, but it's not needed for the education simulation and so I'm partial to waiting until it's a requirement for something else but open to changing if other people feel differently. 
2. any ideas for a clearer prop name are welcome. I could also see the argument for having a `movableViewport` prop that defaults to `true`

## Type of change

Steps to Verify:
----------------
1. `npm start`
2. change `stationaryViewport` to true in the example viewer 
3. camera controls shouldn't work and nothing should happen when you click on an agent.


src/visGeometry/index.ts
```diff
@@ -841,7 +849,10 @@ class VisGeometry {
         return this.threejsrenderer;
     }
 
-    public reparent(parent?: HTMLElement | null): void {
+    public setCanvasOnTheDom(
```
meganrm, line 852: I changed the name of this function because it's doing more than reparenting. If you have a better name please let me know

src/viewport/index.tsx
```diff
@@ -222,8 +224,12 @@ class Viewport extends React.Component<
             simulariumController,
             loadInitialData,
             onError,
+            stationaryViewport,
         } = this.props;
-        this.visGeometry.reparent(this.vdomRef.current);
+        this.visGeometry.setCanvasOnTheDom(
+            this.vdomRef.current,
+            stationaryViewport
+        );
```
interim17, line 232: Very good change, was just looking at `reparent` when debugging mp4 recorder and this would have saved me time. Makes a search for 'canvas' in the repo more productive.

src/visGeometry/index.ts
```diff
@@ -851,7 +862,7 @@ class VisGeometry {
         this.threejsrenderer = this.createWebGL();
         parent.appendChild(this.threejsrenderer.domElement);
         parent["data-has-simularium-viewer-canvas"] = true;
-        this.setupControls();
+        this.setupControls(disableControls);
```
interim17, line 865: I just added `setAllowViewPanning` to disable just panning in NBSV and now we are adding this wholesale method to disable all controls.

Out of scope here, but if front end implementations continue to have requirements about what controls should be allowed, maybe we should consolidate and `setupControls` should take a configuration object so the front end can be explicit about what controls it wants to have active, with current standard controls as default?

src/visGeometry/index.ts
interim17: Really like the name change. Fixing the viewport once on mount seems good for now, but I can definitely imagine a use case for exposing `setupControls` for reactive/ongoing configuration.

interim17:
LGTM, left some ideas about future work on configuring controls.

I don't mind `stationaryViewport` but I might prefer `freezeCamera`,`disableCameraMovement`, `fixedCameraPosition`(and flip the boolean) or something like that, since we are preventing camera movement within the viewport. Not a strong preference though, fine with me to leave as is.

interim17 (approved):

toloudis:
I saw that you don't allow the "follow" behavior.  Do you still want to allow mouse clicking to pick an agent?

toloudis:
> I don't mind `stationaryViewport` but I might prefer `freezeCamera`,`disableCameraMovement`, `fixedCameraPosition`(and flip the boolean) or something like that, since we are preventing camera movement within the viewport. Not a strong preference though, fine with me to leave as is.

or maybe `lockedCamera`...

meganrm:
> I saw that you don't allow the "follow" behavior. Do you still want to allow mouse clicking to pick an agent?

It doesn't really matter if an agent is "picked" as long as we're not following it. For completeness are you saying I should also make it so click on an agent doesn't save the id?

toloudis:
> > I saw that you don't allow the "follow" behavior. Do you still want to allow mouse clicking to pick an agent?
> 
> It doesn't really matter if an agent is "picked" as long as we're not following it. For completeness are you saying I should also make it so click on an agent doesn't save the id?

I don't mind either way but it can result in the agent being selected and outlined.  Just want to make sure we are intentional. My comment was just about being able to document (in the code in comments near where the prop is defined, or something) to know exactly what is supposed to be locked down when this prop is set.

toloudis (approved):

PR #365, opened 2024-03-04
Fix/create root

interim17:
Test-bed viewer was not actually running React 18 because the render API hadn't been switched over to `createRoot`, meaning we were in React 17 legacy mode, this makes that change.

Note: codec issues solved in separate PR, this is just the React change now

### DIFF
--- a/examples/src/index.tsx
+++ b/examples/src/index.tsx
@@ -1,12 +1,17 @@
-import ReactDOM from "react-dom";
 import React from "react";
+import { createRoot } from "react-dom/client";
 
 import Viewer from './Viewer';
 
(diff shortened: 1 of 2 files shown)


toloudis:
I wonder if it could have been a browser version update that happened coincidentally at the same time

interim17:
> I wonder if it could have been a browser version update that happened coincidentally at the same time

Couldn't be me nervously updating Chrome...

Issues persist, but I think I'm on to a (partial) fix for the codec.

blairlyons (approved):

meganrm (approved):

PR #367, opened 2024-03-07
need to install example deps

meganrm:
Time Estimate or Size
=======
small 

Problem
=======
the nightly build wasn't working because the dependencies for the app weren't being installed 

Solution
========
added a pre-script script

## Type of change

Steps to Verify:
----------------
1. delete node_modules, package-lock, example/node_modules, example/package-lock
2. run `npm install`
3. run `npm run gh-build`

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "serve-example": "webpack serve --config webpack.dev.js",
+    "prebuild-example": "npm install",
     "build-example": "webpack --config webpack.dev.js",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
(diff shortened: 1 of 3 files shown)


examples/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "serve-example": "webpack serve --config webpack.dev.js",
+    "prebuild-example": "npm install",
```
toloudis, line 8: I forget, does this get run automatically from the top-level package.json, or do you have to explicitly do it?

meganrm: it gets called from the top level package.json by the GitHub action.

frasercl (approved):
love a pre-script script

toloudis (approved):

PR #369, opened 2024-03-08
onRecordedMovies should be optional

### LINKED ISSUE #368: onRecordedMovie should be optional
## Use Case

If the installing app is not going to use this feature it shouldn't be required 

## Acceptance Criteria

make it optional in the interface 

## Details

interim17:
Review Time
=======
_Small_

Problem
=======
Closes #368 
`onRecordedMovies` should be optional for implementations where movie recording is not desired.

_Note: snuck in a commit in the prettier config to prevent {" "} < ---- those from appearing, I think it works to prevent them showing up when not needed, can add to sim-website too_

Solution
========
Made the `recording` and `onRecordedMovie` props optional. If `onRecordedMovie` is undefined, set `recorder` to null, and adjusted references to account for that. 

When a callback is provided, the feature will be enabled, and if the prop is undefined then no `FrameRecorder` functions should be called even if `recording` is set to `true`.

Added a function in test-bed to check if recording is enabled or not and display a message if so.

To test you can comment out the prop assigment:
 `// onRecordedMovie={this.onRecordedMovie}` Recording buttons should be disabled but viewer should work.
 
 For a further test provide a true value to `isRecordingEnabled`
 ```
                 <RecordMovieComponent
                    isRecording={this.state.isRecording}
                    setIsRecording={this.setIsRecording}
                    isRecordingEnabled={true}
                />
 ```
 
 Buttons will be clickable but should have no effect.

### DIFF
--- a/examples/src/RecordMovieComponent.tsx
+++ b/examples/src/RecordMovieComponent.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from "react";
+import React, { useState, useEffect } from "react";
 
 interface RecordMovieComponentProps {
     isRecording: boolean;
@@ -10,7 +10,7 @@ const RecordMovieComponent = ({
     isRecording,
     setIsRecording,
     isRecordingEnabled,
-}: RecordMovieComponentProps) => {
+}: RecordMovieComponentProps): JSX.Element => {
     // recording time measured in seconds
     let [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
     let [outputStatus, setOutputStatus] = useState<string>("");
@@ -47,26 +47,21 @@ const RecordMovieComponent = ({
                 onClick={startRecording}
                 disabled={
                     !isRecordingEnabled || isRecording || !browserSupported
-                }
-            >
+                }>
                 Start Recording
             </button>
             <button
                 onClick={stopRecording}
                 disabled={
                     !isRecordingEnabled || !isRecording || !browserSupported
-                }
-            >
+                }>
                 Stop Recording
             </button>
             <div>{isRecording ? "Recording..." : ""}</div>
             <div>
-                {" "}
-                {!browserSupported
-                    ? "Browser does not support recording"
-                    : ""}{" "}
+                {!browserSupported ? "Browser does not support recording" : ""}
                 {!isRecordingEnabled
-                    ? "Recording is not enabled, provide a callback to enable feature"
+                    ? "Recording is not enled, provide a callback to enable feature"
                     : ""}
             </div>
             <div>
--- a/examples/src/RecordMovieComponent.tsx
+++ b/examples/src/RecordMovieComponent.tsx
@@ -3,11 +3,13 @@ import React, { useState, useEffect, useRef } from "react";
 interface RecordMovieComponentProps {
     isRecording: boolean;
     setIsRecording: (isRecording: boolean) => void;
+    isRecordingEnabled: boolean;
 }
 
 const RecordMovieComponent = ({
     isRecording,
     setIsRecording,
+    isRecordingEnabled,
 }: RecordMovieComponentProps) => {
     // recording time measured in seconds
     let [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
@@ -43,18 +45,30 @@ const RecordMovieComponent = ({
         <div>
             <button
                 onClick={startRecording}
-                disabled={isRecording || !browserSupported}
+                disabled={
+                    !isRecordingEnabled || isRecording || !browserSupported
+                }
             >
                 Start Recording
             </button>
             <button
                 onClick={stopRecording}
-                disabled={!isRecording || !browserSupported}
+                disabled={
+                    !isRecordingEnabled || !isRecording || !browserSupported
+                }
             >
                 Stop Recording
-                </button>
+            </button>
             <div>{isRecording ? "Recording..." : ""}</div>
-            <div> {!browserSupported ? "Browser does not support recording" : ""} </div>
+            <div>
+                {" "}
+                {!browserSupported
+                    ? "Browser does not support recording"
+                    : ""}{" "}
+                {!isRecordingEnabled
+                    ? "Recording is not enabled, provide a callback to enable feature"
+                    : ""}
+            </div>
             <div>
                 {isRecording
                     ? "Recording duration:  " +
--- a/examples/src/RecordMovieComponent.tsx
+++ b/examples/src/RecordMovieComponent.tsx
@@ -3,17 +3,15 @@ import React, { useState, useEffect } from "react";
 interface RecordMovieComponentProps {
     isRecording: boolean;
     setIsRecording: (isRecording: boolean) => void;
-    isRecordingEnabled: boolean;
 }
 
 const RecordMovieComponent = ({
     isRecording,
     setIsRecording,
-    isRecordingEnabled,
 }: RecordMovieComponentProps): JSX.Element => {
     // recording time measured in seconds
-    let [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
-    let [outputStatus, setOutputStatus] = useState<string>("");
+    const [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
+    const [outputStatus, setOutputStatus] = useState<string>("");
 
     const browserSupported = "VideoEncoder" in window;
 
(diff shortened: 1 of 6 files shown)


examples/src/RecordMovieComponent.tsx
```diff
@@ -43,18 +45,25 @@ const RecordMovieComponent = ({
...
+                }>
                 Start Recording
             </button>
             <button
                 onClick={stopRecording}
-                disabled={!isRecording || !browserSupported}
-            >
+                disabled={
+                    !isRecordingEnabled || !isRecording || !browserSupported
+                }>
```
meganrm, line 57: I know this is the testbed example, but having two triple ORs that only differ by one is confusing to read. I would suggest figuring out a way to make this clearer. if recording is not enabled I would think this whole component wouldn't render? Instead of just disabling the buttons

src/viewport/index.tsx
```diff
@@ -155,9 +155,12 @@ class Viewport extends React.Component<
...
         };
-        this.recorder = new FrameRecorder(() => {
-            return this.visGeometry.renderDom as HTMLCanvasElement;
-        }, this.props.onRecordedMovie);
+        if (this.props.onRecordedMovie === undefined) {
+            this.recorder = null;
+        } else {
+            this.recorder = new FrameRecorder(() => {
+                return this.visGeometry.renderDom as HTMLCanvasElement;
+            }, this.props.onRecordedMovie);}
```
meganrm, line 163: nit: this formatting looks weird

examples/src/Viewer.tsx
```diff
@@ -702,6 +702,16 @@ class Viewer extends React.Component<InputParams, ViewerState> {
         this.setState({ isRecording: isRecording });
     };
 
+    // Remove the Viewer's onRecordedMovie prop assigment to disable feature
+    private isRecordingEnabledInViewer = () => {
+        const viewerElement = this.viewerRef.current;
+        if (viewerElement) {
+            const viewerProps = viewerElement.props;
+            return !!viewerProps.onRecordedMovie;
```
meganrm, line 710: again, I know this is the test app, but it's to demo the feature. If I was installing this app, I wouldn't test my own prop to see if I was going to enable a feature. I would just have a boolean button that toggled  a feature. So here, to demo this I would make a "disable/enable recording" button that changes the app state. and when that changes I would change the `onReordedMovie` prop to null and wouldn't render the RecordMovieCoponent` at all, instead of adding a prop. that way it's mimicking the way we would use the viewer

interim17: Ok, that sounds good

examples/src/RecordMovieComponent.tsx
```diff
@@ -8,10 +8,10 @@ interface RecordMovieComponentProps {
 const RecordMovieComponent = ({
     isRecording,
     setIsRecording,
-}: RecordMovieComponentProps) => {
+}: RecordMovieComponentProps): JSX.Element => {
     // recording time measured in seconds
-    let [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
-    let [outputStatus, setOutputStatus] = useState<string>("");
+    const [recordingTimeElapsed, setRecordingTimeElapsed] = useState<number>(0);
+    const [outputStatus, setOutputStatus] = useState<string>("");
```
toloudis, line 14: I hope this codebase has a prefer-const rule somewhere in eslint or prettier

src/viewport/index.tsx
```diff
@@ -44,8 +44,8 @@ type ViewportProps = {
     showCameraControls: boolean;
     onError?: (error: FrontEndError) => void;
     lockedCamera?: boolean;
-    recording: boolean;
-    onRecordedMovie: (blob: Blob) => void;
+    recording?: boolean;
+    onRecordedMovie?: (blob: Blob) => void; // provide a callback to enable recording
```
toloudis, line 48: It might be helpful to have a comment near these two variables to describe how they are intended to interact.  Or maybe at the point where they are being passed in to the component.
The reason I ask is because it's easy to wonder: what if recording is true but onRecordedMovie is undefined? what's stopping that from happening?

toloudis (changes requested):

toloudis:
Oops my thing says "requested changes" but I meant to do a "general comments".

examples/src/RecordMovieComponent.tsx
```diff
@@ -43,18 +43,18 @@ const RecordMovieComponent = ({
         <div>
             <button
                 onClick={startRecording}
-                disabled={isRecording || !browserSupported}
-            >
+                disabled={isRecording || !browserSupported}>
```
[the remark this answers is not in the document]
interim17: Ok. I'll remove it! Megan asked that I figure out why these artifacts `{" "}` keep showing up so that's what I'm trying to prevent

examples/src/RecordMovieComponent.tsx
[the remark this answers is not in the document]
interim17: I'll look into it. The linting is behaving differently since moving the test bed into its own src folder.

src/viewport/index.tsx
interim17: Right good point. If the feature has not been enabled by providing a callback, then different values for `recording` will have no effect, it serves as an on/off switch for an enabled recorder.

interim17: I'll add comments, but if its preferred we could change it so that instead of receiving a boolean flag, we hold that value in `ViewportState` and expose `startRecording` and `stopRecording` methods that set that state...

interim17:
ok @toloudis this may constitute mission creep but I couldn't bring myself to write the comments on the `recording` flag because its just not how we handle this type of thing in the rest of the repo.

New changes:
- no more `recording` prop
- sim controller has `startRecording` and `stopRecording` noop methods for the front end to call
- viewport provides controller the functions it needs
- FrameRecorder watches the internal `isRecording` state more closely since we have removed the flags in the higher components
- RecordMoviesComponent handles `isRecording` state because Viewer.tsx doesn't need to know it anymore

src/controller/index.ts
```diff
@@ -63,6 +65,8 @@ export default class SimulariumController {
         this.visData = new VisData();
         this.tickIntervalLength = 0; // Will be overwritten when a trajectory is loaded
         this.postConnect = () => noop;
+        this.startRecording = () => noop;
+        this.stopRecording = () => noop;
```
interim17, line 70: sim controller methods are how we usually trigger viewer related actions in the front end

src/viewport/index.tsx
```diff
@@ -265,6 +267,8 @@ class Viewport extends React.Component<
                 simulariumController.initializeTrajectoryFile();
             }
         };
+        simulariumController.startRecording = this.startRecording.bind(this);
+        simulariumController.stopRecording = this.stopRecording.bind(this);
```
interim17, line 271: bind the viewport methods (which have the right context) to the controller methods for the front end to call

src/viewport/index.tsx
```diff
@@ -667,6 +664,20 @@ class Viewport extends React.Component<
...
+        this.recorder.start();
+    }
+
+    public stopRecording(): void {
+        if (!this.recorder) {
+            return;
+        }
+        this.recorder.stop();
+    }
+
```
interim17, line 680: Handle recording with these instead of passing down a boolean from the front end

src/simularium/FrameRecorder.ts
```diff
@@ -88,18 +91,21 @@ export class FrameRecorder {
...
-        } catch (e) {
-            console.log("setup failed", e);
-            return;
-        }
+        if (!this.isRecording)
+            try {
+                await this.setup();
+                this.isRecording = true;
+            } catch (e) {
+                console.log("setup failed", e);
```
toloudis, line 99: I know it's not part of this PR but the console log message needs some love.  It doesn't say WHAT setup failed.  And it makes me wonder if we need to throw a front end error or something.

src/viewport/index.tsx
```diff
@@ -372,13 +376,6 @@ class Viewport extends React.Component<
...
             }
         }
-        if (this.props.recording !== prevProps.recording && this.recorder.supportedBrowser) {
-            if (this.props.recording) {
-                this.recorder.start();
-            } else {
-                this.recorder.stop();
-            }
-        }
     }
```
toloudis, line 379: Do you need to check for changes to the onRecordedMovie prop here?

meganrm:
looks good, just FYI, there seems to be a slightly different behavior for me on safari than before: I can still hit the record button on safari in the test app but the resulting file won't play.

PR #378, opened 2024-05-01
Feature/turn controls back on

meganrm:
Time Estimate or Size
=======
xsmall

Problem
=======
for the education module: I originally thought I would use a separate instance of the viewer when changing to 3d trajectories, but the setup I have is just switching files so I need to be able to turn the controls on and off 

Solution
========
added a toggle function that gets called when the prop is changed 

## Type of change

### DIFF
--- a/src/viewport/index.tsx
+++ b/src/viewport/index.tsx
@@ -310,6 +310,7 @@ class Viewport extends React.Component<
             showPaths,
             showBounds,
             selectionStateInfo,
+            lockedCamera,
         } = this.props;
 
         if (selectionStateInfo) {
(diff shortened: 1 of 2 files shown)


src/viewport/index.tsx
```diff
@@ -547,8 +551,8 @@ class Viewport extends React.Component<
             }
             if (!this.props.lockedCamera) {
                 this.visGeometry.setFollowObject(intersectedObject);
+                this.visGeometry.addPathForAgent(intersectedObject);
```
meganrm, line 554: this is just moved within the block to prevent a path from being render if the user click on an agent while the camera is disabled

ShrimpCryptid (approved):

interim17 (approved):

PR #380, opened 2024-05-07
Feature/change to converted file

interim17:
website branch `fix/conversion-filename-bug` points at these changes to receive file that results from autoconversion

Time Estimate or Size
=======
small-ish

Problem
=======
The handling of autoconverted files was built in the viewer with the example viewer in mind, but now that we're integrating with the website repo, we need to make some minor changes to the viewer. 

The way autoconverted files are currently handled in the viewer is that the viewer requests a file to be converted and immediately requests frame 0. However, in the website, we want to clear any previous trajectory in between these two steps, allowing users to resume interacting with the previous simulation if they cancel the autoconversion request before octopus has replied with the converted data.

Solution
========
- change SimulariumController's `convertAndLoadTrajectory()` to `convertTrajectory()`, and have that only handle sending the conversion request. The clearing of any previous trajectory and loading of frame zero can be done after, using `changeFile()`
- update `changeFile()` to include the option to not create a new remote simulator instance. Overall, if `this.simulator` is a remote simulator and the connectionParams indicate that the newly requested file also uses a remote simulator, we don't need to create a fresh remote simulator instance, we can keep using the existing one, but we don't. For non-autoconversion simulations, creating a fresh instance doesn't cause any issues, so it doesn't matter. For autoconversion, we will have already created a remote simulator instance for requesting the conversion, and we don't want to miss any responses from that by creating a new instance with a new websocket connection, so we need to skip creating a simulator. Right now, I'm only skipping this step for this autoconversion case, although we could expand and use this across remote simulator cases in the future
- allow the website to pass in it's own file name to be used internally for the autoconverted file, since we're generating that in the viewer anyways
- update the example viewer autoconversion form to use these new / updated methods
- point to the new octopus ecs staging instance, instead of the trial ec2 instance's IP address

with @ascibisz 

## Type of change

### DIFF
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -394,7 +394,7 @@ class Viewer extends React.Component<InputParams, ViewerState> {
     public convertFile(obj: Record<string, any>, fileType: TrajectoryType) {
         const fileName = uuidv4() + ".simularium";
         simulariumController
-            .convertAndLoadTrajectory(this.netConnectionSettings, obj, fileType, fileName)
+            .convertTrajectory(this.netConnectionSettings, obj, fileType, fileName)
             .then(() => {
                 this.clearPendingFile();
             })
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { isEqual, findIndex, map, reduce } from "lodash";
+import { v4 as uuidv4 } from "uuid";
 
 import type {
     ISimulariumFile,
@@ -391,10 +392,18 @@ class Viewer extends React.Component<InputParams, ViewerState> {
     }
 
     public convertFile(obj: Record<string, any>, fileType: TrajectoryType) {
+        const fileName = uuidv4() + ".simularium";
         simulariumController
-            .convertAndLoadTrajectory(this.netConnectionSettings, obj, fileType)
+            .convertAndLoadTrajectory(this.netConnectionSettings, obj, fileType, fileName)
             .then(() => {
                 this.clearPendingFile();
+            })
+            .then(() => {
+                simulariumController.changeFile(
+                    { netConnectionSettings: this.netConnectionSettings, },
+                    fileName,
+                    true,
+                )
             })
             .catch((err) => {
                 console.error(err);
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -197,8 +197,8 @@ class Viewer extends React.Component<InputParams, ViewerState> {
             };
         } else if (props.useOctopus) {
             this.netConnectionSettings = {
-                serverIp: "18.223.108.15",
-                serverPort: 8765,
+                serverIp: "staging-simularium-ecs.allencell.org",
+                serverPort: 443,
                 useOctopus: true,
                 secureConnection: true,
             };
--- a/src/controller/index.ts
+++ b/src/controller/index.ts
@@ -351,7 +351,8 @@ export default class SimulariumController {
     public changeFile(
         connectionParams: SimulatorConnectionParams,
         // TODO: push newFileName into connectionParams
-        newFileName: string
+        newFileName: string,
+        keepRemoteConnection: boolean = false
     ): Promise<FileReturn> {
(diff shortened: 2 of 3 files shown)


src/controller/index.ts
```diff
@@ -352,7 +350,8 @@ export default class SimulariumController {
     public changeFile(
         connectionParams: SimulatorConnectionParams,
         // TODO: push newFileName into connectionParams
-        newFileName: string
+        newFileName: string,
+        keepRemoteConnection: boolean = false
```
ShrimpCryptid, line 354: Linting had a comment here to remove boolean
```suggestion
        keepRemoteConnection = false
```

examples/src/Viewer.tsx
```diff
@@ -391,11 +392,19 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+            .convertTrajectory(this.netConnectionSettings, obj, fileType, fileName)
             .then(() => {
                 this.clearPendingFile();
             })
+            .then(() => {
+                simulariumController.changeFile(
+                    { netConnectionSettings: this.netConnectionSettings, },
+                    fileName,
+                    true,
+                )
```
ShrimpCryptid, line 406: Understanding question, does this switch the viewer over to the new file?

src/controller/index.ts
```diff
@@ -250,13 +250,11 @@ export default class SimulariumController {
             return Promise.reject(e);
         }
 
-        return this.simulator
-            .convertTrajectory(dataToConvert, fileType)
-            .then(() => {
-                if (this.simulator) {
-                    this.simulator.requestSingleFrame(0);
-                }
-            });
```
ShrimpCryptid, line 259: Is this no longer needed because of the `simulariumController.changeFile` call in `Viewer.tsx`?

ascibisz: Correct!

examples/src/Viewer.tsx
ascibisz: yes, this call does the bulk of the work of clearing the previous file, and requests the first frame for the new file

src/controller/index.ts
ascibisz: I'm learning new things about typescript every day! it makes sense that you don't have to explicitly define the type in this situation

- [ ] lint fix

ShrimpCryptid (approved):
Thanks for the clarification! LGTM!

examples/src/Viewer.tsx
toloudis: Just checking my understanding.  Is it possible to convertFile, and then load a different file, and then when conversion is done, suddenly the viewer will switch over to the newly converted file?

src/controller/index.ts
```diff
@@ -250,13 +250,11 @@ export default class SimulariumController {
...
+        return this.simulator.convertTrajectory(
+            dataToConvert,
+            fileType,
+            providedFileName
+        );
```
toloudis, line 257: i like this change. simpler, better decoupling

src/controller/index.ts
```diff
@@ -352,7 +350,8 @@ export default class SimulariumController {
     public changeFile(
         connectionParams: SimulatorConnectionParams,
         // TODO: push newFileName into connectionParams
-        newFileName: string
+        newFileName: string,
+        keepRemoteConnection = false
```
toloudis, line 354: Did you consider making this a member variable that's more persistent, like a mode or state of the controller, as opposed to passing it in all the time?

toloudis (approved):

src/controller/index.ts
toloudis: I just re-read the PR text and it makes sense to me how you did it. If you imagine some kind of loadbalancer in place then it might make sense to always have to reconnect.

src/controller/index.ts
interim17: re: switching files, I think so. Alli is looking into this an its one of the questions she is working on in regards to when we open and close new websockiet connections>

It's possible that if you open a new websocket connection when you open the different file, that the converted file would hit a websocket connection that is no longer active in the viewer and not load or cause an error. Typically the conversion happens so fast, and the menu to load other files is disabled during, so it's not something we've tested directly

PR #381, opened 2024-05-08
Feature/trim cache head

### LINKED ISSUE #372: clear out cache of local simulation after a certain size
## Use Case

when running a simulation on the front end, you can just keep adding frames infinitely 

## Acceptance Criteria

The cache should have an upper size limit, 
possibly a setting that lets you define how much you want to keep (number of frames? or a size?)

## Details

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
This PR no longer addresses the ability to provide a desired max size or length for the cache array, but rather establishes a boolean flag for whether to implement caching at all.

An app like binding sim will be able to pass in `false` for that prop and the cache will be limited to the single current frame.

Caching according to size is in development.

Solution
========
#### Viewport
- Take an optional boolean flag in the viewer props `implementCache` and propogate to `visData`

#### VisData:
- add public method `setCachePreferences`, to toggle `implementCache`
- add private method `trimCacheHead` to trim cache down to single frame


blairlyons:
1001 frames is not a limit, just tends to be a good number for file size and playback time for many of our examples

src/simularium/VisData.ts
```diff
@@ -389,6 +399,17 @@ class VisData {
         if (this.firstFrameTime === null) {
             this.firstFrameTime = frames.frameDataArray[0].time;
         }
+
+        if (this.frameDataCache.length > this.maxCacheLength) {
```
toloudis, line 403: Food for thought: how would you do this if you needed to limit the cache based on bytes?  For example "we want to cache up to 2GB of frames but no more".  The reason I ask is because 2000 frames of one trajectory could be wildly different in memory consumed than 2000 frames of a different trajectory.

interim17: @meganrm just made this point in our 1:1 as well

I think that is definitely the next step, and then the length property might be totally discarded or only used for the case of user provided limit, or a minimal/no caching scenario.

We could table this PR on behalf of not using length at all and re-open with a memory based limit, but this felt like the easiest way to introduce the ability to cap the cache for use in `binding-sim`.

examples/src/Viewer.tsx
```diff
@@ -1021,6 +1021,7 @@ class Viewer extends React.Component<InputParams, ViewerState> {
                         onError={this.onError}
                         backgroundColor={[0, 0, 0]}
                         lockedCamera={false}
+                        maxCacheLength={2001}
```
ShrimpCryptid, line 1024: Should this be set to `MAX_CACHE_LENGTH + 1`?

ShrimpCryptid (approved):
LGTM. This seems good as an initial step, though based on your discussion with Dan and Megan it seems like this is likely to be overhauled in the near future?

meganrm:
> LGTM. This seems good as an initial step, though based on your discussion with Dan and Megan it seems like this is likely to be overhauled in the near future? I'll defer to Dan on that one!

Yeah, I'm in favor of no default length, because as Dan says, it's really hard to predict was a real upper limit is. So I think max cache length should only be a setting provided by props, and otherwise not used. And then in the next PR include a max size as we discussed, that is based on browser memory limits (and adjustable by a user prop )

interim17:
@meganrm @toloudis 
I'll adjust this so that there is not default behavior driven by cache length.

If we want to be diligent about knowing what the cache size _actually_ is, then I think we probably need to use array buffer as a data structure when we store the cache? I thought of a couple hacky options, like making a Blob out of the first frame, or every n frames, and then multiplying the blob size by length, but that's not very precise.

toloudis:
addFramesToCache uses a ParsedBundle.  Basically it's caching the AgentData array.  You could write a very simple function that gets the size of an AgentData (which is a fixed size plus variable number of subpoints), and run through the list and add up all the sizes (how many bytes is a `number` in js? 8?  Which brings up the fact that it's probably double what is in the ArrayBuffer representation that came across the network).  You could compute the size before the data is moved into the ParsedBundle and the ParsedBundle could carry an extra member variable like `cachedSize`.

interim17:
> addFramesToCache uses a ParsedBundle. Basically it's caching the AgentData array. You could write a very simple function that gets the size of an AgentData (which is a fixed size plus variable number of subpoints), and run through the list and add up all the sizes (how many bytes is a `number` in js? 8? Which brings up the fact that it's probably double what is in the ArrayBuffer representation that came across the network). You could compute the size before the data is moved into the ParsedBundle and the ParsedBundle could carry an extra member variable like `cachedSize`.

this draft pr https://github.com/simularium/simularium-viewer/pull/384 has a basic implementation. but I don't like storing the extra object `cacheFrameSizes` it definitely needs work

we could take the bundle size, divide by number of frames, and add cachedSize to FrameData?

This is also not performant, when it kicks in during playback it really slows things down to be trimming the cache and then adding to the cache at the same time.

interim17:
So in discussion w @toloudis this week we are making progress on caching according to size, but I don't want to rush it.

Really for binding sim the feature we want is: the ability to not implement caching at all, the user provided size is not critical at this point.

So I think it makes more sense to just have a boolean prop `implementCache`, and when it is false, limit the cache to the single current frame. I tested this in binding-sim app this morning and didn't find any issues.

Later when we implement cache size limits, there can be both the boolean flag, and if true, an optional user provided size, as well as default sizes etc. These would be `cachePreferences` or something like that.

For now `trimCacheHead` is hard coded to take cache down to 1 frame as its only called when `implementCache` is false, but we can make it more flexible when we add other options to cache preferences.

examples/src/Viewer.tsx
```diff
@@ -1035,6 +1035,7 @@ class Viewer extends React.Component<InputParams, ViewerState> {
                         onError={this.onError}
                         backgroundColor={[0, 0, 0]}
                         lockedCamera={false}
+                        disableCache={false}
```
ShrimpCryptid, line 1038: Nit: Slightly prefer inverse of this, like `enableCache` or `cacheEnabled`? It's a little easier to reason about affirmative names!

src/simularium/VisData.ts
```diff
@@ -376,15 +381,29 @@ class VisData {
         }
     }
 
+    public setCachePreferences(disableCache: boolean): void {
+        this.disableCache = disableCache;
+    }
+
```
ShrimpCryptid, line 387: See above note about renaming `disableCache`! I think here I wouldn't know what `setCachePreferences(true)` or `setCachePreferences(false)` mean if I saw it elsewhere in the codebase. You could either change this to `setCacheEnabled()`, or accept an object (`{ cacheEnabled: boolean }`) to make this more clear.

examples/src/Viewer.tsx
interim17: disable > enable was by @meganrm request from meeting yesterday!

src/simularium/VisData.ts
interim17: Ok I can change this! I was anticipating that in future this function might also take in max sizes, but I definitely see what you mean here

examples/src/Viewer.tsx
ShrimpCryptid: Oh, interesting... @meganrm I've generally seen affirmative names be convention for booleans, could you give some context for why you wanted `disableCache` here? :0

meganrm: because it's an optional prop you'd only use to disable the cache. The default behavior is to have a cache and therefore not have the prop. I get what you're saying about readability, but this props is basically going to not be there at all or be `disableCache={true}` (or you can do shorthand to just have `disableCache` which is clearer to me that always have `enableCache={false}` as the only way the prop is really used

meganrm: also, how this prop is then handled in the internals can be different: ie, whatever is more readable there. To me this is similar to like `disabled` as a prop on a button.

src/simularium/VisData.ts
```diff
@@ -376,15 +381,29 @@ class VisData {
...
         Array.prototype.push.apply(
             this.frameCache,
             frames.parsedAgentDataArray
         );
     }
 
+    private trimCacheHead(): void {
+        while (this.frameCache.length > 1) {
+            this.frameCache.shift();
+            this.frameDataCache.shift();
```
meganrm, line 403: is it possible to just not store the cache instead of storing it and then deleting it?

src/simularium/VisData.ts
```diff
@@ -376,15 +381,29 @@ class VisData {
...
     }
 
+    public setCacheDisabled(cacheDisabled: boolean): void {
+        this.disableCache = cacheDisabled;
+    }
+
     // Add parsed frames to the cache and save the timestamp of the first frame
     private addFramesToCache(frames: ParsedBundle): void {
+        if (this.disableCache) {
+            this.trimCacheHead();
```
meganrm, line 391: it seems like you could just return here?

src/viewport/index.tsx
```diff
@@ -123,7 +126,11 @@ class Viewport extends React.Component<
         this.handleTimeChange = this.handleTimeChange.bind(this);
 
         this.visGeometry = new VisGeometry(loggerLevel);
+        this.disableCache = props.disableCache;
```
meganrm, line 129: I don't think you need to save this on the class here? you can just use the prop below

src/simularium/VisData.ts
meganrm: I tried my suggestions and it doesn't render anything: I guess I don't fully understand how this cache is currently working

src/simularium/VisData.ts
interim17: So the cache is pretty tightly integrated into the app/animate cycle and needs to have at least the current frame.

src/viewport/index.tsx
```diff
@@ -57,6 +58,7 @@ const defaultProps = {
     showPaths: true,
     showBounds: true,
     lockedCamera: false,
+    disableCache: true,
```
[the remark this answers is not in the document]
interim17: ah you're right!

src/simularium/VisData.ts
```diff
@@ -287,6 +292,14 @@ class VisData {
...
      * */
     public hasLocalCacheForTime(time: number): boolean {
+        // this check seems to have some issues with the time comparison
+        // it returns true when it shouldn't
+        // TODO: debug compareTimes
+        // in the meantime, if we're not using the cache we know it's not
+        // in there
+        if (this.disableCache) {
+            return false;
+        }
```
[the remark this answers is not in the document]
interim17: That's funny, I was also able to make it work by reducing the possible cache size by one more frame, your fix is better, but we're going to need `compareTimes` to work properly for later cache work.

src/viewport/index.tsx
interim17: yes, change made

src/simularium/VisData.ts
```diff
@@ -23,6 +23,7 @@ const EOF_PHRASE: Uint8Array = new TextEncoder().encode(
 class VisData {
     private frameCache: AgentData[][];
     private frameDataCache: FrameData[];
+    private disableCache: boolean;
```
toloudis, line 26: as others have said, this should be the affirmative version of the variable, like `useCache` here.

src/simularium/VisData.ts
```diff
@@ -376,15 +389,29 @@ class VisData {
         }
     }
 
+    public setCacheDisabled(cacheDisabled: boolean): void {
```
toloudis, line 392: prefer setCacheEnabled

src/simularium/VisData.ts
toloudis: if "disabled" means only cache one frame, then maybe all you need to do is:
```
this.frameDataCache = [frames.frameDataArray];
this.frameCache = [frames.parsedAgentDataArray];
```
instead of trimCacheHead.

toloudis (commented):
I wonder if this "only cache the current frame" makes sense for all localSimulators!

src/simularium/VisData.ts
interim17: Ok, I think I'll change it back then, since there seems to be a majority preferring the affirmative.

interim17:
@toloudis @meganrm just so everyone is on the same page: changed the data management bits to `enableCache` but left the front end prop as `disableCache` per updated feedback from megan, at this point I'm agnostic :)

interim17:
@toloudis your suggestion above works with a spread operator so I removed `trimCacheHead`

src/simularium/VisData.ts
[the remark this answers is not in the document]
toloudis: I think the latest changes imply that there are always cached frames (if enableCache=false then it's only the current frame), so at least the comment above should change?

toloudis (approved):
one minor thing about an outdated comment but this looks ok!

ShrimpCryptid (approved):

src/simularium/VisData.ts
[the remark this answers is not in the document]
meganrm: In response to "compareTimes is doing the right thing?": it's not. The education app never requests times in the past, but something with this calculation (I think the epsilon value) sometimes leads it to return true even though the newly requested time is never cached

PR #382, opened 2024-05-09
Fix/no firstframetime

### LINKED ISSUE #362: Start time can't be larger than zero
## Description
If a trajectory starts at a time value larger than zero, time playback controls other than 'play' don't work.

## Expected Behavior
Time controls should work no matter what time the simulation starts at

## Reproduction
A version of the same trajectory when [time starts at zero](https://staging.simularium.allencell.org/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/time_starts_at_zero.simularium) and when [time starts larger than zero](https://staging.simularium.allencell.org/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/time_starts_after_zero.simularium). 

1. load the trajectory with time larger than zero
2. click or drag the time bar to skip to a time with or without hitting play first
3. the time bar is frozen with spinners on all the buttons

## Environment
prod or staging, Firefox

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
Closes #362 

Solution
========
Recent changes to octopus and this @ascibisz [PR from yesterday](https://github.com/simularium/octopus/pull/82) seem to have mostly resolved these bugs. We don't need to subtract firstFrameTime anymore, when calling `goToRemoteSimulationTime` and so `visData.firstFrameTime` isn't being used anywhere. I just removed all reference to it. The front end gets the correct `firstFrameTime` by reading the `timeData` when the trajectory lands, we dont need to store it in the viewer.

Prior to realizing Alli had already fixed this I wrote code that I think improves on how we were storing `firstFrameTime` if we want it later, but for now I think we can just remove it

Start time of 0 was hard coded into the test bed slider. I changed it to act more like the website and get the value from `timeData` on initial play.

Steps to Verify:
----------------
1. Install viewer in website
1. Try all the playback controls for the trajectories linked in the issue without hitting play first and see if you get hung up on anything and if first frame time displays correctly.

### DIFF
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -413,9 +413,6 @@ class Viewer extends React.Component<InputParams, ViewerState> {
         const simulariumFile = fileName.includes(".simularium")
             ? trajectoryFile
             : null;
-        // if (!fileName.includes(".simularium")) {
-        //     return new
-        // }
         this.setState({ initialPlay: true})
         return simulariumController
             .handleFileChange(simulariumFile, fileName, geoAssets)
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -107,6 +107,8 @@ interface ViewerState {
     serverHealthy: boolean;
     isRecordingEnabled: boolean;
     trajectoryTitle: string;
+    initialPlay: boolean;
+    firstFrameTime: number;
 }
 
 interface BaseType {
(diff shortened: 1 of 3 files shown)


examples/src/Viewer.tsx
```diff
@@ -412,6 +416,7 @@ class Viewer extends React.Component<InputParams, ViewerState> {
         // if (!fileName.includes(".simularium")) {
         //     return new
         // }
```
meganrm, line 418: I know it's not part of this code, but could this commented out code be removed ?

meganrm (approved):

examples/src/Viewer.tsx
interim17: sure :)

frasercl (approved):

PR #386, opened 2024-05-20
Fix/plotdata

meganrm:
Time Estimate or Size
=======
 xsmall

Problem
=======
`getPlotData` is typed to return `Plot[]` but when I logged what I got, it returns an object with `{version: 1, data: Plot[]}`, so I changed the functions to actually return Plot[]

Solution
========
Adjust the return of getPlotData to unpack the data

## Type of change


toloudis (commented):
What about all the plots in our example trajectories that are displayed in the simularium-website? Do they go through a different code path than this?

meganrm:
> What about all the plots in our example trajectories that are displayed in the simularium-website? Do they go through a different code path than this? (would this change break them or are they already broken and this is fixing it?)

this is a breaking change, but the fix is really small: remove a `.data` when unpacking the data. I could have changed the typing but my thinking here is that the version is really for the internals of the viewer, and anyone consuming the plot data shouldn't be managing/caring about the version number

toloudis:
> this is a breaking change, but the fix is really small: remove a `.data` when unpacking the data. I could have changed the typing but my thinking here is that the version is really for the internals of the viewer, and anyone consuming the plot data shouldn't be managing/caring about the version number

So the fix goes in simularium-website, or where? Is there already another PR that fixes the breakage?

meganrm:
> > this is a breaking change, but the fix is really small: remove a `.data` when unpacking the data. I could have changed the typing but my thinking here is that the version is really for the internals of the viewer, and anyone consuming the plot data shouldn't be managing/caring about the version number
> 
> So the fix goes in simularium-website, or where? Is there already another PR that fixes the breakage?

it's here https://github.com/simularium/simularium-website/pull/522

ShrimpCryptid:
(Left an approval since I saw that this was urgent, but will defer to Dan since there was an existing discussion)

PR #390, opened 2024-05-31
With octopus, we can assume frame per message

ascibisz:
Time Estimate or Size
=======
Small

Problem
=======
Part of cleanup work for transition away from simularium-engine, included in this ticket
[Link to ticket](https://github.com/simularium/simularium-viewer/issues/389)

Solution
========
With simularium engine, frame data messages arrived with as much frame data as possible included, up to a certain size. We decided this didn't really optimize performance enough to justify some of the bugs it caused, so octopus only sends one frame at a time. Now, we can handle frame messages from remote simulations nearly the same as we do for local file simulations.

One weird issue that came up is that local file simulation's binary frame data keeps the `nAgents` param as a binary encoded **int**, which is what happens in simularium file format. However, octopus / simularium-engine encode `nAgents` as a binary encoded **float**.  This became obvious when I started combining the local frame data and remote frame data decoding logic. It feels silly to case out how the `nAgents` value is decoded based on whether it's a local or remote file simulation, so I'm changing how nAgents is encoded in octopus to be an int instead of a float. The PR for that small change is [here](https://github.com/simularium/octopus/pull/97). I'll have to be intentional about how these PRs get merged, because if one is pushed into staging or production before the other, that environment will break, but since deploying octopus is so easy I'm not overly concerned.

### DIFF
--- a/src/controller/index.ts
+++ b/src/controller/index.ts
@@ -140,7 +140,7 @@ export default class SimulariumController {
                 this.visGeometry.geometryStore.cacheLocalAssets(geoAssets);
             }
             this.simulator.setTrajectoryDataHandler(
-                this.visData.parseAgentsFromLocalFileData.bind(this.visData)
+                this.visData.parseAgentsFromFrameData.bind(this.visData)
             );
         } else if (netConnectionConfig) {
             const webSocketClient = new WebsocketClient(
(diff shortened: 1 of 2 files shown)


interim17 (commented):
This looks great! What a nice amount of code removed. Just to confirm, this will be broken if we try to run this with production Octopus currently?

meganrm (approved):

interim17 (approved):

PR #395, opened 2024-06-10
spread null agent object when parsing frames

interim17:
Introduced a bug when I [merged this PR.](https://github.com/simularium/simularium-viewer/pull/392)

Not all agents rendering in viewport on load.

Source of bug: I created `NULL_AGENT` constant to stay DRY but ended up with pass-by-reference issues

Fix: using spread operator seems to be a quick fix.

Not sure how I missed this initially or if something changed in viewer or octopus, it seemed to be working well before, sorry.


frasercl (approved):
Referenceeeeess!

toloudis (approved):
An other way to do this is to make NULL_AGENT a function that returns a new object every time, like a default constructor of sorts.

interim17:
> Referenceeeeess!

（＞_＜）

PR #396, opened 2024-06-13
Fix/null agent function

interim17:
avoid pass by reference issues by getting null agents with a function instead of a constant

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -34,7 +34,7 @@ export const enum TrajectoryType {
 
 export const DEFAULT_FRAME_RATE = 60; // frames per second
 
-export const createNullAgent = (): AgentData => {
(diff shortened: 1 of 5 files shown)


src/constants.ts
```diff
@@ -34,16 +34,18 @@ export const enum TrajectoryType {
...
-    y: 0,
-    z: 0,
-    xrot: 0,
-    yrot: 0,
-    zrot: 0,
-    cr: 0,
-    subpoints: [],
+export const nullAgent = (): AgentData => {
+    return {
+        "vis-type": -1,
```
toloudis, line 39: should use `visType`?

ShrimpCryptid (approved):

toloudis (approved):
Looks good!

PR #401, opened 2024-08-12
export AgentData type properly from /types and / simularium

interim17:
Time Estimate or Size
=======
_tiny_

Problem/Solution
=======
`AgentData` isn't currently being exported as a type, which creates an error, this gets it exported as a type and in the same way as the other types.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ export type {
     EncodedTypeMapping,
     VisDataMessage,
     Plot,
+    AgentData,
 } from "./simularium";
 export type { ISimulariumFile } from "./simularium/ISimulariumFile";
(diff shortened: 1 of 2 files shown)


tyler-foster:
[I was _literally just about to make a PR to fix this_.](https://github.com/simularium/simularium-viewer/compare/bugfix/agentdata_type_export?expand=1)

tyler-foster (approved):

interim17:
> [I was _literally just about to make a PR to fix this_.](https://github.com/simularium/simularium-viewer/compare/bugfix/agentdata_type_export?expand=1)

Ha! I've had a fix for a while but foolishly buried it in a different PR.

meganrm (approved):

PR #411, opened 2024-09-11
Caching: refactor, memory improvements, trimming

### LINKED ISSUE #372: clear out cache of local simulation after a certain size

interim17:
Time Estimate or Size
=======
_large: many lines that touch many important parts of the viewer, but some of the large line count is tests!_

Problem
=======
Closes #372 and also refactors to address memory and performance concerns, and lay the groundwork for future updates to caching.

paired with @toloudis 

Solution
========
Initial attempts to trim the cache during playback encountered performance and memory bottlenecks that dramatically slowed down the viewer. These seem to be primarily related to triggering an excess of garbage collection events. Poking around on memory profiling revealed several places in the code that data was being duplicated, particularly copies of `AgentData`. The result of these explorations is a refactored cache that has a number of other changes accompanying it, both to provide the new functionality, and to address these memory and performance concerns.

### Linked list cache
The cache is now it's own class in `VisDataCache.ts`. It takes an optional settings prop:
```
interface VisDataCacheSettings {
    maxSize: number;
    cacheEnabled: boolean;
}
```
and provides a public method, `changeSettings` to update those properties after initializing.

Like before, if the cache is disabled, only the current frame is ever stored.

When the cache is enabled and a `maxSize` has been provided, the cache will be trimmed before adding any new frame to keep its size under the limit. The trimming behavior could probably be made more sophisticated in the future.

Refactoring the cache to be a linked list of `CachedFrames` was effective in reducing the garbage collection events that seemed to be slowing things down. Maybe a matter of opinion but I also find that refactoring it into its own file and class makes working with cache methods easier than the previous two arrays of data and metadata. 

For the future: this cache format can be refactored to facilitate a multi-part cache with non-continuous sections of cached frames, which could reduce load on the back end by preserving local cached data whenever possible.

### Move parsing of `AgentData` to `updateScene`
We have moved the parsing of `AgentData` from the functions that handle incoming binary and JSON data to the call of `updateScene` in `VisGeometry`.
This makes `updateScene` more expensive, but does not increase the number of times data needs to be parsed.
By caching the raw incoming data largely unparsed (except for critical metadata like time, frameNumber, agentCount) we reduce the memory footprint of the cached data, and the processing needed to cache the frames.
### Object pooling of `VisAgentInstances`
Some trajectories efficiently reuse agent Ids and others do not. In cases where the number of unique Ids climbs throughout playback, the `VisAgentInstances` would get larger and larger, holding excess copies of stale `AgentData`
We now only store current agent instances. The unused `VisAgent` objects released by old agents are pooled in an array and re-used, rather than instantiating new ones every time. In cases where it matter this has a dramatic effect on memory use.

## Type of change

Steps to Verify:
----------------
1. In the test bed load and run binary trajectories with different settings (cache enabled/not, cache enabled with and without maxSize provided), confirm that everything runs smoothly with unlimited cache and with trimming (tip: adding a log in `trimCache` to confirm that you have hit the limit and begun trimming makes this easier)
2. Locally load A JSON trajectory and confirm they it loads properly correctly.
3. Installing a dev build into the website is also a nice way to confirm things aren't busted.


src/simularium/VisData.ts
```diff
@@ -1,95 +1,48 @@
...
-            }
-            parsedAgentData.push(agentData);
-        }
-        parsedAgentDataArray.push(parsedAgentData);
-
-        return {
-            parsedAgentDataArray,
-            frameDataArray,
-        };
+        return frameData;
```
interim17, line 36: This function has a much reduced role, simply constructs the cachedFrame object including the raw binary in the data field.

src/simularium/VisData.ts
```diff
@@ -98,91 +51,62 @@ class VisData {
...
-                return this.frameDataCache[0];
-            } else if (this.cacheFrame >= this.frameDataCache.length) {
-                return this.frameDataCache[this.frameDataCache.length - 1];
-            } else {
-                return this.frameDataCache[this.cacheFrame];
-            }
-        }
+    public setOnErrorCallback(callback: (error: FrontEndError) => void): void {
+        this.onError = callback;
+    }
```
interim17, line 65: We are now passing down the `on.Error` prop to VisData and we do in `VisGeometry`, mostly to catch the edge case where an incoming frame is larger than the maxSize of the cache.

I'm curious if I should do the same and pass this error handling all the way down into `VisDataCache` as well to catch frame access errors, or alternately if there is a different approach to be using here.

src/simularium/VisData.ts
```diff
@@ -259,32 +179,38 @@ class VisData {
...
+            parsedMsg.size > this.frameCache.maxSize
+        ) {
+            this.onError(
+                new FrontEndError(
+                    `Frame size exceeds cache size: ${parsedMsg.size} > ${this.frameCache.maxSize}`,
+                    ErrorLevel.ERROR
+                )
+            );
+            return;
+        }
```
interim17, line 194: Providing error handling callback to `VisData` for basically just this scenario, so that the user sees a pleasant warning modal and not an uncaught runtime error if their cache is too small for their data.

src/simularium/VisData.ts
```diff
@@ -1,95 +1,47 @@
...
-import type { ParsedBundle } from "./VisDataParse";
+import { VisDataMessage, CachedFrame } from "./types";
 import { parseVisDataMessage } from "./VisDataParse";
-import { nullAgent } from "../constants";
+import { VisDataCache } from "./VisDataCache";
+import { ErrorLevel, FrontEndError } from "./FrontEndError";
 
 class VisData {
-    private frameCache: AgentData[][];
-    private frameDataCache: FrameData[];
```
interim17, line 17: These two arrays are the old cache, now replace by `VisDataCache`

src/simularium/VisData.ts
```diff
@@ -1,95 +1,47 @@
...
         const intView = new Uint32Array(data);
-        const parsedFrameData = {
-            time: floatView[1],
+
+        const frameData: CachedFrame = {
+            data: data,
             frameNumber: floatView[0],
+            time: floatView[1],
+            agentCount: intView[2],
+            size: data.byteLength + 8 * 4,
```
interim17, line 32: This function has a much reduced role, simply constructs the cachedFrame object including the raw binary in the data field.

src/simularium/VisData.ts
```diff
@@ -259,32 +179,38 @@ class VisData {
...
+        const parsedMsg: CachedFrame = parseVisDataMessage(visDataMsg);
+        if (
+            this.frameCache.cacheSizeLimited &&
+            parsedMsg.size > this.frameCache.maxSize
+        ) {
+            this.onError(
+                new FrontEndError(
+                    `Frame size exceeds cache size: ${parsedMsg.size} > ${this.frameCache.maxSize}`,
+                    ErrorLevel.ERROR
+                )
```
interim17, line 191: Providing error handling callback to VisData for basically just these scenarios (one for binary, one for JSON), so that the user sees a pleasant warning modal and not an uncaught runtime error if their cache is too small for their data.

src/visGeometry/index.ts
```diff
@@ -1542,51 +1553,49 @@ class VisGeometry {
...
+            if (path && visAgent) {
+                lastx = visAgent.agentData.x;
+                lasty = visAgent.agentData.y;
+                lastz = visAgent.agentData.z;
             }
 
             if (!visAgent) {
-                visAgent = this.createAgent();
+                if (this.availableAgentPool.length > 0) {
+                    visAgent = this.availableAgentPool.pop() as VisAgent;
```
toloudis, line 1593: any concern about leftover data in this visAgent, or does the following code completely overwrite its internals?

src/test/util.test.ts
```diff
@@ -69,4 +74,146 @@ describe("util", () => {
             expect(result).toEqual(`/${path}`);
         });
     });
+
+    describe("getNextAgentOffest", () => {
```
toloudis, line 78: do you have a test for what happens if you pass a bad offset?

src/visGeometry/index.ts
```diff
@@ -119,6 +126,7 @@ class VisGeometry {
     public followObjectId: number;
     public visAgents: VisAgent[];
     public visAgentInstances: Map<number, VisAgent>;
+    private availableAgentPool: VisAgent[] = [];
```
toloudis, line 129: this availableAgentPool is awesome

src/simularium/VisData.ts
interim17: At this point the default is unlimited.

There was discussion of setting default limits based on different browsers, but I deferred thinking that through. This will only be reached if a default cache size has been manually set.

src/controller/index.ts
```diff
@@ -66,6 +66,9 @@ export default class SimulariumController {
 
         this.handleTrajectoryInfo = (/*msg: TrajectoryFileInfo*/) => noop;
         this.onError = (/*errorMessage*/) => noop;
+        if (this.onError) {
+            this.visData.setOnError(this.onError);
+        }
```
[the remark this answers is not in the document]
interim17: No, I think this is a mistake, I tried a couple different things related to error handling and I left this is in by accident.

I'm still calling `visData.setOnError` in the viewport if the prop is provided.

src/visGeometry/workers/visDataWorker.ts
```diff
@@ -1,16 +1,7 @@
...
     (e: MessageEvent) => {
-        const visDataMsg = e.data;
-        const { frameDataArray, parsedAgentDataArray } =
-            parseVisDataMessage(visDataMsg);
-
-        postMessage({
-            frameDataArray,
-            parsedAgentDataArray,
-        });
+        postMessage(e.data);
```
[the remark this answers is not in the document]
interim17: Made a note to follow up on this after we talked about it!

src/simularium/VisData.ts
```diff
@@ -98,91 +50,58 @@ class VisData {
         if (util.ThreadUtil.browserSupportsWebWorkers()) {
             this.setupWebWorker();
         }
-        this.frameCache = [];
-        this.frameDataCache = [];
-        this.cacheFrame = -1;
-        this.enableCache = true;
+        this.currentFrameNumber = -1;
```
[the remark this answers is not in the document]
interim17: We should yes because `clearForNewTrajectory` calls `clearCache` which resets it to -1

src/simularium/VisData.ts
```diff
@@ -1,95 +1,47 @@
...
-    private frameDataCache: FrameData[];
-    private enableCache: boolean;
+    public frameCache: VisDataCache;
     private webWorker: Worker | null;
 
     private frameToWaitFor: number;
     private lockedForFrame: boolean;
-    private cacheFrame: number;
+
+    protected currentFrameNumber: number;
```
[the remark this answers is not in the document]
interim17: because it sounds cool? changing to private.

src/visGeometry/index.ts
interim17: Before pooling we reset any property that doesn't get explicitly set in the following code.

We could be more careful and reset everything when we pool, but since the following code should overwrite those properties (like agentData) I tried to make the pooling process as lean as possible.

src/test/util.test.ts
interim17: Nope, and don't have error handling for that either, but working on it now.

src/simularium/VisDataCache.ts
```diff
@@ -0,0 +1,247 @@
...
+    cacheEnabled: boolean;
+}
+
+class VisDataCache {
+    public head: LinkedListNode | null;
+    public tail: LinkedListNode | null;
+    public numFrames: number;
+    public size: number;
+    public maxSize: number;
+    public cacheEnabled: boolean;
```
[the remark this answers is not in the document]
interim17: Hmm, yeah, if I add getters, because I do read them outside the class even though I provide the method for changing them.

src/test/DummyRemoteSimulator.ts
```diff
@@ -38,39 +38,32 @@ export class DummyRemoteSimulator extends RemoteSimulator {
...
-                    1,
-                    0,
-                ],
-            };
-            bundleData.push(data);
-            frameNumber++;
-        }
-
-        msg.bundleData = bundleData;
+        const data: VisDataFrame = {
```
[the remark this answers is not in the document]
interim17: If I'm understanding right, the bundleSize would have only varied when the backend was sending more than one frame per message.

src/simularium/VisDataCache.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    private getFrameAtTimeOrFrameNumber(
+        condition: "time" | "frameNumber",
+        value: number
+    ): CachedFrame {
+        if (!this.head) {
+            return this.frameAccessError("No data in cache.");
+        }
+        const frame = this.walkLinkedList(
+            (node) => node.data[condition] === value
```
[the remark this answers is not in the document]
interim17: See questions above...

src/test/util.test.ts
interim17: Added some checks and tests for misshapen data that would result from bad offsets.

src/simularium/VisData.ts
```diff
@@ -98,91 +50,58 @@ class VisData {
...
-
-        // Find the index of the frame that has the time matching our target time
-        const frameNumber = this.frameDataCache.findIndex((frameData) => {
-            return compareTimes(frameData.time, time, this.timeStepSize) === 0;
-        });
-
-        // frameNumber is -1 if findIndex() above doesn't find a match
-        if (frameNumber !== -1) {
-            this.cacheFrame = frameNumber;
+        const frameNumber = this.frameCache.getFrameAtTime(time)?.frameNumber;
```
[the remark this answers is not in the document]
interim17: `?` checks for undefined or null, not just falsy, so 0 should not trigger it.

Happy to make this more robust, but fwiw `getFrameAtTime` just returns a call of `getFrameAtTimeOrFrameNumber` with the "time" flag provided. That function has error handling and should either return the data for the frame, or throw an error.

src/simularium/VisData.ts
```diff
@@ -194,10 +113,29 @@ class VisData {
...
+            return;
+        }
+        if (maxCacheSize < 0) {
+            this.frameCache.changeSettings({ maxSize: -1 });
+            return;
+        }
+        // cache must have at least one frame
+        const newCacheSize = maxCacheSize > 0 ? maxCacheSize : 1;
+        this.frameCache.changeSettings({ maxSize: newCacheSize });
+    }
```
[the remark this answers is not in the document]
interim17: Making this change!

Possible that I did this because this degree of chaining was starting to make me uncomfortable:
`this.props.simulariumController.visData.frameCache.changeSettings`

src/simularium/VisDataParse.ts
```diff
@@ -1,119 +1,108 @@
-import {
-    FrameData,
-    VisDataMessage,
-    AgentData,
-    AGENT_OBJECT_KEYS,
-} from "./types";
+import { VisDataMessage, AGENT_OBJECT_KEYS, CachedFrame } from "./types";
```
toloudis, line 1: is this now the only module that needs to use AGENT_OBJECT_KEYS?

src/simularium/VisDataParse.ts
```diff
@@ -1,119 +1,108 @@
...
+    return frameData;
+}
+
+function calculateBufferSize(data: number[]): number {
+    let size = AGENT_HEADER_SIZE * 4; // Header size in bytes
+    let index = 0;
+
+    while (index < data.length) {
+        size += FRAME_DATA_SIZE * 4; // Agent header size in bytes
+        const nSubPoints = data[index + FRAME_DATA_SIZE - 1];
```
toloudis, line 100: Nit: I noticed two different ways you compute the offset to subpoints.  Above here you also use `AGENT_OBJECT_KEYS.indexOf("nSubPoints")]` --  is that the same as `FRAME_DATA_SIZE-1` ?  Or are they doing different things in each place where they are used?

src/test/DummyRemoteSimulator.ts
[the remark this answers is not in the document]
toloudis: I guess that's true! So are you not testing that scenario because the client can't handle it anymore, or because it just will never happen?  Because you might as well just get rid of the bundleSize arg to `getDataBundle` then

src/util.ts
```diff
@@ -74,3 +77,68 @@ export function loadSimulariumFile(file: Blob): Promise<ISimulariumFile> {
...
         });
 }
+
+export function calculateCachedSize(parsedAgentData: AgentData[]): number {
+    let totalSize = 0;
+
+    // Calculate the size of parsedAgentDataArray
+    for (let j = 0; j < parsedAgentData.length; ++j) {
+        const agent = parsedAgentData[j];
+        totalSize += BYTES_PER_AGENT; // 10 number properties in AgentData (excluding subpoints)
```
toloudis, line 87: this comment makes me think that you should have something like AgentData.getSizeBytes() which adds the subpoints internally.

toloudis: also there are more generic ways to do this where you don't have to worry about the object changing its number of properties, like looping over Object.keys and using typeof and hasOwnProperty... I am certain there's a npm library that would do it.  But your custom way is sure to be quicker and more efficient - at a cost of remembering to change stuff if you ever modify the AgentData class.

src/util.ts
```diff
@@ -74,3 +77,68 @@ export function loadSimulariumFile(file: Blob): Promise<ISimulariumFile> {
...
+        totalSize += agent.subpoints.length * BYTE_SIZE_64_BIT_NUM;
+    }
+
+    return (totalSize += 16); // should always be 16 bytes bceause its just one frame per message with octopus
```
toloudis, line 91: what takes up the extra 16 bytes ?

toloudis (approved):
Just a few more comments.  Some more minor maintenance/readability things but two more important requests from me: make sure that the time equality check is really safe, since it is a big change from how it used to be done,  and definitely follow up to remove the web worker that parses agent data !

Overall this is a really good change and I'm glad it's going in.

src/simularium/VisDataParse.ts
```diff
@@ -1,119 +1,108 @@
...
+        // Copy agent data
+        for (let i = 0; i < FRAME_DATA_SIZE; i++) {
+            view[writeIndex++] = frame.data[readIndex++];
+        }
 
-    return {
-        parsedAgentDataArray,
-        frameDataArray,
+        // Validate data integrity
+        if (--readIndex + nSubPoints > frame.data.length) {
```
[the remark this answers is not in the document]
interim17: It was intentional!

src/constants.ts
```diff
@@ -49,3 +50,9 @@ export const nullAgent = (): AgentData => {
         subpoints: [],
     };
 };
+
+// the size of the header before the agent data in the binary file
+export const AGENT_HEADER_SIZE = 3; // frameNumber, time, agentCount
+
+export const BYTE_SIZE_64_BIT_NUM = 8;
+export const BYTES_PER_AGENT = BYTE_SIZE_64_BIT_NUM * AGENT_OBJECT_KEYS.length;
```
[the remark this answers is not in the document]
interim17: See comment below but I think I can just get rid of this, leaving the BYTE_SIZE_64_BIT but removing BYTES_PER_AGENT

src/simularium/VisData.ts
```diff
@@ -1,95 +1,47 @@
...
         const intView = new Uint32Array(data);
-        const parsedFrameData = {
-            time: floatView[1],
+
+        const frameData: CachedFrame = {
+            data: data,
             frameNumber: floatView[0],
+            time: floatView[1],
+            agentCount: intView[2],
+            size: data.byteLength + 8 * 4,
```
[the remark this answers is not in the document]
interim17: You're right. It's bytes per number * number of metadata fields in `CachedFrame`
```
const numMetadataFields = Object.keys(frameData).length - 1; // exclude "data" field
frameData.size = data.byteLength + (numMetadataFields * BYTE_SIZE_64_BIT_NUM);

File: src/simularium/VisDataParse.ts (hunk 19 above)
Comment (interim17) [reply src/simularium/VisDataParse.ts:1]: No it's also used in `getAgentDataFromBuffer`

I think you are right that things could get broken if `AGENT_OBJECT_KEYS` and the shape of `AgentData` got out of sync. Deriving one from the other would take a little bit of fussy typing probably but be doable.  I don't entirely follow the suggestion to make it part of Agent (VisAgent?) class though

File: src/simularium/VisDataParse.ts (hunk 20 above)
Comment (interim17) [reply src/simularium/VisDataParse.ts:100]: Yeah they are the same, I'll change this one to match the others

File: src/util.ts (hunk 21 above)
Comment (interim17) [reply src/util.ts:87]: I've got myself confused about my own code apparently, but this function doesn't seem to get called anymore, which makes sense because we aren't caching parsed AgentData anymore, sorry you gave it such thorough review I think it can be deleted...

File: src/util.ts (hunk 22 above)
Comment (interim17) [reply src/util.ts:91]: removed

File: src/simularium/VisDataCache.ts (hunk 16 above)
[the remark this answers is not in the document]
Comment (interim17) [reply src/simularium/VisDataCache.ts:158]: Ok I am making these less tightly bound to property names, `getFrameAtTime` and `getFrameAtFrameNumber` will provide conditional checks instead of strings, and `getFrameAtTimeOrFrameNumber` becomes `getFrameAtCondition`, it does error handling mostly. 

For the time check I am reinstating `compareTimes` but with a zero time step, I still think this a) matches what was in the code before b) is not functionally different than a straight equality.

Not giving up on this question though I've reached out to @ascibisz and adding this to lest of next steps, if we need a robust time check here I will follow up to ensure that `compareTimes` gets the proper time step, considering that out of scope for now.

File: src/simularium/VisDataCache.ts [hunk 26]
```diff
@@ -0,0 +1,247 @@
...
+            currentNode = currentNode.next;
+        }
+        return null;
+    }
+
+    public containsTime(time: number): boolean {
+        if (time < this.getFirstFrameTime() || time > this.getLastFrameTime()) {
+            return false;
+        }
+        return !!this.walkLinkedList((node) => node.data.time === time);
```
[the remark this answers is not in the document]
meganrm: `timeStepSize` gets set when you get a new trajectory in `onTrajectoryFileInfo`. My memory of this is that the website requests changes in terms of time, but the backend only knows about frame numbers so we want to make sure we're not making mistakes via conversion

src/simularium/VisData.ts
```diff
@@ -98,91 +53,58 @@ class VisData {
...
+            this.currentFrameNumber = currentData.frameNumber;
+        } else if (
+            this.frameCache.containsFrameAtFrameNumber(this.currentFrameNumber)
+        ) {
+            currentData = this.frameCache.getFrameAtFrameNumber(
+                this.currentFrameNumber
+            );
+        }
+        return currentData;
     }
```
frasercl, line 85: A few observations here:

1. Nit: since the function immediately knows its return value when any of the `if`s pass, it's not clear that you gain anything by keeping `currentData` around rather than just early-returning and inlining `return nullCachedFrame();` as the default case at the end.
2. This function makes a decent argument for `frameNumber` as the primary key of `VisDataCache`. It's possible that in practice the cache is always small enough to be efficiently traversable, but in principle a developer might be surprised that using this `get` accessor - without `()` after it or anything - triggers an O(n) operation!
3. In fact, it currently triggers two: `containsFrameAtFrameNumber` and `getFrameAtFrameNumber`, which redundantly find the same entry (or discover that it doesn't exist) twice. I'd suggest instead that `getFrameAtFrameNumber` should just handle misses without erroring, and its return type should be expanded to `CacheEntry | undefined`. Then the check above can look more like this:

```typescript
const frame = this.frameCache.getFrameAtFrameNumber(this.currentFrameNumber);
if (frame !== undefined) {
    return frame;
}
```

Mini-soapbox: in general, I'm extremely wary of throwing errors in non-catastrophic circumstances, and instead try to include the failure case in a function's return type. If the type signature of a function is a contract, errors represent a "leak" in the contract. E.g. this method already doesn't just return a `CachedFrame` - it returns a `CachedFrame` *or* it errors out - but now TypeScript doesn't know to enforce handling that case. (Note that this is far from a widely shared opinion, or errors would work differently at the language level!)

src/simularium/VisData.ts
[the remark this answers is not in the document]
frasercl: The `?` is also currently not required - `getFrameAtTime`'s current return type is just `CachedFrame` because it errors out when a frame at `time` is missing. (Hopefully it's clear from the above that I'm advocating for it to become required, rather than for it to be removed.)

src/simularium/VisData.ts
frasercl, line 1: Really into how much simpler this class becomes with `VisDataCache` around to manage data.

src/visGeometry/index.ts
frasercl: +1 to this, it's great

src/simularium/VisDataParse.ts
```diff
@@ -1,119 +1,109 @@
...
+                    frame.data.length
+                }, expected ${readIndex + nSubPoints}.`,
+                ErrorLevel.ERROR
+            );
+        }
+
+        // Copy subpoints
+        for (let i = 0; i < nSubPoints; i++) {
+            view[writeIndex++] = frame.data[readIndex++];
+        }
```
frasercl, line 68: Optional suggestion: could try something like

```typescript
const subpoints = frame.data.slice(readIndex, readIndex + nSubPoints);
view.set(subpoints, writeIndex);
readIndex += nSubPoints;
writeIndex += nSubPoints;
```

The runtime may be able to turn that `set` call into something more efficient than a loop? If so, looks like there's a spot above to try it too?

src/simularium/VisData.ts
toloudis: On deciding how serious the error is: what are possible scenarios where we might actually request a frame number that can't be found?  Are they indicative of bad data or just data that hasn't arrived yet? Could it happen in normal usage of the viewer, like moving the time slider way ahead of what's in cache?

src/simularium/VisDataCache.ts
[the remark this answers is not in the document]
toloudis: @frasercl   I think the main reason for the cache is to just give us a way to limit the amount of memory a trajectory can consume.  Fast random access lookups are pretty secondary or tertiary concern.  There was a reason to use linked lists that was related to some weird GC behavior we were seeing early on. @interim17 can maybe explain that...

src/simularium/VisData.ts
toloudis: Also this is a great point about calling both containsFrameAtNumber and getFrameAtNumber.. lookup into the data structure twice (ie walking the list twice) does seem wasteful

src/simularium/VisDataCache.ts
```diff
@@ -0,0 +1,270 @@
...
+    ): CachedFrame {
+        if (!this.head) {
+            return this.frameAccessError("No data in cache.");
+        }
+        const frame = this.walkLinkedList(condition);
+        if (frame) {
+            return frame.data;
+        }
+        return this.frameAccessError(
+            `Frame not found at provided ${condition}. Attempting to access frame ${value}.`
```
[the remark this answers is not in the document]
toloudis: yeah putting `condition` in the string does seem a bit extreme!

src/simularium/VisDataParse.ts
toloudis: the numbers of subpoints are generally not huge. I'd benchmark this if you are really wondering if the suggestion is faster.  If both variables were typedarrays then you could just do a buffer copy for sure.

src/simularium/VisDataCache.ts
```diff
@@ -0,0 +1,270 @@
...
+         * maxSize > 0 and cacheEnabled will cause cache to trim frames
+         * when incoming frame pushes size over max
+         */
+        this.head = null;
+        this.tail = null;
+        this.numFrames = 0;
+        this.size = 0;
+        this._maxSize = -1;
+        this._cacheEnabled = true;
+        this._cacheSizeLimited = this._maxSize > 0;
```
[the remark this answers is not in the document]
interim17: I like doing the default as `Infinity`! Great to remove the unnecessary property and the cacheSizeLimited getter can return `this._maxSize !== Infinity`.

It's more accurate anyway than saying the the maxSize is -1, even though conventionally I think people get what that means.

src/simularium/VisDataCache.ts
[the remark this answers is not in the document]
interim17: Hm, I see what you mean, also this isn't actually correct given some other changes that were made, since we could be looking for a `time` and not a `frameNumber`

src/simularium/VisDataCache.ts
```diff
@@ -0,0 +1,270 @@
...
+    public get cacheSizeLimited(): boolean {
+        return this._cacheSizeLimited;
+    }
+
+    private frameAccessError(msg?: string): CachedFrame {
+        throw new FrontEndError(
+            `Error accessing frame: ${msg}`,
+            ErrorLevel.WARNING
+        );
+    }
```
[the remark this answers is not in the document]
interim17: Never knew that, very nice.

src/simularium/VisData.ts
```diff
@@ -303,6 +207,26 @@ class VisData {
...
+    private addFrameToCache(frame: CachedFrame): void {
+        if (
+            this.frameCache.cacheSizeLimited &&
+            frame.size > this.frameCache.maxSize
+        ) {
+            this.frameExceedsCacheSizeError(frame.size);
+            return;
+        }
+        this.frameCache.addFrame(frame);
+    }
```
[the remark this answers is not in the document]
interim17: This also means I don't need the `VisData` methods `addFrameToCache` or `frameExceedsCacheSizeError`, which in turn means I don't really need any of the error handling I was doing in `VisData`... but that also concerns me that I'm forgetting a reason why I did all that... I'll test it out but would be great to just remove all of it and just handle the error in `addFrame`

src/simularium/VisData.ts
interim17: 1. Ok, so I like this. I'll return `nullCachedFrame()` if `hasFrames` fails,  as well as at the end.
2. I'm going to focus on addressing your more immediate concerns and then circling back to these deeper questions about Maps/ring buffers/linked lists and optimizing the cache data structure.
3. Yes to your snippet and not calling both those methods, and yes to adding undefined to the return type of `getFrameAtTime` and `getFrameAtFrameNumber`.
But what about the errors in `getFrameAtCondition`, what do you think should happen if a bad argument gets passed in? Console warnings? These methods can be called via the controller API so in theory a valid trajectory could be loaded and wacky arguments could get passed in.

src/simularium/VisDataParse.ts
interim17: Why not, let's go. We're typescript developers we shouldn't be writing loops ;)

src/simularium/VisData.ts
interim17: Ok so I'm coming along for this ride. Pretty much removing all error handling from `VisDataCache` on the assumption that if you are a function who tried to get a frame and received `undefined` it's your problem to decide what to do.

src/simularium/VisDataCache.ts
```diff
@@ -148,36 +130,40 @@ class VisDataCache {
...
+    public getFrameAtFrameNumber(frameNumber: number): CachedFrame | undefined {
+        // return this.getFrameAtCondition(
+        //     (node) => node.data["frameNumber"] === frameNumber,
+        //     frameNumber
+        // );
+        const frame = this.getFrameAtCondition(
+            (node) => node.data["frameNumber"] === frameNumber
         );
+        return frame ? frame : undefined;
     }
```
frasercl, line 164: Does the commented-out code want to stay?

src/simularium/VisDataCache.ts
```diff
@@ -78,18 +67,17 @@ class VisDataCache {
...
-    ): CacheNode | null {
+    ): CacheNode | undefined {
         let currentNode = firstNode || this.head;
         while (currentNode) {
             if (condition(currentNode)) {
                 return currentNode;
             }
             currentNode = currentNode.next;
         }
-        return null;
```
frasercl: miiiight still keep the explicit `return undefined;` for completeness, even though `undefined` is implicitly returned anyways

src/simularium/VisDataCache.ts
```diff
@@ -0,0 +1,270 @@
...
+        }
+        this.numFrames--;
+        this.size -= node.data.size;
+    }
+
+    public trimCache(incomingDataSize?: number): void {
+        while (
+            this.hasFrames() &&
+            this.size + (incomingDataSize || 0) > this._maxSize &&
+            this.head !== null
```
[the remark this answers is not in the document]
frasercl: Not really. It's okay to stay, just a bit odd.

frasercl (approved):
Looks good! Couple more comments this round - all minor, none blocking.

src/simularium/VisDataCache.ts
interim17: :')

src/simularium/VisDataCache.ts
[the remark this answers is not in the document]
interim17: I could do this inside the while loop:
```
const head = this.head
if (!head) break;
```
Not sure its a huge improvement, going to leave as is for now.

PR #413, opened 2024-09-16
Feature/vcell omezarr script

### LINKED ISSUE #403: convert test data from vcell to ome-zarr, and display in volume-viewer
Convert attached test data from vcell to ome-zarr and verify in volumeviewer.

[zarr.tgz](https://github.com/user-attachments/files/16754636/zarr.tgz)

interim17:
Closes #403

Providing a python script to convert the vcell simulation data we received into usable `ome.zarr`

The incoming vcell data is provided here for use in review and development, but will be removed before the `volume-agents` branch heads to `main`.

The `zarr` needs to be converted to int32, and then the `.zattrs` and `.zarray` files are parsed to provide the necessary arguments for the `OmeZarrWriter` from `bioio` to write out an `ome.zarr` that can be visualized on our software (tested with AGAVE).


meganrm (approved):
looks good. in the future if you have files needed for a PR but you aren't planning on committing them, you can include them in a zip directory as a comment in the PR. these files aren't large, so it won't be that big of a deal. But even when you remove files, unless you fully overwrite the history, git keeps a log of them, so you're increasing the overall size of the repo

interim17:
> looks good. in the future if you have files needed for a PR but you aren't planning on committing them, you can include them in a zip directory as a comment in the PR. these files aren't large, so it won't be that big of a deal. But even when you remove files, unless you fully overwrite the history, git keeps a log of them, so you're increasing the overall size of the repo

thanks i hadn't considered the way the git log would grow!

frasercl (approved):
LGTM!

Issue #406, opened 2024-08-26
add Volume as a geometry type in GeometryStore


toloudis:

toloudis:
Decisions to be made here are : 
where does the Volume object live? 
Do we want to require time series loading for this task or make it a separate task to update the volume over time?

Acceptance criteria:
test viewer can load a volume agent and we can console-log out to prove that the volume data actually loaded for time t=0 at least.

PR #416, opened 2024-09-25
update dependencies: typescript, three, jest

frasercl:
Review time: small (5-10min)

I had to update several dependencies on my feature branch for #406 (`feature/volume-geometry`) for a variety of reasons:

- `typescript`, to support newer interface description syntax used by the .d.ts files generated from `volume-viewer`
- `three`, to match the version used by `volume-viewer` and fix type errors
- `jest`, because npm was complaining about a dependency conflict and wouldn't let me install anything until I'd resolved it
- a couple smaller cleanups (`rimraf`, an `@types` package which should have been a dev dependency, etc.), while I was there

This maintenance task was large enough to be its own PR straight to `main` rather than cluttering the eventual PR from that feature branch. This branch is almost entirely cherry-picked off of `feature/volume-geometry`.


frasercl:
I also ended up bumping the `jest-coverage-report-action`, since it was pulling in a very old version of node that was causing problems. Looks like the new version wants to spray a whole bunch of "not covered" warnings all over the diff, which I don't think was happening before. Do we want those? Looks like they can be disabled: https://github.com/marketplace/actions/jest-coverage-report#change-annotations

meganrm (approved):

interim17 (approved):

Issue #405, opened 2024-08-26
bring volume-viewer code into simularium-viewer


toloudis:

frasercl:
Discussed with @toloudis and we agreed that adding volume-viewer the traditional way (as an npm package) may give us everything we need with minimal overall changes to the API surface. Our plan for development of this feature is:
- Create a branch on volume-viewer on which to make these API changes.
- During development of this feature, install volume-viewer as a local package, i.e. with a `dependencies` entry of the form `"@aics/volume-viewer": "file:../volume-viewer",`. Assume that developers and reviewers have volume-viewer stored at this location and on the correct branch.
- Once initial implementation of volume agents is complete, PR the branch on volume-viewer. Assess whether the changes are superficial enough to merge and release, or if we need to take a more unusual approach to integrating the code into simularium.

PR #417, opened 2024-09-28
Volume geometry type

### LINKED ISSUE #406: add Volume as a geometry type in GeometryStore

frasercl:
Review time: smallish (10-20min)

Resolves #406 (and #405?): adds a volume agent and geometry type using imports from volume-viewer, with enough rendering code to show the volume's bounding box.

NOTE: per the strategy discussed in #405, this change pulls in volume-viewer as a file dependency. The code in this PR only works if volume-viewer is built from the `simularium-volumes` branch in a sibling directory to this repo. (Also, we may have to loosen some checks to let this and future PRs on this feature merge?)

### Screenshot

you can tell it's at least loading volume metadata because the box is the right shape

image


ShrimpCryptid:
Looks like the unit tests/type checks are failing?

ShrimpCryptid (approved):
Left a couple minor questions/comments!

src/visGeometry/VolumeModel.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    public setImage(volumeObject: Volume): void {
+        this.image = new VolumeDrawable(volumeObject, {});
+    }
+
+    public setAgentData(data: AgentData): void {
+        if (this.image) {
+            this.image.setTranslation(new Vector3(data.x, data.y, data.z));
+            this.image.setRotation(new Euler(data.xrot, data.yrot, data.zrot));
+            const r = data.cr * 2;
```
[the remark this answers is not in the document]
frasercl: The `r` is definitely radius and the `c` I'm almost certain is "collision."

src/visGeometry/VolumeModel.ts
```diff
@@ -0,0 +1,39 @@
...
+import { AgentData } from "../simularium/types";
+
+// TEMPORARY HACK to make things typecheck
+import { VolumeRenderImpl } from "@aics/volume-viewer/es/types/VolumeRenderImpl";
+interface TempRayMarchedVolume extends VolumeRenderImpl {
+    boxHelper: Box3Helper;
+}
+
+export default class VolumeModel {
+    // TODO what to do with this?
```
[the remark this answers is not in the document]
frasercl: Things weren't working until this class had a `cancelled` property. I'm not 100% sure how it should be properly set though, or whether it should even apply to volumes. I'll go ahead and make this comment a bit more specific 🙃

src/visGeometry/GeometryStore.ts
```diff
@@ -378,6 +384,18 @@ class GeometryStore {
...
+        //   Should this class get a `VolumeLoaderContext` going?
+        const model = new VolumeModel();
+        this.setGeometryInRegistry(url, model, GeometryDisplayType.VOLUME);
+        const loader = await createVolumeLoader(url);
+        // TODO onChannelLoaded callback?
+        const volume = await loader.createVolume(new LoadSpec());
+        model.setImage(volume);
+        return model;
+    }
+
```
interim17, line 398: Not sure if this relevant, but the caching changes in review are removing the need for workers in that part of the code base, and there was discussion of workers affecting builds and publishing with NBSV.

Just a thought that we should use them if we need them, but if achieve 0 workers (automation fears??) then there may be some arguments for not adding them back in?

interim17 (approved):
Moving right along :)

src/visGeometry/GeometryStore.ts
toloudis: unless I am forgetting, doesn't PDB geometry still use a worker to calculate levels of detail?

frasercl: The context here may be slightly different. In volume-viewer we moved loading to a separate worker because playing volume time series requires loading and processing quite a lot of data per frame, and it was a problem when that processing blocked rendering. We have an externally available type, called `VolumeLoaderContext`, for handling offloading that processing to a worker. Questions are

1. is it as worth it to use `VolumeLoaderContext` here as it is in volume-viewer, given your concern and that simulation volumes may be smaller on average than microscope volumes; and
3. if 1, who should own `VolumeLoaderContext` and how should it grant access to it?

To be resolved by future work.

PR #419, opened 2024-10-11
Volume agent rendering

### LINKED ISSUE #406: add Volume as a geometry type in GeometryStore

frasercl:
Review size: medium (20-25min)

Resolves #406: makes volume agents render properly!

Adds some new semantics for volume agent data: if the agent has subpoints, subpoint 0 represents the time point of the volume to load and display, and the rest are indexes of channels to enable. By default, if no subpoints are included, timepoint 0 is loaded and only channel 0 is enabled.

Note that this change is accompanied by changes on the companion branch in volume-viewer, `simularium-volumes`. As with the previous PR in this series (#417), this branch doesn't work without the changes on that branch. (allen-cell-animated/volume-viewer#255 will merge those changes into the viewer proper, after which we'll be able to use a plain old npm dependency and get tests to pass here.)

Note also that I've hardcoded values for more than a few volume render settings. We may still have more semantics to figure out around whether to pick sensible defaults for these settings, or whether to add some new settings to the subpoints schema described above.

## Screenshot

image

### DIFF
--- a/src/visGeometry/VolumeModel.ts
+++ b/src/visGeometry/VolumeModel.ts
@@ -43,9 +43,9 @@ export default class VolumeModel {
         this.volume = volumeObject;
         this.drawable = new VolumeDrawable(this.volume, {});
         this.volume.addVolumeDataObserver(this);
-        this.drawable.setBrightness(0.7);
+        this.drawable.setBrightness(0.5);
         this.drawable.setGamma(0.15, 0.9, 1.0);
-        this.drawable.setDensity(0.7);
+        this.drawable.setDensity(0.1);
     }
 
     public setAgentData(data: AgentData): void {
@@ -64,6 +64,7 @@ export default class VolumeModel {
                 const numPoints = data.subpoints.length;
                 const time = numPoints > 0 ? data.subpoints[0] : 0;
                 if (this.volume.loadSpec.time !== time) {
+                    console.log(`Updating volume to time ${time}`);
                     this.volume.updateRequiredData({ time });
                 }
                 // If there are more subpoints, they are enabled channel idxes.
--- a/examples/src/simulators/VolumeSimulator.ts
+++ b/examples/src/simulators/VolumeSimulator.ts
@@ -17,6 +17,7 @@ const DEGREES_TO_RADIANS = 3.14159265 / 180.0;
 export default class VolumeSim implements IClientSimulatorImpl {
     agentdata: number[];
     size: [number, number, number];
+    curFrame: number = 0;
 
     constructor() {
         this.agentdata = [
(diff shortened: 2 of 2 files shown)


src/visGeometry/VolumeModel.ts
```diff
@@ -1,40 +1,136 @@
...
-interface TempRayMarchedVolume extends VolumeRenderImpl {
-    boxHelper: Box3Helper;
-}
-
 export default class VolumeModel {
     // TODO what to do with this `cancelled` property? Type check fails without it.
     // When should it be set, if ever; what should it be used for, if anything?
     public cancelled = false;
-    private image?: VolumeDrawable;
+    private drawable?: VolumeDrawable;
```
toloudis, line 16: If I recall, I think the "cancelled" property mentioned above is for when the trajectory has been changed but there is still geometry download requests in-flight.  We can use `cancelled` to just drop the arriving stuff.

src/visGeometry/GeometryStore.ts
```diff
@@ -384,15 +385,29 @@ class GeometryStore {
...
         // TODO should this be in a worker? Are we already in a worker here?
         //   Should this class get a `VolumeLoaderContext` going?
         const model = new VolumeModel();
         this.setGeometryInRegistry(url, model, GeometryDisplayType.VOLUME);
-        const loader = await createVolumeLoader(url);
-        // TODO onChannelLoaded callback?
-        const volume = await loader.createVolume(new LoadSpec());
+        const context = await this.getVolumeLoaderContext();
+        const loader = await context.createLoader(url);
+        const loadCallback = model.onChannelLoaded.bind(model);
```
ShrimpCryptid, line 407: Why do you need to bind the callback here? :0

src/visGeometry/VolumeModel.ts
```diff
@@ -1,40 +1,146 @@
...
+import { Lut, Volume, VolumeDrawable } from "@aics/volume-viewer";
 
 import { AgentData } from "../simularium/types";
 
-// TEMPORARY HACK to make things typecheck. TODO remove!
-import { VolumeRenderImpl } from "@aics/volume-viewer/es/types/VolumeRenderImpl";
-interface TempRayMarchedVolume extends VolumeRenderImpl {
-    boxHelper: Box3Helper;
-}
-
```
ShrimpCryptid, line 12: 🎉

src/visGeometry/VolumeModel.ts
```diff
@@ -1,40 +1,146 @@
...
+    public onChannelLoaded(_vol: Volume, channelIndex: number): void {
+        if (this.drawable) {
+            const isEnabled = this.channelsEnabled[channelIndex];
+            this.drawable.setVolumeChannelEnabled(channelIndex, isEnabled);
+            this.drawable.updateScale();
+            this.drawable.onChannelLoaded([channelIndex]);
+            if (this.volume) {
+                const histo = this.volume.getHistogram(channelIndex);
+                const min = histo.findBinOfPercentile(0.5);
+                const max = histo.findBinOfPercentile(0.983);
```
ShrimpCryptid, line 94: I saw the note you had in the PR description, maybe tag all the hardcoded values with a TODO so they're easier to track down? Totally optional

src/visGeometry/VolumeModel.ts
```diff
@@ -1,40 +1,146 @@
...
+        height: number,
+        orthoScale?: number
+    ): void {
+        if (this.drawable) {
+            const isOrtho = orthoScale !== undefined;
+            this.drawable.setIsOrtho(isOrtho);
+            if (isOrtho) {
+                this.drawable.setOrthoScale(orthoScale);
+            }
+            this.drawable.setResolution(width, height);
```
ShrimpCryptid, line 114: What does setting the resolution here do?

src/visGeometry/index.ts
```diff
@@ -161,7 +166,7 @@ class VisGeometry {
     public lightsGroup: Group;
     public agentPathGroup: Group;
     public instancedMeshGroup: Group;
-    public tempVolumeGroup: Group; // TODO remove
+    public volumeGroup: Group; // TODO remove
```
ShrimpCryptid, line 169: Is the TODO still needed?

ShrimpCryptid (approved):

interim17 (commented):
Commenting to note the chat we had about import issues yesterday, mostly so you can re-request me when you want eyes on this again.

interim17:
What's the status of this branch?

PR #425, opened 2024-10-28
remove pauseOn property

interim17:
Time Estimate or Size
=======
_tiny_

Problem
=======
The `pauseOn` property seems to be dead code. I'm fiddling with `pause` behavior on another branch and noticed this.

`pauseOn` is initialized to -1, never reassigned, and only access to do one comparison. 

We can just ask if `currentFrame < 0`, unless I'm missing something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,7 @@ class Viewer extends React.Component {
         this.viewerRef = React.createRef();
 
         this.state = {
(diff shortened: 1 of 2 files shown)


ShrimpCryptid (approved):
LGTM!

PR #427, opened 2024-11-05
fix test bed download button

interim17:
Time Estimate or Size
=======
_tiny_

Problem
=======
Somewhere along the line we broke the test-bed download button, it was losing the context and state was coming back undefined, arrow function in button `onClick` restores context

### DIFF
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -942,7 +942,7 @@ class Viewer extends React.Component<InputParams, ViewerState> {
                 >
                     Get available metrics
                 </button>
-                <button onClick={() => this.downloadFile()}>download</button>
+                <button onClick={this.downloadFile.bind(this)}>download</button>
                 <button
                     onClick={() =>
                         simulariumController.getPlotData(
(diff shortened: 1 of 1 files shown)


tyler-foster (approved):
Could this also be fixed by putting `this.downloadfile = this.downloadfile.bind(this)` in the constructor?

interim17:
> Could this also be fixed by putting `this.downloadfile = this.downloadfile.bind(this)` in the constructor?

Yes, in the constructor or right here in the prop.

I think I'll change it to the more explicit binding since that matches the rest of the component better! I'm just very used to working in functional components so jumped to this solution first.

toloudis (approved):
seems fine, but why not echo the pre-existing pattern of `()=>{this.downloadFile();}` ?

interim17:
> seems fine, but why not echo the pre-existing pattern of `()=>{this.downloadFile();}` ?

When I look at the constructor and the various `onClick` props I don't actually see a consistently applied pattern, we use a mix of arrow functions, binding in the constructor, and binding as methods are passed as props.

One of several places I think the test-bed could use some TLC.

toloudis:
I was looking up at the ~80 lines above the change and they all looked the same but I didn't look over the whole file

Issue #428, opened 2024-11-07
Test-bed viewer TLC


interim17:
Draft issue to be discussed/prioritized at dev or PM meeting.

## What needs to happen?

Test-bed viewer:

- The UI is getting very cluttered. I am personally to blame for some of this, but as we add/test features we are accumulating interactive elements above the viewport without a design scheme. A small effort to style and organize the UI would make for a better experience in using the viewer and reviewing PRs. I think this can be whipped up without UX input, its not production code.

- `Viewer.tsx` is 1000+ lines and growing, also it isn't terribly well commented or organized. The html could be dried out, constants/types could be moved, and much of the html could be factored into separate component files.

- Use consistent code patterns: A recent PR raised the question of wether methods passed as props should be bound in the constructor, bound as they are passed down, or used as arrow functions. All three approaches co-exist currently.

- Refactor from class to functional component? This is not essential, but React continues to move on from class based components. I find that my work is easier when the test bed is more similar to the production front end environment, so refactoring the viewer as a functional component could be worth doing at some point and would not be that hard.

- Worth doing a pass through on functionality to make sure that everything in the UI still works as expected.

## Why should we do this?

Increased productivity and the peace of a well kept space.

## When does this need to get done?

Not urgent, but why not now.

toloudis:
Love this ticket.  And it can be done incrementally... e.g. "convert to react functional component" is a good self contained task here.

toloudis:
I would like to ask for no overlays over the viewer canvas, please?
And keep it as easy as possible to add a new button.  This is one of the things that was good about the big monolithic file was that it's easy to just throw in a button and hook it up, to try something out. The whole point of the testbed.

interim17:
> I would like to ask for no overlays over the viewer canvas, please? And keep it as easy as possible to add a new button. This is one of the things that was good about the big monolithic file was that it's easy to just throw in a button and hook it up, to try something out. The whole point of the testbed.


No overlays, got it! Thank you for making a request!

Yes, I will make adding things as easy as possible. I know that is critical.

PR #429, opened 2024-11-09
testbed file organization

interim17:
Time Estimate or Size
=======
_medium_

Problem
=======
Starting work on #428 

Solution
========
- adds constants file
- adds types file
- moves into new `Components` directory: `RecordMoviesComponent`, `ColorPicker`, `AgentMetadata`, `ConversionForm`
- adds stylesheet
- indicates which UI elements have been factored into separate components by wrapping them in a border and some padding

Screenshot 2024-11-08 at 4 21 34 PM


examples/src/Components/AgentMetadata.tsx
```diff
@@ -1,18 +1,18 @@
...
 interface AgentMetadataProps {
     agentData: AgentData;
 }
 
 const AgentMetadata = ({ agentData }: AgentMetadataProps): JSX.Element => {
-
     const getContents = () => {
         if (agentData.instanceId === -1) {
-            return <div>No agent selected</div>;
+            return <div className="ui-container">No agent selected</div>;
```
interim17, line 11: Existing custom components get this wrapper to show what they contain, goal is to factor all the loose UI html into components and then get it into a scheme.

examples/src/Components/RecordMovieComponent.tsx
```diff
@@ -43,10 +48,15 @@ const RecordMovieComponent = ({
     };
 
     return (
-        <div>
+        <div className="ui-container">
+            <button onClick={() => setRecordingEnabled()}>
+                {isRecordingEnabled ? "Disable" : "Enable"} Recording
+            </button>
```
interim17, line 54: Moved this into the component so that recording related things are all contained inside the `ui-container` class.

examples/src/Components/ConversionForm/index.tsx
frasercl, line 1: Optional: while you're here, this file uses a lodash `map` that could be replaced by a native `map`

frasercl: InputSwitch has one too, and CollectionInput has a `map` and a `reduce`

examples/src/Viewer.tsx
```diff
@@ -507,13 +471,18 @@ class Viewer extends React.Component<InputParams, ViewerState> {
             []
         );
         const uniqueTags: string[] = [...new Set(allTags)];
-        if (isEqual(uiDisplayData, this.state.selectionStateInfo.appliedColors)) {
+        if (
+            isEqual(uiDisplayData, this.state.selectionStateInfo.appliedColors)
+        ) {
```
frasercl, line 476: idk if anyone's especially committed to it, but hot take this repo's 80 column limit is a bit too small

examples/src/Components/RecordMovieComponent.tsx
```diff
@@ -1,13 +1,18 @@
+import { set } from "lodash";
```
frasercl, line 1: linter says this is unused - accidental auto-import?

examples/src/Components/AgentMetadata.tsx
```diff
@@ -28,7 +28,7 @@ const AgentMetadata = ({ agentData }: AgentMetadataProps): JSX.Element => {
         );
     };
 
-    return <div>Agent Metadata: {getContents()}</div>;
+    return getContents();
```
frasercl, line 31: Can delete `getContents` and turn its `return`s into this component's `return`s

frasercl (approved):
Looks good, like the organization! Suggested a couple minor edits

examples/src/Viewer.tsx
interim17: Sizzler! I like it though... @meganrm what do you think about bumping the prettier printWidth to 120? We will vibe with volume-viewer then.

AICS style guide anyone? Would actually be nice if this sort of thing was consistent across repos.

examples/src/Components/ConversionForm/index.tsx
interim17: Taking suggestion to remove the lodash `map` from ConversionForm, but leaving the calls in InputSwitch and CollectionInput for now.

The input collections are objects not arrays, so that actually is the use case for the lodash `map` and `reduce` as I understand them and so feels out of scope to change rn.

examples/src/Components/RecordMovieComponent.tsx
interim17: sigh

examples/src/Viewer.tsx
meganrm: that's fine with me

meganrm (approved):

PR #432, opened 2024-11-27
test-bed fixes

interim17:
One line of this PR is relevant to making the `simularium-viewer` package build correctly
- adding the `.js` extension in `VisDataCache`

The rest is related to the test bed, there was much unhappiness in the test bed.
I think this improves things and makes it run!

- install the viewer in the `package.json` of the test bed and import things from the package instead of from `src` or `es`
- add a couple exports from the viewer that were needed
- remove `.js` extensions from imports local to the test bed
- add `tsconfig.json` to the test bed to make linter happy

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -1,28 +1,29 @@
 {
-  "name": "testbed_viewer",
-  "version": "1.0.0",
-  "description": "",
-  "main": "index.js",
-  "scripts": {
-    "serve-example": "webpack serve --config webpack.dev.js",
-    "prebuild-example": "npm install",
-    "build-example": "webpack --config webpack.dev.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
-  },
-  "author": "",
-  "license": "ISC",
-  "dependencies": {
-    "detect-collisions": "^9.2.3",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0"
-  },
-  "devDependencies": {
-    "copy-webpack-plugin": "^12.0.2",
-    "css-loader": "^6.9.1",
-    "html-webpack-plugin": "^5.6.0",
-    "mini-css-extract-plugin": "^2.7.7",
-    "webpack": "^5.95.0",
-    "webpack-cli": "^5.1.4",
-    "webpack-dev-server": "^5.0.4"
-  }
+    "name": "testbed_viewer",
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+        "serve-example": "webpack serve --config webpack.dev.js",
+        "prebuild-example": "npm install",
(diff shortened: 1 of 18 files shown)


examples/src/Viewer.tsx
```diff
@@ -24,41 +12,42 @@ import SimulariumViewer, {
...
-import ColorPicker from "./Components/ColorPicker.js";
-import RecordMovieComponent from "./Components/RecordMovieComponent.js";
-import ConversionForm from "./Components/ConversionForm/index.js";
-import AgentMetadata from "./Components/AgentMetadata.js";
-
-import { agentColors } from "./constants.js";
-import { BaseType, CustomType } from "./types.js";
+    AgentData,
+} from "@aics/simularium-viewer";
+import "../../style/style.css";
```
toloudis, line 25: this should come out of simularium-viewer too

toloudis (approved):
Yes to this.  The biggest hesitation is that you now have to rebuild the viewer library instead of having it be "watched" for changes.  I wonder if we can configure webpack to watch a dependency.  I am trying to modify the test app to use vite instead of webpack...

interim17:
> Yes to this. The biggest hesitation is that you now have to rebuild the viewer library instead of having it be "watched" for changes. I wonder if we can configure webpack to watch a dependency. I am trying to modify the test app to use vite instead of webpack...

Ah yeah, interesting point that is really nice when developing, definitely worth trying to keep that HMR

toloudis:
> Ah yeah, interesting point that is really nice when developing, definitely worth trying to keep that HMR

I have a branch that can follow this PR to switch the test app to vite!  Still haven't figured out how to have vite (or webpack) watch a build dependency. (As opposed to direct imports from src)

src/simularium/VisDataCache.ts
```diff
@@ -1,4 +1,4 @@
-import { compareTimes } from "../util";
+import { compareTimes } from "../util.js";
 import { CachedFrame, CacheNode } from "./types";
```
frasercl, line 2: do we also need a .js after types?

examples/package.json
frasercl, line 1: just curious: what effect might putting `type: "module"` in here have, rather than removing `.js` extensions from imports? Is such a thing possible or desirable?

frasercl (approved):
Looks good, I think!

examples/package.json
toloudis: I think my next PR will add this, for "vite" purposes at least. #433 (very much in draft state)

src/simularium/VisDataCache.ts
interim17: You would think so, it seems to be working fine with or without and adding it would be more consistent.

examples/package.json
interim17: Will let you take this one then Dan

PR #433, opened 2024-12-02
Feature/test app vite

toloudis:
Time Estimate or Size
=======
_medium, small but consequential changes_

Problem / Solution
=======
@toloudis did this work and @interim17 is writing this up after pushing a couple tweaks.

Move to using `vite`, following recent change to `vitest` and keeping viewer in line with general trend of AICS build tooling.

Old webpack config retained for comparison and debugging, can probably be removed once migration is complete.

Test-bed and website install seem to be running well. In general test-bed and other front ends should be more in sync as we are now importing from the built package rather than from `src`.

Unrelated to this change set: small new rule in `eslint.config.mjs` to remove some warning about numeric property names in tests.

Steps to Verify:
----------------
1. Install, run the test bed


examples/src/index.tsx
```diff
@@ -3,13 +3,11 @@ import { createRoot } from "react-dom/client";
...
 
-declare const SIMULARIUM_USE_LOCAL_BACKEND: boolean;
-
 const container: HTMLElement | null = document.getElementById("root");
 
 const root = createRoot(container!);
 root.render(
     <Viewer
-        localBackendServer={SIMULARIUM_USE_LOCAL_BACKEND}
+        localBackendServer={false /*SIMULARIUM_USE_LOCAL_BACKEND*/}
```
toloudis, line 11: I think this still needs to be fixed - I remember now that I only commented this out with the intention of it being temporary - see vite.config docs for how to provide build-time defined vars like this

interim17: Ok I'm logging off but will follow up Monday

examples/src/env.d.ts
```diff
@@ -0,0 +1,6 @@
+// env.d.ts
+declare global {
+    const SIMULARIUM_USE_LOCAL_BACKEND: boolean;
+}
+
+export {};
```
interim17, line 6: export statement needed to put the variable in the _actual_ global namespace not a local namespace named global, apparently...

examples/vite.config.ts
```diff
@@ -0,0 +1,15 @@
+import type { UserConfig } from "vite";
+import react from "@vitejs/plugin-react";
+
+export default {
+    plugins: [react()],
+    define: {
+        SIMULARIUM_USE_LOCAL_BACKEND: Boolean(
+            process.env.npm_config_localserver || false
+        ),
```
interim17, line 9: marginally more explicit/readable with the `|| false` included

meganrm (approved):
I was able to run it fine. the typecheck error is the same one that is keeping the test app from auto building in GitHub actions. I have only looked into it a little bit. I wonder if we have to go back to referencing it by the file structure?

eslint.config.mjs
```diff
@@ -59,6 +59,13 @@ export default [{
             selector: "property",
             format: ["camelCase", "UPPER_CASE"],
             leadingUnderscore: "allow",
+        }, {
+            selector: "property",
+            format: null,
+            filter: {
+                regex: "^[0-9]+$",
+                match: true,
+            }
```
toloudis, line 68: thank you for this! I was wondering how best to handle it

examples/src/env.d.ts
toloudis: This was befuddling me too.  The weird extra export statement makes sense and is also disappointing at the same time.

toloudis:
> I was able to run it fine. the typecheck error is the same one that is keeping the test app from auto building in GitHub actions. I have only looked into it a little bit. I wonder if we have to go back to referencing it by the file structure?

The bad typechecks certainly seem like a good reason to go back to importing from src.  Bummer.

interim17:
> > I was able to run it fine. the typecheck error is the same one that is keeping the test app from auto building in GitHub actions. I have only looked into it a little bit. I wonder if we have to go back to referencing it by the file structure?
> 
> The bad typechecks certainly seem like a good reason to go back to importing from src. Bummer.

I'm trying a few things on another branch to see if I can avoid it, but won't grind on it for too long.

interim17:
@toloudis @meganrm ok so this new config set up seems to work.

I set it up so that the github action runs `generateTypes` and tweaked `tsconfig` and `vite` config so they know to look in the type declarations folders that the build generates

I think its solid but would appreciate a double check,  if possible I definitely think its preferable to have the example viewer checking types in a way that doesn't directly import from `src` and this avoids that!

meganrm:
> @toloudis @meganrm ok so this new config set up seems to work.
> 
> I set it up so that the github action runs `generateTypes` and tweaked `tsconfig` and `vite` config so they know to look in the type declarations folders that the build generates
> 
> I think its solid but would appreciate a double check, if possible I definitely think its preferable to have the example viewer checking types in a way that doesn't directly import from `src` and this avoids that!

I used to go `localhost:8080/public/` to run the testbed, the terminal says `Files in the public directory are served at the root path. Instead of /public/, use /.` you should update the README too to stop directing people to `/public/`

toloudis:
There are benefits to doing the import both ways.
1. from a built package: the test viewer looks more like a "real world" client that would import from npm and things like bundling work more in the same way.  So you might find problems that wouldn't normally show up until you publish a versioned release.
2. from src: potentially much faster to iterate on if method 1 requires a manual rebuild on every change. 

So if you have a solution where vite automatically picks up changes in the actual simularium-viewer code without having to manually stop the server and rebuild, then that's a big win.

interim17:
> > @toloudis @meganrm ok so this new config set up seems to work.
> > I set it up so that the github action runs `generateTypes` and tweaked `tsconfig` and `vite` config so they know to look in the type declarations folders that the build generates
> > I think its solid but would appreciate a double check, if possible I definitely think its preferable to have the example viewer checking types in a way that doesn't directly import from `src` and this avoids that!
> 
> I used to go `localhost:8080/public/` to run the testbed, the terminal says `Files in the public directory are served at the root path. Instead of /public/, use /.` you should update the README too to stop directing people to `/public/`. Otherwise seems like it's working

Ok, I can do that. I always just ran `localhost:8080` Will update the README as long as there isn't another issue with this.

interim17:
> There are benefits to doing the import both ways.
> 
> 1. from a built package: the test viewer looks more like a "real world" client that would import from npm and things like bundling work more in the same way.  So you might find problems that wouldn't normally show up until you publish a versioned release.
> 2. from src: potentially much faster to iterate on if method 1 requires a manual rebuild on every change.
> 
> So if you have a solution where vite automatically picks up changes in the actual simularium-viewer code without having to manually stop the server and rebuild, then that's a big win.

I know you saw this but linking for others, this PR works on HMR/watchmode:
https://github.com/simularium/simularium-viewer/pull/444

I think it works pretty well? I changed the output dir for the CSS which I think is what your comment was about, let me know if  that's not what you meant

interim17:
> I used to go `localhost:8080/public/` to run the testbed, the terminal says `Files in the public directory are served at the root path. Instead of /public/, use /.` you should update the README too to stop directing people to `/public/`. Otherwise seems like it's working

@meganrm I'm glad you mentioned this. Was `/public` not resolving when you ran it? I'm going to push some commits so that it 
- correctly serves and defaults to`localhost:8080`
- uses `vite` to build the test-bed
- update the README...

but its also still serving effectively at `localhost:8080/public/` on my machine which, seems to be a default behavior of vite to satisfy that request although it warns in the console:
```
Files in the public directory are served at the root path.
Instead of /public/, use /.
```

examples/index.html
```diff
@@ -21,6 +21,7 @@
 
 </head>
 <body>
-    <main id="root"><main>
+    <main id="root"></main>
 </body>
+<script type="module" src="./src/index.tsx">    </script>
 </html>
```
frasercl, line 27:
```suggestion
<script type="module" src="./src/index.tsx"></script>
</html>

```

frasercl (approved):

PR #438, opened 2025-01-07
Fix/autoconversion

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
During refactor of `ISimulator` I wanted to make sure I hadn't broken auto-conversion of a smoldyn file via the conversion form, but it doesn't seem to be working on main to begin with.

Solution
========
This is a simple mimic of how we do it in `simularium-website`. I'm not aiming to do any update or big refactor here just restore basic functionality so that we can have a before/after on future PRs that touch this code.

Change summary:
---------------
Flag when a conversion is active in state, check that flag when receiving a file from the backend, and handle converted files appropriately.

Also restored the import of lodash `map` as it is used in the conversion process (may have been deleted in a clean up attempt that didn't test autoconversion flow), and added an alias for it, since it can be ambiguous when we are calling native vs lodash `map`.

Steps to Verify:
----------------
1. Go to conversion form, load a smoldyn file, and then see if it will convert and load

### DIFF
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -119,7 +119,7 @@ const initialState: ViewerState = {
     trajectoryTitle: "",
     initialPlay: true,
     firstFrameTime: 0,
-    followObjectData: nullAgent(),
+    followObjectData: null,
     conversionFileName: "",
 };
 
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { map, isEqual, findIndex, reduce } from "lodash";
+import { isEqual, findIndex, reduce, map as lodashMap } from "lodash";
 import { v4 as uuidv4 } from "uuid";
 import { InputParams } from "tweakpane";
 
@@ -87,7 +87,7 @@ interface ViewerState {
     initialPlay: boolean;
     firstFrameTime: number;
     followObjectData: AgentData;
-    convertingFile: boolean;
+    conversionActive: boolean;
     conversionFileName: string;
 }
 
@@ -122,7 +122,7 @@ const initialState: ViewerState = {
     initialPlay: true,
     firstFrameTime: 0,
     followObjectData: nullAgent(),
-    convertingFile: false,
+    conversionActive: false,
     conversionFileName: "",
 };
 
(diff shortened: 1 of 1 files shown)


examples/src/Viewer.tsx
```diff
@@ -430,8 +432,33 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+                },
+                this.state.conversionFileName
+            )
+            .then(() => {
+                simulariumController.gotoTime(0);
+            })
+            .then(() =>
+                this.setState({
+                    conversionActive: false,
+                    conversionFileName: "",
```
meganrm, line 449: these seem redundant since you set both of them at the same time. can you just check for the existence of the name?

examples/src/Viewer.tsx
```diff
@@ -430,8 +432,33 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+                })
+            )
+            .catch((e) => {
+                console.warn(e);
+            });
+    }
+
     public handleTrajectoryInfo(data: TrajectoryFileInfo): void {
         console.log("Trajectory info arrived", data);
+        if (this.state.conversionActive === true) {
```
meganrm, line 459:
```suggestion
        if (this.state.conversionActive) {
```

examples/src/Viewer.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { isEqual, findIndex, reduce } from "lodash";
+import { isEqual, findIndex, reduce, map as lodashMap } from "lodash";
```
meganrm, line 2: are we using another map somewhere? just curious why this can't be imported as `map`

ascibisz (approved):

examples/src/Viewer.tsx
interim17: There has been some confusion in a previous PRss whether we were using native JS map or lodash map, which led to this import being deleted.

This alias is so that it's obvious to anyone reading the code that we are using the `lodash` method and it doesn't get "cleaned up" again.

examples/src/Viewer.tsx
interim17: I suppose you could, I was trying to stay ahead of an edge case where the filename somehow was left as an empty string, I'll consolidate.

examples/src/Viewer.tsx
interim17: actually going to use `!!this.state.conversionFileName` per discussion above

meganrm (approved):

PR #441, opened 2025-01-08
Feature/isimulator

### LINKED ISSUE #440: refactor ISimulator
## What needs to happen? / Why should we do this?
There are a few old todos in the code related to the `ISimulator` interface and it is due for a refactor.
```
    // these should probably be abstracted to just "start"
    // and have a separate interface for initialization?
    // For example, "start" = initialize, and "resume" = play
   ```

Working on cache updates, particularly pre-fetching, highlighted how nice it would be to have those todos addressed and cleaned up.

The idea is to aim for a better separation of concerns, especially with code on the interface and in `RemoteSimulator` that does not pertain to other simulators, or really to any simulator. We can reduce boiler plate and `noop` methods and calls, as well as identify dead/unimplemented code and potential optimizations by making this all a little bit clearer.

The basic approach to the refactor will be to:
- move away from using the language of remote/networked simulation in abstractions and implementations that are not in fact remote
- move remote specific code either into `RemoteSimulator` or a new abstraction
- remove unimplemented methods that pertain to live mode and pre-run sims (as this development is moving in a different implementation direction)
- add documentation/comments

Related but out of scope: the todos (of a similar vintage) on the controller about streamlining the public API for initializing the controller with a simulator

### PR DESCRIPTION
**Pull request by interim17**
Time Estimate or Size

_medium-large_

Problem
=======
Closes #440 and advances #423 and #424

Solution
========
The main changes broadly speaking are:
- rename `ISimulator` methods
- remove unimplemented methods from `ISimulator`
- move some remote specific code form `ISimulator` into `RemoteSimulator`
- take non-simulation code (backend services like auto-conversion) out of `RemoteSimulator` and move into new abstraction: `OctopusClient`
- update `controller` to adapt to new methods
- remove "type checks" on simulators and replace with validity checks on back end services provided by new `OctopusClient`

Steps to Verify:
----------------
1. Try running local, client, and remote simulations in the test bed
2. Try converting a file with the new backend abstraction in place

### CODE REVIEW

File: src/controller/index.ts [hunk 1]
```diff
@@ -45,6 +47,7 @@ interface SimulatorConnectionParams {
 export default class SimulariumController {
     public simulator?: ISimulator;
     public remoteWebsocketClient?: WebsocketClient;
+    public octopusClient?: OctopusServicesClient;
```
interim17, line 50: New abstraction for handling back end services that are not specifically about running a particular simulation and so don't really belong on `RemoteSimulator`,  but also don't really belong in the connection management nuts and bolts of `WebsocketClient`. At this point it's primarily auto-conversion related.

src/simularium/ISimulator.ts
```diff
@@ -15,46 +22,27 @@ export interface ISimulator {
...
 
-    /**
-     * Connect
-     * */
-    connectToRemoteServer(address: string): Promise<string>;
-    socketIsValid(): boolean;
-    isConnectedToRemoteServer(): boolean;
-    getIp(): string;
-    disconnect(): void;
+    setErrorHandler(handler: (msg: Error) => void): void;
```
interim17, line 25: Error handling previously existed only on `RemoteSimulator` but a path to propagate any error from a simulator back to its implementation actually seems to fit alongside these other callbacks.

src/simularium/ISimulator.ts
```diff
@@ -15,46 +22,27 @@ export interface ISimulator {
...
     ): void;
 
-    /**
-     * Connect
-     * */
-    connectToRemoteServer(address: string): Promise<string>;
-    socketIsValid(): boolean;
-    isConnectedToRemoteServer(): boolean;
-    getIp(): string;
-    disconnect(): void;
```
interim17, line 25: These can be handled in an implementation specific way during a call of `initialize`

src/simularium/ISimulator.ts
```diff
@@ -15,46 +22,27 @@ export interface ISimulator {
...
+     * destroy: stop sending frames, release resources, close connections
+     * stream: request a stream of frames
+     * pause: request that frames stop being sent, keep the precomputed streaming head where it is
      *
      */
-    // these should probably be abstracted to just "start"
-    // and have a separate interface for initialization?
-    // For example, "start" = initialize, and "resume" = play
-    startRemoteSimPreRun(timeStep: number, numTimeSteps: number): void;
-    startRemoteSimLive(): void;
```
interim17, line 44: Live mode and pre-runs will be great. It is somewhat telling that the recent BioSimulators API fits the definition of a `remoteSimPreRun` but does not use this method. Several methods here have no meaningful implementation and are being removed. `sendUpdate` is flexible and available to establish any number of implementation specific protocols between a front end and an API for live and pre-run sims.

src/simularium/OctopusClient.ts
```diff
@@ -0,0 +1,76 @@
...
+interface IOctopusServices {
+    convertTrajectory(
+        dataToConvert: Record<string, unknown>,
+        fileType: TrajectoryType,
+        providedFileName?: string
+    ): Promise<void>;
+    cancelConversion(): void;
+    checkServerHealth(): Promise<void>;
+    setHealthCheckHandler(handler: () => void): void;
+}
```
interim17, line 13: When I made the argument to @ascibisz that these services don't belong on `RemoteSimulator` I wanted to move them to `WebsocketClient` but she argued and I agree that `WebsocketClient` should be more bare bones about about managing that connection, not about what calls you can make to our backend API.

I like this also as it seems more in line with open source thinking, it's plausible someone might want to set up their own websocket client and could use that interface with their own services in mind, and so keeping our back end related things in a couple more specific buckets is nice.

src/test/DummyRemoteSimulator.ts
```diff
@@ -107,18 +107,18 @@ export class DummyRemoteSimulator extends RemoteSimulator {
...
     }
-    public resumeRemoteSim(): void {
+    public stream(): void {
         this.isStreamingData = true;
     }
-    public abortRemoteSim(): void {
+    public destroy(): void {
         this.isStreamingData = false;
         this.isConnected = false;
     }
```
ShrimpCryptid, line 119: Overheard you and Dan talking about this before... what was the issue with using `abort`?

interim17: I'm forgetting the sequence of conversations.. but I think Dan mentioned something about moving away form abort? Maybe I'm making that up, and Cameron suggested destroy, which apparently is in line with some other graphics related codebases/protocols.

Our messaging protocols still have `abort` language in them, and I think both abort and destroy imply the fact that and we are not going to be able to request data from the sim we have been running without re-initializing.

src/controller/index.ts
```diff
@@ -335,25 +340,18 @@ export default class SimulariumController {
...
         this.visData.clearForNewTrajectory();
 
-        this.stop();
-
-        // Do I still need this? test...
-        // if (this.simulator) {
-        //     this.simulator.disconnect();
-        // }
-
+        const configureNewSimulator = !(
```
ShrimpCryptid, line 349: Nit:
```suggestion
        const shouldConfigureNewSimulator = !(
```

src/controller/index.ts
```diff
@@ -166,30 +168,39 @@ export default class SimulariumController {
...
-        if (this.simulator && this.simulator.socketIsValid()) {
-            this.simulator.disconnect();
+        if (this.simulator) {
+            this.simulator.destroy();
         }
 
         this.createSimulatorConnection(config);
     }
 
+    public remoteOctopusClientIsConfigured(): boolean {
```
ShrimpCryptid, line 178: Nit: Boolean convention would be `isRemoveOctopusClientConfigured`

src/simularium/OctopusClient.ts
```diff
@@ -0,0 +1,76 @@
...
+        };
+    }
+
+    public setHealthCheckHandler(handler: () => void): void {
+        this.healthCheckHandler = handler;
+        this.webSocketClient.addJsonMessageHandler(
+            NetMessageEnum.ID_SERVER_HEALTHY_RESPONSE,
+            () => this.healthCheckHandler()
+        );
+    }
```
ShrimpCryptid, line 33: Does this replace the existing message handler, or just add a new one?

ShrimpCryptid (commented):
Left a bunch of minor comments and a few questions!

src/simularium/ISimulator.ts
```diff
@@ -15,46 +22,27 @@ export interface ISimulator {
...
+     * General Simulation Control
      *
-     * Simulation Run Modes:
-     *  Live : Results are sent as they are calculated
-     *  Pre-Run : All results are evaluated, then sent piecemeal
-     *  Trajectory File: No simulation run, stream a result file piecemeal
+     * initialize: prepare a simulation for streaming, async to accommodate connection steps
+     * destroy: stop sending frames, release resources, close connections
+     * stream: request a stream of frames
+     * pause: request that frames stop being sent, keep the precomputed streaming head where it is
```
toloudis, line 38: I suggest moving these to above each of the individual functions.

src/simularium/ISimulator.ts
```diff
@@ -15,46 +22,27 @@ export interface ISimulator {
...
-    requestTrajectoryFileInfo(fileName: string): void;
-
-    pauseRemoteSim(): void;
-
-    resumeRemoteSim(): void;
-
-    abortRemoteSim(): void;
-
+    initialize(fileName: string): Promise<void>;
+    destroy(): void;
```
toloudis, line 42: Nit:  this might be my C++ or RAII bias but a class method called `destroy` to me implies that I'm destroying _this object_ (or all of its internals) , basically indicating that no one will make any valid calls on the object again.  We discussed this and I'm ok with the naming but it would be good to have a comment to say what this is supposed to do, i.e. what is the expectation of callers of this function.

src/simularium/ISimulator.ts
```diff
@@ -15,46 +22,27 @@ export interface ISimulator {
...
+    stream(): void;
+    pause(): void;
     requestSingleFrame(startFrameNumber: number): void;
-
-    // what do these mean for live mode?? can go backward but not forward
-    gotoRemoteSimulationTime(time: number): void;
+    requestFrameByTime(time: number): void;
+    requestTrajectoryFileInfo(fileName: string): void;
```
toloudis, line 47: I'd love for ISimulator to have a nice comment above each method in the interface describing expected behavior.

src/simularium/OctopusClient.ts
```diff
@@ -0,0 +1,76 @@
+import { TrajectoryType } from "../constants.js";
+import { WebsocketClient, NetMessageEnum } from "./WebsocketClient.js";
+
+interface IOctopusServices {
```
toloudis, line 4: Is there an expectation of more than one different kind of implementation of IOctopusServices?  If not, then the extra layer is not needed.

src/simularium/RemoteSimulator.ts
```diff
@@ -349,79 +318,7 @@ export class RemoteSimulator implements ISimulator {
...
-    public cancelConversion(): void {
-        this.webSocketClient.sendWebSocketRequest(
-            {
-                msgType: NetMessageEnum.ID_CANCEL_CONVERSION,
-                fileName: this.lastRequestedFile,
-            },
-            "Cancel the requested autoconversion"
-        );
-        this.lastRequestedFile = "";
+    public sendUpdate(obj: Record<string, unknown>): Promise<void> {
```
toloudis, line 321: If I remember right, I think naming the param with `_obj` would fix lint

src/test/DummyRemoteSimulator.ts
toloudis: > destroy to me implies that an object is permanently unusable and that all of its resources have been freed or are no longer accessible

ha, I just wrote a similar comment. But I think as long as the ISimulator interface file has a comment about what exactly `destroy` is supposed to do, then we can reasonably use destroy or abort or stop or terminate or byebye. (Ok maybe not the last one)

src/simularium/ISimulator.ts
ShrimpCryptid: Agreed!

src/simularium/OctopusClient.ts
interim17: Good question I think it replaces it, but the message handler that was provided before has been removed from remote simulator so this should be the only one.

src/simularium/ISimulator.ts
interim17: switching back to abort!

src/simularium/ISimulator.ts
```diff
@@ -1,60 +1,49 @@
...
+    stream(): void;
+    /** request that streaming of frames stop */
+    pause(): void;
+    /** request a single frame of data by frame number */
     requestSingleFrame(startFrameNumber: number): void;
-
-    // what do these mean for live mode?? can go backward but not forward
-    gotoRemoteSimulationTime(time: number): void;
+    /** request a single frame of data by time stamp */
+    requestFrameByTime(time: number): void;
```
toloudis, line 46: How do you feel about renaming requestSingleFrame to just requestFrame, so that it looks more like requestFrameByTime ?  (take out the word Single)

src/controller/index.ts
```diff
@@ -166,30 +168,39 @@ export default class SimulariumController {
...
-        if (this.simulator && this.simulator.socketIsValid()) {
-            this.simulator.disconnect();
+        if (this.simulator) {
+            this.simulator.abort();
         }
 
         this.createSimulatorConnection(config);
     }
 
+    public isRemoteOctopusClientIsConfigured(): boolean {
```
toloudis, line 178: I think this should be either `isRemoteOctopusClientConfigured` or `remoteOctopusClientIsConfigured`

toloudis (approved):
I left 2 more simple comments but not showstoppers

ShrimpCryptid (approved):
LGTM! Thanks for making those changes!

PR #442, opened 2025-01-30
prefetching

interim17:
Time Estimate or Size
=======
_large_
_fix for lint warnings in another open PR_

Problem
=======
Prefetching frames is continuation of caching upgrades, and necessary for front end features like buffering indicators and playback speed settings.

Solution
========

Conceptually, we are establishing a distinction between the "playback head" and the "streaming head", which will get fleshed out with more complex caching behavior, some forthcoming Octopus adjustments, and the ability to move in and out of live mode.

I'm keeping controller method names and functionality intact to avoid breaking changes. That said, I have some _opinions_ about some practices I'd like to move away from, primarily that I'd like to request frames, and use frame number, whenever possible, in preference over using simulation time. If the PR doesn't make it clear why, or if anyone disagrees, I am happy to discuss!

### CacheAndStreamingLogs
This component may not need to be a permanent part of the test-bed, could be deleted/retired once some critical caching work is done, along with its callback. But for now getting some detailed readouts from the cache are essential for dev quality of life. Production front ends don't need to subscribe to the cache logs if they are unneeded, and the test bed readout is hidden until revealed. Better design/appearance TBD. Also changes to the recording component are just to save some visual space.

### totalDuration / totalSteps and the general frame vs time issue
Time values like `totalDuration` in the test bed are derived from `totalSteps` and the `timeStep`. When we use a time value to request frames from the backend, we first have to derive it from frame number, and then Octopus has to convert it back into a frame number. Using this type of derived value for elements like the slider adds floats where we could be dealing with ints and potential front end back end sync issues. I think we should move away from calling `gotoTime` and towards using the new `movePlaybackFrame`. The functionality of `gotoTime` has been retained.

### playing/streaming states and methods
Playback tracked via `isPlaying` instead of `isPaused` and both states held at controller level.
Some functions related to playback have been retained with the same naming (`resume`/`pause`) although I would like to make them more explicit in the future like: `pausePlayback` or even `pausePreComputedSim` and `pauseLiveSim` as we distinguish playback modes.

### onStreamingChange
To subscribe the front end to changes in when we are streaming or not

### onCacheUpdate
Subscribe the front end to logging from the cache, need is dev only for now, so `currentStreamingHead` is distinguished in visData as the piece necessary for production

### getFrameAtTime
`gotoTime` is still a thing, as is `simulator.requestFrameByTime` but under the hood `gotoTime` is routed through this function, and then to `movePlaybackFrame`. When calling `gotoTime` the simulator does not actually request a frame by time.

### remoteStreamingHeadPotentiallyOutOfSync
The async nature of our websocket stream leaves the small possibility that the viewer and the backend get out of sync in terms of what the "current" frame is in cases when when a frame has been rejected due to cache limit being reached. 

This code to handle this behavior without major issues, but a long term fix to this is coming, as we will update Octopus messaging to include the ability to reset the backend's "current frame" without requesting a frame (currently the only way to do that).

Steps to Verify:
----------------
1. In the test bed you can inspect the state of playback, streaming, and caching on the `Show cache and streaming info` button. The current default will cache about 15 frames of the COVID trajectory which is my usually test trajectory. Try playing, pausing, and advancing/retreating frames.
2. The cache limit is set via a prop on the Viewer, try different limits.
3. When playing through cached frames you should see a very rapid and smooth playback, that turns slow and stepwise when displaying a live stream.


examples/src/Viewer.tsx
```diff
@@ -70,7 +72,7 @@ interface ViewerState {
     agentColors: number[] | string[];
     showPaths: boolean;
     timeStep: number;
-    totalDuration: number;
+    totalSteps: number;
```
interim17, line 75: First example of a move away from time values towards frame values, see comments in write up!

examples/src/Viewer.tsx
```diff
@@ -791,14 +820,16 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+                    step={1}
+                    value={this.state.currentFrame}
+                    max={this.state.totalSteps}
+                    onChange={this.handleScrubFrame}
                 />
                 <label htmlFor="slider">
-                    {this.state.currentTime} / {this.state.totalDuration}
+                    {this.state.currentFrame * this.state.timeStep +
+                        this.state.firstFrameTime}
+                    / {this.state.totalSteps * this.state.timeStep}
```
interim17, line 827: The slider just uses the frame numbers as it's values and derives times for display values.

ShrimpCryptid (commented):
Left some comments and suggestions!

examples/src/Components/CacheAndStreamingLogs.tsx
```diff
@@ -0,0 +1,59 @@
...
+                            ? "streaming"
+                            : "not streaming"}
+                    </div>
+                    <div>Cache Size: {size}</div>
+                    <div>Cache Enabled: {enabled ? "Yes" : "No"}</div>
+                    <div>Max Size: {maxSize}</div>
+                    <div>First Frame Number: {firstFrameNumber}</div>
+                    <div>Last Frame Number: {lastFrameNumber}</div>
+                    <div>Current playback frame: {playbackFrame}</div>
+                    <div> Current streaming head: {streamingHead}</div>
```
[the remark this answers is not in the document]
interim17: I'm agnostic, happy to change.

src/controller/index.ts
```diff
@@ -187,6 +191,21 @@ export default class SimulariumController {
         return this.isFileChanging;
     }
 
+    public setOnStreamingChange(
```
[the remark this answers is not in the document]
interim17: Sure :)

src/controller/index.ts
```diff
@@ -261,15 +280,16 @@ export default class SimulariumController {
...
+            this.handleStreamingChange(false);
+            // todo add frame argument once octopus supports this
             this.simulator.pause();
-            this.isPaused = true;
         }
     }
 
     public paused(): boolean {
-        return this.isPaused;
+        return !!this.isPlaying;
```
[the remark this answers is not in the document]
interim17: Whoops, this got changed from `get` at one point

src/controller/index.ts
```diff
@@ -261,15 +280,16 @@ export default class SimulariumController {
...
+    public pauseStreaming(): void {
         if (this.simulator) {
+            this.handleStreamingChange(false);
+            // todo add frame argument once octopus supports this
             this.simulator.pause();
-            this.isPaused = true;
         }
     }
 
     public paused(): boolean {
```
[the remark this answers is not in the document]
interim17: Could do, tried to avoid changing the names of public controller methods on this PR as those are breaking changes for front-ends.

src/controller/index.ts
```diff
@@ -278,28 +298,78 @@ export default class SimulariumController {
         }
     }
 
-    public gotoTime(time: number): void {
-        // If in the middle of changing files, ignore any gotoTime requests
+    private getFrameAtTime(time: number): number {
+        let frame = Math.round(time / this.visData.timeStepSize);
+        const frameBeyondMax = frame >= this.visData.totalSteps - 1;
+        if (frameBeyondMax) {
```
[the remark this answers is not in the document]
interim17: Just did it for readability.

src/controller/index.ts
```diff
@@ -278,28 +298,78 @@ export default class SimulariumController {
...
-            this.visData.clearCache();
-            this.simulator.requestFrameByTime(time);
+        if (this.visData.hasLocalCacheForFrame(frameNumber)) {
+            this.visData.gotoFrame(frameNumber);
+            this.resumeStreaming();
+        } else if (this.simulator) {
+            this.clearLocalCache();
+            this.visData.WaitForFrame(frameNumber);
+            this.visData.currentFrameNumber = frameNumber;
+            this.resumeStreaming(frameNumber);
```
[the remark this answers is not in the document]
interim17: Yes, this is behavior we want to make more sophisticated in the next steps. Luckily in lots of cases we would have arrived at t=50 in a way where we also cached 48, and 49, so this block wouldn't get hit. But in the situation where we jumped forward to 50, and then go back, yes we will clear the cache in an inefficient way.

src/controller/index.ts
```diff
@@ -278,28 +298,78 @@ export default class SimulariumController {
...
+
+    // pause playback
+    public pause(): void {
+        this.visData.isPlaying = false;
+        if (
+            this.visData.currentFrameNumber >
+            this.visData.frameCache.getFirstFrameNumber()
+        ) {
+            this.resumeStreaming();
+        }
```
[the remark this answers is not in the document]
interim17: This for the case where streaming may have stopped when the cache limit was reached, but we can now trim/stream to prioritize frames ahead of the playback head.

src/simularium/VisData.ts
```diff
@@ -177,14 +203,65 @@ class VisData {
         this.parseAgentsFromFrameData(msg);
     }
 
-    private addFrameToCache(frame: CachedFrame): void {
+    /**
+     * Incoming frame management
+     */
```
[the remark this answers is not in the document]
interim17: code section

src/simularium/VisData.ts
```diff
@@ -177,14 +203,65 @@ class VisData {
...
+        this.frameCache.trimCache(this.currentFrameData.size);
+        this.frameCache.addFrame(frame);
+    }
+
+    private resetCacheWithFrame(frame: CachedFrame): void {
+        this.clearCache();
+        this.frameCache.addFrame(frame);
+    }
+
+    private handleCacheOverflow(frame: CachedFrame): boolean {
```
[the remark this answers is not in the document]
interim17: Hmm, looking at this I think I'd like to split it into `doesFrameCauseCacheOverflow` and `handleCacheOverflow` instead, your comment made me realize the function is sort of muddled as it is.

src/simularium/VisData.ts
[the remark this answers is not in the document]
ShrimpCryptid: Probably better to format as 
```
////////// Section name //////////////
```
or something similar, since the `/** */` is conventionally used as as a method comment

ShrimpCryptid (approved):

src/simularium/VisDataCache.ts
```diff
@@ -185,6 +220,7 @@ class VisDataCache {
             this.tail = newNode;
             this.numFrames++;
             this.size += data.size;
+            // todo: handle this logic at a higher level
```
tyler-foster, line 223: still todo?

tyler-foster (approved):
I'm generally unfamiliar with this codebase, but at a glance this seems sound. You might want someone else to hop on as a third reviewer. Leaving a few comments.

src/simularium/VisDataCache.ts
interim17: yep

src/simularium/VisData.ts
```diff
@@ -112,6 +136,8 @@ class VisData {
 
     public clearForNewTrajectory(): void {
         this.clearCache();
+        this.currentStreamingHead = -1;
```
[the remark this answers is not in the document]
interim17: Maybe? Frame numbers start at 0 so -1 essentially means not initialized, first frame not received yet.

src/simularium/VisData.ts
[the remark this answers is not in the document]
interim17: Ok I've added some new testing!

Will add more as this feature evolves, and appreciate the feedback as our test data for `VisData.ts` was out of date and had multiframe data bundle which the back end no longer serves.

examples/src/Viewer.tsx
```diff
@@ -401,7 +419,7 @@ class Viewer extends React.Component<InputParams, ViewerState> {
         }
         this.setState({ currentFrame, currentTime });
         if (currentFrame < 0) {
-            simulariumController.pause();
+            simulariumController.pauseStreaming();
```
meganrm, line 422: this seems really specific for a user of the app to know they have to call this `pauseStreaming`. Generally all the functions we expose should be pretty intuitive. why can't this just be pause anymore?

examples/src/Viewer.tsx
```diff
@@ -477,18 +495,17 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
             currentFrame: 0,
             currentTime: 0,
             trajectoryTitle: data.trajectoryTitle,
         });
     }
 
-    public handleScrubTime(event): void {
-        simulariumController.gotoTime(parseFloat(event.target.value));
+    public handleScrubFrame(event): void {
+        simulariumController.movePlaybackFrame(parseInt(event.target.value));
```
meganrm, line 508: `goToFrame` seems a lot simpler, and would mirror the `goToTime` function. I haven't fully ingested why you decided to move to frames, but I could see an argument for keeping both

src/controller/index.ts
```diff
@@ -56,10 +56,11 @@ export default class SimulariumController {
     public postConnect: () => void;
     public startRecording: () => void;
     public stopRecording: () => void;
+    public onStreamingChange: (streaming: boolean) => void;
     public onError?: (error: FrontEndError) => void;
 
-    private isPaused: boolean;
-    private isFileChanging: boolean;
+    public isFileChanging: boolean;
+    public streaming: boolean;
```
meganrm, line 63: this should be `isStreaming`. and I am still having trouble remembering what streaming is. is it caching? could you call it `isPreFetching`? because I can think of a lot things `streaming` could be referring to in this app

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
-        // If in the middle of changing files, ignore any gotoTime requests
+    private clampFrameNumber(frame: number): number {
+        return Math.max(0, Math.min(frame, this.visData.totalSteps - 1));
+    }
+
+    private getFrameAtTime(time: number): number {
+        const frameNumber = Math.round(time / this.visData.timeStepSize);
+        const clampedFrame = this.clampFrameNumber(frameNumber);
+        return clampedFrame;
+    }
```
meganrm, line 335: previously, this was logic that was in the server, what's the reason for moving it into the viewer?

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
-            // else reset the local cache,
-            //  and play remotely from the desired simulation time
-            this.visData.clearCache();
-            this.simulator.requestFrameByTime(time);
+        if (this.visData.hasLocalCacheForFrame(clampedFrame)) {
+            this.visData.gotoFrame(clampedFrame);
+            this.resumeStreaming();
+        } else if (this.simulator) {
+            this.clearLocalCache();
+            this.visData.WaitForFrame(clampedFrame);
```
meganrm, line 348: why is this uppercase? (`WaitForFrame`)

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
+        this.visData.isPlaying = true;
+        this.resumeStreaming();
+    }
+
     public clearFile(): void {
         this.isFileChanging = false;
         this.playBackFile = "";
         this.visData.clearForNewTrajectory();
         this.simulator?.abort();
-        this.pause();
```
meganrm, line 337: why was this removed?

src/simularium/VisDataCache.ts
```diff
@@ -36,14 +39,44 @@ class VisDataCache {
...
+                maxSize: this._maxSize,
+                enabled: this._cacheEnabled,
+                firstFrameNumber: this.getFirstFrameNumber(),
+                firstFrameTime: this.getFirstFrameTime(),
+                lastFrameNumber: this.getLastFrameNumber(),
+                lastFrameTime: this.getLastFrameTime(),
+                framesInCache: this.getListOfCachedFrameNumbers(),
+            });
+        }
+    }
```
meganrm, line 70: I logged from this function, and it gets called constantly even when the simulation is paused

examples/src/Viewer.tsx
interim17: The general idea behind prefetching is that things that pause/resume playback don't necessarily pause/resume streaming, so there is some added complexity in terms of managing those two behaviors.

In general during managing playback you can still call `pause`, but there are scenarios where we have to manage the streaming behavior.

src/simularium/VisDataCache.ts
interim17: I see, was this what was causing things to be slow?

src/controller/index.ts
interim17: Sure we can change to `isStreaming` I could also try to distinguish a state for `preFetching` but they wouldn't be the same thing. `isStreaming` refers to whether not we have called `stream` or `pause`/`abort` on the current instance of `ISimulator`

src/controller/index.ts
interim17: I think there are several good arguments for prioritizing working with frame numbers over times when requesting frames from the backend. The logic is still there in the server if you want to request a frame by time, but having the front and and back end both working with frame numbers makes it easier to keep things in sync.

src/controller/index.ts
interim17: Just a typo I think?

interim17: The commit defining `WaitForFrame` is 5 years old

src/simularium/VisDataCache.ts
interim17: For context, this should get called every time we cache a frame, or remove a frame from the cache, so for a client sim like `BindingSim` that is emitting frames rapidly and continuoulsy, we're going to get a lot of calls and that is the desired behavior at least for now.

examples/src/Viewer.tsx
meganrm: but what is the case for the installing app to manage this? it seems like the viewer should be in charge of it. what are the scenarios?

src/controller/index.ts
meganrm: what does stream do? it's not prefetching frames?

src/simularium/VisDataCache.ts
meganrm: right but it's being called continuously when the simulation loads but you don't push play (so there was just the initial frame added to the cache)

src/controller/index.ts
```diff
@@ -265,15 +284,16 @@ export default class SimulariumController {
...
     }
 
-    public pause(): void {
+    public pauseStreaming(): void {
         if (this.simulator) {
+            this.handleStreamingChange(false);
+            // todo add frame argument once octopus supports this
             this.simulator.pause();
-            this.isPaused = true;
         }
```
meganrm, line 292: what's the use case for this being a public method? I'm trying to follow the logic but it feels circular. is the point to allow the app that installs this to turn on and off prefetching? if so, why wouldn't that just be controlled by the `disabledCache` prop? Or if it's controlled by a function, why does it also have to call this callback? Presumably the installing app called the function because its state changed, so it already knows that streaming is off. It seems like this is state is being stored in a lot of places

src/simularium/VisDataCache.ts
interim17: Oh ok I see the issue is that on initialization we're calling `stream` on the `simulator` but `isPlaying` is false.

This is a pretty easy fix and can default to being playing or being paused on load. We just need to make sure that for client simulators, or any context where we aren't wanting prefetching, we keep streaming and playing in sync.

examples/src/Viewer.tsx
meganrm: will totalDuration still work or will the front end have to be re-written with this change set?

src/controller/index.ts
```diff
@@ -564,6 +636,18 @@ export default class SimulariumController {
...
         this.visGeometry?.setCameraType(ortho);
     }
+
+    public isPlaying(): boolean {
+        return this.visData.isPlaying;
+    }
+
+    public currentPlaybackHead(): number {
+        return this.visData.currentFrameNumber;
+    }
```
meganrm, line 646: this isn't used anywhere, and the installing app should get this through the onTimeChange callback. I don't think we should be adding more public functions to the controller unless there is a clear use case for them to try to keep the API easy to understand, same with the function below

examples/src/Viewer.tsx
```diff
@@ -711,6 +724,17 @@ class Viewer extends React.Component<InputParams, ViewerState> {
         this.setState({ followObjectData: agentData });
     };
 
+    public handleCacheUpdate = (log: CacheLog) => {
+        this.setState({
+            cacheLog: log,
+            playbackPlaying: simulariumController.isPlaying(),
```
meganrm, line 730: I wouldn't think you should be updating the the playbackState when the cache is updated (ie when the app is paused I wouldn't think the cache would be updated, and then you risk the two states being out of sync)
this piece of state doesn't really seem needed in general.

examples/src/Viewer.tsx
```diff
@@ -1035,7 +1069,12 @@ class Viewer extends React.Component<InputParams, ViewerState> {
                         backgroundColor={[0, 0, 0]}
                         lockedCamera={false}
                         disableCache={false}
-                        maxCacheSize={Infinity} //  means no limit, provide limits in bytes, 1MB = 1000000, 1GB = 1000000000
+                        //  For no limit use Infinity. Provide limits in bytes, 1MB = 1e6, 1GB = 1e9
+                        maxCacheSize={2e6}
+                        onCacheUpdate={this.handleCacheUpdate.bind(this)}
+                        onStreamingChange={(streaming) => {
+                            this.handleStreamingChange(streaming);
+                        }}
```
meganrm, line 1077: delete this

src/controller/index.ts
meganrm: The two states that are independent are `isPlaying` and `isPrefetching`. Each will call `stream` in different conditions, but `isStreaming` is a confusing state (because it can mean 2 different situations). basic flow is 'isPlaying == true' ? do I have data? if so, render it, if not, steam and render`. `isPrefetching === true` ? do I have the next frame, if so, moved on, have I filled up the cache? if not, stream until full.

src/controller/index.ts
```diff
@@ -190,6 +194,21 @@ export default class SimulariumController {
...
+    public setOnStreamingChangeCallback(
+        onStreamingChange: (streaming: boolean) => void
+    ): void {
+        this.onStreamingChange = onStreamingChange;
+    }
+
+    private handleStreamingChange(streaming: boolean): void {
+        this.streaming = streaming;
+        this.onStreamingChange(streaming);
+    }
```
meganrm, line 206: I don't these these functions are necessary, these kinds of callback are typically used for something in response to something the installing app can't know about, ie when the time has finished updating, or when the file has finished loading and what the data is. all the other `setOn_Change` in the app are in the viewport or decedents of the viewport. The for the controller, the installing app can just call `isStreaming`

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
+        return Math.max(0, Math.min(frame, this.visData.totalSteps - 1));
+    }
+
+    private getFrameAtTime(time: number): number {
+        const frameNumber = Math.round(time / this.visData.timeStepSize);
+        const clampedFrame = this.clampFrameNumber(frameNumber);
+        return clampedFrame;
+    }
+
+    public movePlaybackFrame(frameNumber: number): void {
```
meganrm, line 337: change to `goToFrame`

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
+        if (this.streaming) {
```
meganrm, line 338: if (isPlaying || isPrefetching) this.pauseStreaming. 

if (isPlaying) pause 
if (isPrefetching) pausePrefetching

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
     }
 
-    public resume(): void {
+    public initalizeStreaming(): void {
+        if (this.simulator) {
+            this.simulator.requestFrame(0);
+            this.simulator.stream();
+            this.handleStreamingChange(true);
+        }
+    }
```
meganrm, line 370: this doesn't seem to be used

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
+
+    public resumeStreaming(startFrame?: number): void {
```
meganrm, line 372: should be private

src/controller/index.ts
```diff
@@ -304,37 +324,86 @@ export default class SimulariumController {
...
+            }
             this.simulator.stream();
-            this.isPaused = false;
+            this.handleStreamingChange(true);
         }
     }
 
+    // pause playback
+    public pause(): void {
+        this.visData.isPlaying = false;
```
meganrm, line 393: if not prefetching, and streaming, pause streaming

src/controller/index.ts
```diff
@@ -395,18 +464,21 @@ export default class SimulariumController {
                 const error = e as Error;
                 this.simulator = undefined;
                 console.warn(error.message);
-                this.isPaused = false;
+                this.visData.isPlaying = false;
```
meganrm, line 467: make this a function: this.visData.setIsPlaying()

src/controller/index.ts
```diff
@@ -395,18 +464,21 @@ export default class SimulariumController {
...
-            return this.start() // will reject if no simulator
+            return this.start()
                 .then(() => {
                     if (this.simulator) {
                         this.simulator.requestFrame(0);
                     }
                 })
+                .then(() => {
+                    this.resumeStreaming();
+                })
```
meganrm, line 481: only call if we are prefetching

src/simularium/VisData.ts
```diff
@@ -14,9 +14,14 @@ class VisData {
     private frameToWaitFor: number;
     private lockedForFrame: boolean;
 
-    private currentFrameNumber: number;
+    public currentFrameNumber: number; // playback head
+    public currentStreamingHead: number;
```
meganrm, line 18: _currentStreamingHead ?

src/controller/index.ts
meganrm: 1. change visData.goToFrame to visData.goToCachedFrame
2. future direction: visData.goToFrame which returns boolean of if it was cached, and then the controller can either clear the cache or not.

src/simularium/VisData.ts
```diff
@@ -78,13 +92,23 @@ class VisData {
...
+    }
+
     public gotoTime(time: number): void {
         const frameNumber = this.frameCache.getFrameAtTime(time)?.frameNumber;
         if (frameNumber !== undefined) {
             this.currentFrameNumber = frameNumber;
         }
     }
 
+    public gotoFrame(frameNumber: number): void {
```
meganrm, line 106: rename goToCachedFrame, future, figure out how to only check once: throw error error if not in cache

src/simularium/VisDataCache.ts
```diff
@@ -36,14 +39,44 @@ class VisDataCache {
...
+            this.cacheUpdateCallback({
+                size: this.size,
+                numFrames: this.numFrames,
+                maxSize: this._maxSize,
+                enabled: this._cacheEnabled,
+                firstFrameNumber: this.getFirstFrameNumber(),
+                firstFrameTime: this.getFirstFrameTime(),
+                lastFrameNumber: this.getLastFrameNumber(),
+                lastFrameTime: this.getLastFrameTime(),
+                framesInCache: this.getListOfCachedFrameNumbers(),
```
meganrm, line 67: feedback for future work: separate out things that we truly need for the installing app to know about, and what you're using for debugging. either in the Advanced settings panel, or a dev mode check or something.

src/simularium/VisDataCache.ts
```diff
@@ -242,6 +281,7 @@ class VisDataCache {
         this.tail = null;
         this.numFrames = 0;
         this.size = 0;
+        this.onCacheUpdate();
```
meganrm, line 284: idea: would it make sense to just have the changed values sent? 
ie, this.cacheUpdateCallback would be updated only with new values. 
maybe not helpful depending on the way this ends up being set up for the installing app/website

src/viewport/index.tsx
```diff
@@ -130,10 +132,16 @@ class Viewport extends React.Component<
...
+            onUpdate: props.onCacheUpdate,
         });
         if (props.onError) {
             this.props.simulariumController.visData.setOnError(props.onError);
         }
+        if (props.onStreamingChange) {
+            this.props.simulariumController.setOnStreamingChangeCallback(
+                props.onStreamingChange
+            );
+        }
```
meganrm, line 144: delete this

PR #444, opened 2025-01-30
watch mode for HMR

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
Ongoing updates to build system, particularly the removal of imports direct from `src` broke HMR in the test bad app.

Solution
========
New scripts, need some road testing but seem to work fine.
Use: `npm run start:watch` for watch mode/HMR. I like have the option to actually make the default non-HMR personally.


package.json
```diff
@@ -8,7 +8,12 @@
     "types": "type-declarations",
     "scripts": {
         "start": "npm run build && cd examples && npm run serve-example",
+        "start:watch": "npm-run-all --parallel watch \"serve-example:local\"",
+        "serve-example:local": "cd examples && npm run serve-example",
         "install-examples": "npm install && cd examples && npm install",
+        "watch": "npm-run-all --parallel watch:babel watch:css",
+        "watch:babel": "cross-env TS_NODE_PROJECT=tsconfig.es.json BABEL_ENV=es babel src --out-dir es --extensions .ts,.tsx,.js,.jsx --ignore 'src/test' --watch",
+        "watch:css": "postcss src/viewport/index.css --output es/style/style.css --watch",
```
toloudis, line 16: output dir is different than the one in the build step.  duplicating these complicated lines can be error prone!

interim17: Ok I think I fixed it if you were pointing to the CSS output

interim17: FWIW we could also switch this around so that HMR is the default when running `npm start` but I like having the two options, as HMR can sometimes be a performance drag when making changes you don't expect to be functional right away.

package.json
```diff
@@ -8,7 +8,12 @@
     "types": "type-declarations",
     "scripts": {
         "start": "npm run build && cd examples && npm run serve-example",
+        "start:watch": "npm-run-all --parallel watch \"serve-example:local\"",
+        "serve-example:local": "cd examples && npm run serve-example",
         "install-examples": "npm install && cd examples && npm install",
+        "watch": "npm-run-all --parallel watch:babel watch:css",
+        "watch:babel": "cross-env TS_NODE_PROJECT=tsconfig.es.json BABEL_ENV=es babel src --out-dir es --extensions .ts,.tsx,.js,.jsx --ignore 'src/test' --watch",
+        "watch:css": "postcss src/viewport/index.css --output style/style.css --watch",
```
toloudis, line 16: Adding all these things makes me nervous about complexity of build.  I wonder if the ones that match could be put close to each other, like `watch:css` next to `processCSS` and maybe rename processCSS to `build:css` or something more consistent.

toloudis: or even css:build and css:watch !  just riffing...

package.json
```diff
@@ -9,17 +9,19 @@
     "scripts": {
         "start": "npm run build && cd examples && npm run serve-example",
         "install-examples": "npm install && cd examples && npm install",
-        "gh-build": "npm run processCSS && cd examples && npm run build-example",
+        "serve-test-bed": "cd examples && npm run serve-example",
```
toloudis, line 12: I think this line should basically be "start" and you maybe don't need a separate serve-test-bed, right?

interim17: Forgot I was allowed to change "start"

toloudis: should "start" and "watch" be the same thing?  I think what I was going for in my head was to have "start" always watch the source files so you could have automatic updates.   It seems like you could handle automatic build+watch just in the "start" script.

interim17: Left a comment above about why I had them separate but happy to defer and have it always watch, easy enough to just turn dev server off if you don't want it updating as you code.

interim17: Can't seem to get it all into one script because `run-p`/`npm-run-all --parallel` doesn't want to run the shell commands, but  if I pull `cd examples && npm run serve-example` back into a separate script we can ditch "watch" and just have:
```
        "start": "run-p 'processCSS -- --watch' 'transpileES -- --watch' 'generateTypes -- --watch' 'serve-example --prefix examples'",
        "serve-example": "cd examples && npm run serve-example",
```

toloudis (approved):
so just to clarify my understanding of where this is at:
if I clone this repo and do npm start, will it launch the viewer testapp correctly?

interim17:
> so just to clarify my understanding of where this is at: if I clone this repo and do npm start, will it launch the viewer testapp correctly?

```
npm i
npm run install-examples
npm start
``

meganrm (approved):

Issue #424, opened 2024-10-23
Caching updates


interim17:
## Use Case

Refactors/updates to frame caching in the viewer, high level issue for subtasks.

Initial features requested (ability to trim cache and warming) led to a variety of refactors and optimizations

## Acceptance Criteria

- #372 
- #422
- #423
- ?

## Details

_Please provide any helpful specifications_

PR #445, opened 2025-01-31
playback speed control

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
Users want to control playback speed
Advances #424 
Advances #446 

Continuation of prefetching work and pointing at that feature branch pending review.

This feature still needs UX input and was not prioritized, but basically was building it was basically part of debugging the prefetching.

Website implementation pending UX input.

Solution
========
Added a viewer prop to let the front end set playback speed in frames per second.
Added `shouldFrameAdvance` and `advanceFrame` methods to viewport to control playback speed
In viewport:
- store `clockTimePerRender` and `lastFrameAdvanceTime`
- refactored the `animate` loop by pulling out a few functions, some renaming, and relying on a timestamp provided by `requestAnimationFrame` (instead of `Date.now()`, variables initialized with `performance.now()`
Added front end controls in test-bed to preview feature.
Added `PlaybackControls` component to test-bed to hold previously existing controls, and new speed control.

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. Try the speed buttons!


examples/src/Components/PlaybackControls.tsx
```diff
@@ -0,0 +1,89 @@
+import React from "react";
+import { PLAYBACK_SPEEDS } from "../constants";
```
interim17, line 2: I've defined a range here from 5-60 fps in the proof of concept.

The playback speed will be throttled by the animation frame rate so any value higher at 60fps.

examples/src/constants.ts
```diff
@@ -17,3 +17,6 @@ export const agentColors: string[] = [
     "#9f516c",
     "#00aabf",
 ];
+
+export const DEFAULT_PLAYBACK_SPEED_INDEX = 3;
+export const PLAYBACK_SPEEDS: number[] = [5, 10, 15, 20, 30, 45, 60];
```
interim17, line 22: I've defined this range fairly arbitarily, any number can be provided to the prop, but the playback speed will be throttled by the animation frame rate to 60 at the highest.

examples/src/Components/PlaybackControls.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const PlaybackControls: React.FC<PlaybackControlsProps> = ({
+    simulariumController,
+    totalSteps,
+    timeStep,
+    firstFrameTime,
+    currentFrame,
+    setSpeed,
+    currentSpeed,
+}) => {
```
interim17, line 22: This component includes the new speed controls, and the other existing playback controls, small refactor as part of cleaning up test-bed.

src/viewport/index.tsx
```diff
@@ -109,7 +115,12 @@ class Viewport extends React.Component<
 
     private hit: boolean;
     private animationRequestID: number;
-    private lastRenderedAgentTime: number;
+    private currentSimulationTime: number;
+    private clockTimePerRender: number;
+    private clockTimeSinceRender: number;
+    private totalClockTimeElapsed: number;
+    private clockTimeSinceFrameAdvance: number;
+    private clockTimePerFrameAdvance: number;
```
interim17, line 123: These will make more sense below, but as part of refactoring the `animate` method I pulled out variables that were scoped within that function to make them available to the whole class, and tried to name them more explicitly.

src/viewport/index.tsx
```diff
@@ -634,50 +655,79 @@ class Viewport extends React.Component<
...
+        if (this.shouldFrameAdvance()) {
+            this.advanceFrame();
+        }
+        this.stats.begin();
+        this.visGeometry.render(this.totalClockTimeElapsed);
+        if (this.recorder?.isRecording) {
+            this.recorder.onFrame();
+        }
+        this.stats.end();
+        this.lastRenderTime = Date.now();
```
interim17, line 730: Initially I was piling new code into this already somewhat hard to read function, but as I was chasing bugs I figured it might be time to factor it out on behalf of readability. It does a lot and I think having helper methods makes this easier to understand what is going on.

interim17, line 730: Initially I was piling new code into this already somewhat hard to read function, but as I was chasing bugs I figured it might be time to factor it out on behalf of readability. It does a lot and I think having helper methods makes this easier to understand what is going on.

src/viewport/index.tsx
```diff
@@ -634,50 +655,79 @@ class Viewport extends React.Component<
...
-                    this.visGeometry.update(currentFrame);
-                    this.lastRenderedAgentTime = currentFrame.time;
-                    this.updateFollowObjectData();
-                }
-            }
+    private updateAnimationTime(): void {
+        const now = Date.now();
+        this.clockTimeSinceRender = now - this.lastRenderTime;
+        this.totalClockTimeElapsed = now - this.startTime;
+        this.clockTimeSinceFrameAdvance += this.clockTimeSinceRender;
```
ShrimpCryptid, line 680: I find it a little strange that `this.clockTimeSinceRender` and `this.totalClockTimeElapsed` are saved as private variables since they can be derived on the fly and are only used in one or two places. You could have a `getClockTimeSinceLastRender()` method that does the same thing, or have this method return the calculated values.

`this.clockTimeSinceFrameAdvance` also seems like it should just be a calculated value since it's only used in one place, and instead you could save the timestamp of the last frame advance as your private variable instead of having to add up the time deltas. (It would also be slightly more accurate since adding up deltas can be inconsistent.)

src/viewport/index.tsx
```diff
@@ -402,6 +420,9 @@ class Viewport extends React.Component<
                 this.visGeometry.destroyGui();
             }
         }
+        if (playbackSpeed !== prevProps.playbackSpeed) {
+            this.clockTimePerFrameAdvance = 1000 / playbackSpeed;
```
ShrimpCryptid, line 424: Do you need to store `clockTImePerFrameAdvance` or can we just calculate it off of `props.playbackSpeed` each time?

ShrimpCryptid (approved):
Left some non-blocking comments. I think you can get rid of a lot of private variables in `viewport/index.tsx`!

src/viewport/index.tsx
interim17: I really appreciate this feedback.

Did a second pass after naming methods and they also seemed strange and one off...

Did a small refactor to use local variables inside `animate` (as before), but also utilize the timestamp from `requestAnimationFrame` to keep things in sync.

src/viewport/index.tsx
```diff
@@ -627,57 +638,98 @@ class Viewport extends React.Component<
...
+            currentFrame.agentCount > 0
+        ) {
+            this.currentSimulationTime = currentFrame.time;
+            this.dispatchUpdatedSimulationTime({
+                time: this.currentSimulationTime,
+                frameNumber: currentFrame.frameNumber,
+            });
+            this.visGeometry.update(currentFrame);
+            this.updateFollowObjectData();
+        }
```
meganrm, line 696: can you add a note about what this is doing? I haven't looked at these functions in a long time but I'm not clear on why `renderCurrentFrame` would be first checking if the time isn't matching and then doing nothing if it does match. Is there a case where this is called and currentFrame.time === this.currentSimulationTime? 
Maybe it's just the name of this function. I would think `renderCurrentFrame` would always be doing something, not just conditionally. Maybe call this, updateToCurrentFrame?

src/viewport/index.tsx
```diff
@@ -627,57 +638,98 @@ class Viewport extends React.Component<
...
+    }
+
+    public animate(timestamp?: number): void {
```
meganrm, line 699: would you mind pointing out anything that's new code, the diff is not great, but I'm assuming most of it is just moved?

meganrm:
I'm getting an error when I try to run this 
`[plugin:vite:import-analysis] Failed to resolve import "@aics/simularium-viewer/style/style.css" from "src/Viewer.tsx". Does the file exist?`

I ran npm start, and then tried npm run install-example + npm start, and then tried npm run serve example 

all three gave the same error

interim17:
> I'm getting an error when I try to run this `[plugin:vite:import-analysis] Failed to resolve import "@aics/simularium-viewer/style/style.css" from "src/Viewer.tsx". Does the file exist?`
> 
> I ran npm start, and then tried npm run install-example + npm start, and then tried npm run serve example
> 
> all three gave the same error

https://github.com/simularium/simularium-viewer/pull/456

examples/src/Components/PlaybackControls.tsx
```diff
@@ -0,0 +1,87 @@
...
+                <input
+                    name="slider"
+                    type="range"
+                    min={0}
+                    step={1}
+                    value={currentFrame}
+                    max={totalSteps}
+                    onChange={(event) =>
+                        simulariumController.movePlaybackFrame(
+                            parseInt(event.target.value)
```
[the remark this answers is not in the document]
interim17: It's not not clear to me that you should be able to operate the slider when in live simulation mode. Or I at least think it would be good to clarify.

Given that different client simulators handle time differently during their cycle of calling update, what does it mean to click ahead on the time slider in a way that generalizes for all client simulators?

I have a draft branch where I've been drafting "live mode controls" and a way to toggle in and out of live mode with a client sim configured. I'd love to pair on it if you're interested. Alternatively, you can just tell me what behavior you expect when clicking on different points on the slider while running a client sim.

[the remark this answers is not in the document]
meganrm: Yeah, I agree, I only pointed it out because it seemed like it could be an example of an edge case that could be handled better, ie, not break the simulation. But after our discussions, handling this gracefully is probably outside of the scope of this PR.

Issue #450, opened 2025-02-19
Playback modes: live mode, configured sim runs, pre-computed trajectories


interim17:
## Use Case
`IClientSimulatorImpl` defines these playback modes:

```
export declare const enum ClientPlayBackType {
    ID_LIVE_SIMULATION = 0,
    ID_PRE_RUN_SIMULATION = 1,
    ID_TRAJECTORY_FILE_PLAYBACK = 2,
    LENGTH = 3
}
```

In order to use BioSimulators (and other) APIs, and to manage playback, caching, and interaction with live/running client simulators we need to meaningfully distinguish between these playback modes in both local and remote contexts.

List of tasks (to be expanded on):

- [ ] distinguish playback modes for live sims in simularium controller
- [ ] configure `RemoteSimulator`, particularly `sendUpdate`, for running live sims (should there be more than one type of `RemoteSimulator`?)
- [ ] configure caching options for live/client simulations
- [ ] manage caching if pre-run exceeds cache limits
- [ ] build JSON writer to export live and preconfigured sim runs, capture color changes, manage discontinuities of frame that result from cache trimming
- [ ] expand to JSON writer to binary writer?
- [ ] expand user configuration of display features (geometries, etc.) on pre-run sims
- [ ] current BioSimulators API integration is a pre-run that utilizes auto-conversion, expand to live mode?
- [ ] update test bed with necessary features to manage playback modes and take to UX to explore website/other front end implementations

## Acceptance Criteria

_TBD_
_Please describe how you know this is done_

## Details

_Please provide any helpful specifications_

_TBD_

PR #451, opened 2025-02-19
Feature/brownian

interim17:
Time Estimate or Size
=======
_medium_

A lot of these changes are just moving existing functionality or data into components/constants. I will try to highlight actual changes in comments!

Problem
=======
Advances #428
Advances #450

Setting up interaction with the Brownian Motion simulator hosted on Octopus.
(currently requires running a dev branch of octopus `test/fix-json-brownian-simulator`)

Solution
========
Pulled functionality for file selection into component, some refinement of file loading and configuration methods.
Loading client sims, networked trajectories, auto-conversion, and Biosimulators API run should all still work as before.

### `FileSelect`
Ongoing effort to pull big chunks of JSX into components to make the test-bed easier to read. Agent selection functionality is unchanged, just nested down. File selection uses `selectedFile` in state over previous `playbackFile`.

### Controller and RemoteSimulator
Take an optional flag for `requestJson` in `SimulatorConnectionParams` to send the right requests to Octopus when using the brownian motion simulator

### DIFF
--- a/examples/src/Components/FileSelect.tsx
+++ b/examples/src/Components/FileSelect.tsx
@@ -1,0 +1,103 @@
+import React, { useCallback, useEffect } from "react";
+import {
+    AWAITING_CONVERSION,
+    AWAITING_SMOLDYN_SIM_RUN,
+    TRAJECTORY_OPTIONS,
+} from "../constants";
+
+interface FileSelectionProps {
+    selectedFile: string;
+    onFileSelect: (file: string) => void;
+    loadSmoldynFile: () => void;
+    clearFile: () => void;
+    loadSmoldynPreConfiguredSim: () => void;
+    setRabbitCount: (count: string) => void;
+}
+
+const FileSelection = ({
+    selectedFile,
+    onFileSelect,
+    loadSmoldynFile,
+    clearFile,
+    loadSmoldynPreConfiguredSim,
+    setRabbitCount,
+}: FileSelectionProps): JSX.Element => {
+    useEffect(() => {
+        const urlParams = new URLSearchParams(window.location.search);
+        if (urlParams.has("file")) {
+            const queryStringFile = urlParams.get("file") || "";
+            onFileSelect(queryStringFile);
+        }
+    }, [onFileSelect]);
+
+    const handleFileSelect = useCallback(
+        (file: string) => {
+            if (
+                selectedFile === AWAITING_SMOLDYN_SIM_RUN ||
+                selectedFile === AWAITING_CONVERSION
+            ) {
+                return;
+            }
+            onFileSelect(file);
+        },
+        [selectedFile, onFileSelect]
+    );
+
+    const notInList =
+        selectedFile &&
+        selectedFile !== AWAITING_SMOLDYN_SIM_RUN &&
+        selectedFile !== AWAITING_CONVERSION &&
+        !TRAJECTORY_OPTIONS.some((t) => t.id === selectedFile);
+
+    return (
+        <div className={"ui-container"}>
+            <select
+                value={selectedFile}
+                onChange={(e) => handleFileSelect(e.target.value as string)}
+                style={{ maxWidth: 200 }}
+            >
+                <option value="" disabled>
+                    choose a file
+                </option>
+                {selectedFile === AWAITING_SMOLDYN_SIM_RUN && (
+                    <option value={AWAITING_SMOLDYN_SIM_RUN} disabled>
+                        Awaiting Smoldyn Sim Run...
+                    </option>
+                )}
+                {selectedFile === AWAITING_CONVERSION && (
+                    <option value={AWAITING_CONVERSION} disabled>
+                        Awaiting Autoconversion...
+                    </option>
+                )}
+                {Object.values(TRAJECTORY_OPTIONS).map((traj) => (
+                    <option key={traj.id} value={traj.id}>
+                        {traj.name}
+                    </option>
+                ))}
+                {notInList && (
+                    <option value={selectedFile}>{selectedFile}</option>
+                )}
+            </select>
+            <br></br>
+            <button onClick={() => clearFile()}>Clear trajectory </button>
+            <button onClick={loadSmoldynFile}>
+                Convert a smoldyn trajectory with
+            </button>
+            <button onClick={loadSmoldynPreConfiguredSim}>
+                Run pre-config Smoldyn sim via BioSimulators API
+            </button>
+            <br></br>
+            <label>
+                Initial Rabbit Count:
+                <input
+                    defaultValue="100"
+                    onChange={(event) => {
+                        setRabbitCount(event.target.value);
+                    }}
+                />
+            </label>
+        </div>
+    );
+};
+
+export default FileSelection;
--- a/examples/src/Viewer.tsx
+++ b/examples/src/Viewer.tsx
@@ -19,6 +19,8 @@ import type {
     UIDisplayData,
     SelectionStateInfo,
     SelectionEntry,
+    AgentData,
+    IClientSimulatorImpl,
 } from "@aics/simularium-viewer";
 
 import PointSimulator from "./simulators/PointSimulator.ts";
@@ -34,8 +36,9 @@ import ColorPicker from "./Components/ColorPicker.tsx";
 import RecordMovieComponent from "./Components/RecordMovieComponent.tsx";
 import ConversionForm from "./Components/ConversionForm/index.tsx";
 import AgentMetadata from "./Components/AgentMetadata.tsx";
+import FileSelection from "./Components/FileSelect.tsx";
 
-import { agentColors } from "./constants.ts";
+import { agentColors, AWAITING_CONVERSION, AWAITING_SMOLDYN_SIM_RUN, TRAJECTORY_OPTIONS } from "./constants.ts";
 import { BaseType, CustomType } from "./types.ts";
 import {
     SMOLDYN_TEMPLATE,
@@ -47,14 +50,6 @@ import {
 
 import "@aics/simularium-viewer/style/style.css";
 import "./style.css";
-
-let playbackFile = "TEST_LIVEMODE_API";
-let queryStringFile = "";
-const urlParams = new URLSearchParams(window.location.search);
-if (urlParams.has("file")) {
-    queryStringFile = urlParams.get("file") || "";
-    playbackFile = queryStringFile;
-}
 
 interface ViewerState {
     renderStyle: RenderStyle;
@@ -75,16 +70,18 @@ interface ViewerState {
         template: { [key: string]: any };
         templateData: { [key: string]: any };
     } | null;
+    selectedFile: string;
     simulariumFile: {
         name: string;
         data: ISimulariumFile | null;
     } | null;
+    clientSimulator: boolean;
     serverHealthy: boolean;
     isRecordingEnabled: boolean;
     trajectoryTitle: string;
     initialPlay: boolean;
     firstFrameTime: number;
-    followObjectData: AgentData;
+    followObjectData: AgentData | null;
     conversionFileName: string;
 }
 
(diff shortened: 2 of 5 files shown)


examples/src/Viewer.tsx
```diff
@@ -181,11 +180,11 @@ class Viewer extends React.Component<InputParams, ViewerState> {
 
     public onError = (error: FrontEndError) => {
         if (error.level === ErrorLevel.ERROR) {
-            window.alert(
+            console.warn(
```
interim17, line 183: These alerts are driving me crazy, quality of life choice, happy to revert if people prefer.

examples/src/Viewer.tsx
```diff
@@ -548,95 +557,136 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+        });
+    }
+
+    private configureLocalClientSimulator(selectedFile: string) {
+        const config: { clientSimulator?: IClientSimulatorImpl } = {};
+
+        switch (selectedFile) {
+            case "TEST_LIVEMODE_API":
+                console.log("Using Live Mode API: PointSimulatorLive");
+                config.clientSimulator = new PointSimulatorLive(4, 4);
```
interim17, line 575: Same functionality as before, just separated into separate method.

examples/src/Viewer.tsx
```diff
@@ -749,85 +813,37 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+                />
                 <button onClick={() => this.translateAgent()}>
                     TranslateAgent
                 </button>
                 <button onClick={() => simulariumController.clearFile()}>
                     Clear
                 </button>
                 <button onClick={() => this.loadSmoldynFile()}>
-                    Load a smoldyn trajectory
+                    Convert a smoldyn trajectory
```
ascibisz, line 835: It looks like this is also included in `FileSelect`, so we should probably remove this one?

examples/src/Viewer.tsx
```diff
@@ -749,85 +813,37 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
                 </button>
                 <br />
                 <label>
                     Initial Rabbit Count:
                     <input
                         defaultValue="100"
-                        onChange={(event) => {this.smoldynInput = event.target.value}}
+                        onChange={(event) => {
+                            this.smoldynInput = event.target.value;
+                        }}
```
ascibisz, line 844: Similar to above comment, there's `Initial Rabbit Count` in `FileSelect` too, so this one should probably be removed

examples/src/Components/FileSelect.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <button onClick={() => clearFile()}>Clear trajectory </button>
+            <button onClick={loadSmoldynFile}>
+                Convert a smoldyn trajectory with
+            </button>
+            <button onClick={loadSmoldynPreConfiguredSim}>
+                Run pre-config Smoldyn sim via BioSimulators API
+            </button>
+            <br></br>
+            <label>
+                Initial Rabbit Count:
```
ascibisz, line 91: nit: it would be nice to have the `Run pre-config Smoldyn sim via BioSimulators API` button and `Initial Rabbit Count:` next to eachother, since that initial rabbit count is just the input value for the simulation that will be run at the click of that button

examples/src/Viewer.tsx
```diff
@@ -735,6 +785,20 @@ class Viewer extends React.Component<InputParams, ViewerState> {
         this.setState({ isStreaming: streaming });
     };
 
+    public updateBrownianSimulator = () => {
```
ascibisz, line 788: is this actually getting called anywhere?

examples/src/constants.ts
```diff
@@ -17,3 +17,124 @@ export const agentColors: string[] = [
...
+    },
+    {
+        id: "blood-plasma-1.0.simularium",
+        name: "Blood Plasma",
+        networkedTrajectory: true,
+    },
+    { id: "TEST_POINTS", name: "Point Simulation", clientSimulator: true },
+    {
+        id: "TEST_LIVEMODE_API",
+        name: "Point Simulation LIVE",
```
ascibisz, line 130: It's my understnading that both TEST_POINTS and TEST_LIVEMODE_API  are live simulations, just one uses ClientSimulator and one uses Octopus. Assuming I'm correct about that, could we change the names to be like `Point Simulation Octopus` and `Point Simulation ClientSim` or something?

examples/src/Components/FileSelect.tsx
```diff
@@ -0,0 +1,103 @@
...
+    loadSmoldynPreConfiguredSim,
+    setRabbitCount,
+}: FileSelectionProps): JSX.Element => {
+    useEffect(() => {
+        const urlParams = new URLSearchParams(window.location.search);
+        if (urlParams.has("file")) {
+            const queryStringFile = urlParams.get("file") || "";
+            onFileSelect(queryStringFile);
+        }
+    }, [onFileSelect]);
```
ShrimpCryptid, line 31: Is `onFileSelect` expected to change?

examples/src/Components/FileSelect.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const handleFileSelect = useCallback(
+        (file: string) => {
+            if (
+                selectedFile === AWAITING_SMOLDYN_SIM_RUN ||
+                selectedFile === AWAITING_CONVERSION
```
ShrimpCryptid, line 37: Should there be a separate status variable/enum apart from the `selectedFile`? It seems like a strange convention to say that we've selected the file "awaiting_conversion" if you're trying to overload `selectedFile` with the file's status.

examples/src/Components/FileSelect.tsx
```diff
@@ -0,0 +1,103 @@
...
+                    </option>
+                ))}
+                {notInList && (
+                    <option value={selectedFile}>{selectedFile}</option>
+                )}
+            </select>
+            <br></br>
+            <button onClick={() => clearFile()}>Clear trajectory </button>
+            <button onClick={loadSmoldynFile}>
+                Convert a smoldyn trajectory with
```
ShrimpCryptid, line 84: with?

examples/src/Components/FileSelect.tsx
ShrimpCryptid: Should this be something besides a `useEffect`, or should the `useEffect` dependencies be empty?

examples/src/constants.ts
```diff
@@ -17,3 +17,124 @@ export const agentColors: string[] = [
...
+        clientSimulator: true,
```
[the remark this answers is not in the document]
ShrimpCryptid: (e.g. should this be an enum instead of a bunch of separate flags?)

examples/src/Viewer.tsx
interim17: Yep meant to remove that

examples/src/Viewer.tsx
interim17: I either forgot to remove these or re-introduced them in a merge :(

examples/src/Components/FileSelect.tsx
interim17: I'll wrap them in a ui container together for now, I think a cleaner configuration ui will be desirable going forward, but out of scope

examples/src/constants.ts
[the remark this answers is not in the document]
interim17: Definitely open to doing this a better way. 

The Brownian Motion simulator hosted on octopus. It is a a "client simulator implementation" and that flag is relevant in that we may call `controller.sendUpdate` on it, unlike a remote pre-computed trajectory, but it is networked like the other pre computed network trajectories in that we need to configure a `RemoteSimulator` and web socket client to interact with it, and it takes the same net connection config as the other networked trajectories.

Some renaming is probably in order, and I don't think this is at all a final draft of how we will configure or interact with these things.

examples/src/Components/FileSelect.tsx
interim17: I think this is an improvement on what was there before, because previously the methods that triggered conversion or sent things to BioSImulators did not interact with the UI in any way, but I do see your point.

Mostly looking for a simple way to indicate that something is going on without making the logic behind `configureAndLoad` any more complicated.

examples/src/Components/FileSelect.tsx
interim17: I think yes an empty dependency array would mimic the old behavior, thanks.

examples/src/Viewer.tsx
interim17: It should be! Like above I'm not sure if I deleteed by accident or did that in a merge artifact, adding back in.

examples/src/Components/FileSelect.tsx
interim17: Ok! So I switched this up to track both `selectedFile` and `conversionFileName` and use a local state variable as the value for the `select`. Hopefully that's better!

Admittedly not the most sophisticated thing in the world, but I think it should be good enough and satisfies the above critique.

examples/src/constants.ts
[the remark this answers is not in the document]
interim17: I think the may evolve but I pushed change to track one value per object that is an enum `SimulatorModes` instead of multiple flags, I think that's better. Thanks for feedback, what do you think?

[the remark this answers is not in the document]
ShrimpCryptid: I like the enum!

examples/src/Viewer.tsx
```diff
@@ -479,28 +484,28 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
                 console.warn(e);
             });
     }
 
     public handleTrajectoryInfo(data: TrajectoryFileInfo): void {
         console.log("Trajectory info arrived", data);
-        const conversionActive = !!this.state.conversionFileName;
-        if (conversionActive) {
+        const autoConversionActive =
+            !!this.state.conversionFileName &&
```
ShrimpCryptid, line 506:
```suggestion
            this.state.conversionFileName &&
```

examples/src/constants.ts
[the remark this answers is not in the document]
ShrimpCryptid: Looking at your code it looks like everything has its own path! So yes this enum looks good 👍

examples/src/Components/FileSelect.tsx
```diff
@@ -16,57 +13,58 @@ interface FileSelectionProps {
...
-    }, [onFileSelect]);
+    }, []);
+
+    useEffect(() => {
+        if (conversionFileName !== "") {
+            setSelectValue("Awaiting file conversion...");
+        } else {
+            setSelectValue(selectedFile);
+        }
+    }, [selectedFile, conversionFileName]);
```
ShrimpCryptid, line 39: This probably doesn't need to be a `useEffect` + state? 
```suggestion
const isAwaitingFileName = conversionFileName !== "";
const selectValue = isAwaitingFileName ? "Awaiting file conversion..." : selectedFile;
```

examples/src/Components/FileSelect.tsx
```diff
@@ -16,57 +13,58 @@ interface FileSelectionProps {
...
+        selectedFile && !TRAJECTORY_OPTIONS.some((t) => t.id === selectedFile);
 
     return (
         <div className={"ui-container"}>
             <select
-                value={selectedFile}
+                value={selectValue}
                 onChange={(e) => handleFileSelect(e.target.value as string)}
                 style={{ maxWidth: 200 }}
+                disabled={conversionFileName !== ""}
```
ShrimpCryptid, line 60: Higher-level feedback, but I feel like making a boolean constant (`isAwaitingFileConversion`) based on this would make it easier to understand the code. It's not immediately obvious to me that the conversion filename being non-empty means something?

It could also be a sign that `conversionFileName` could be `pendingConversionFileName` to indicate that it's a temporary value.

ShrimpCryptid (approved):
Left some last comments, thanks for making changes!

ascibisz (approved):
LGTM!

PR #452, opened 2025-02-19
test-bed layout

interim17:
Time Estimate or Size
_small_
The viewer diff is large because lot's of things are _moving_, but not really changing.

Problem
=======
Advances #428 
Test bed UI updates are ongoing, this is a big fix to the layout.

Solution
========
Updated the test-bed layout, pulled some functionality for agent selection into a component, refined file loading and configuration methods.

### `AgentSelection`
Agent selection functionality is unchanged, just nested down.

### Layout
Fixing the viewer next to a sidebar and header of UI and resizing it so the whole canvas is always visible. For me, a big quality of life change. The purple is a personal choice, thoughts? Pink era?

As requested by @toloudis there are no overlays and I think added functionality should be as simple as ever, devs can choose to style it or just slap it into the sidebar like we used to do at the top of the JSX.

Screenshots (optional):
-----------------------

Screenshot 2025-02-18 at 5 17 43 PM


examples/src/Viewer.tsx
```diff
@@ -812,286 +817,321 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
+                                        onClick={() =>
+                                            this.updateBrownianSimulator()
+                                        }
+                                    >
+                                        Update (Live Mode)
+                                    </button>
+                                    <button
+                                        onClick={() => this.translateAgent()}
+                                    >
+                                        Translate Agent (point sim live)
```
ascibisz, line 940: Do we need both of these buttons? When I'm running this locally, I _think_ both of these buttons successfully update the trajectory both with a local ClientSimulator and an Octopus live simulation. The only difference I noticed is the `Update (Live Mode)` only moves one agent while the `Translate Agent (point sim live)` moves two agents, and the `Translate Agent (point sim live)` moves the two agents slightly more. The wording on the buttons led me to feel like only one of the buttons would do anything for a given simulation, but since both seemed to work, I got a little confused.

ascibisz (approved):
This looks so much better than the example viewer looked before!! I did a lot of testing using this branch and it seems to be working well :)

examples/src/Viewer.tsx
interim17: So what actually need is 
- typing on the interface each client simulator accepts when  `updateSimulationState` gets called
- controls components specific to the client simulator in question

Each one of these functions was defined to work with a specific simulator, so the fact that they are interoperable is a bug not a feature, unless we want to define universal controls for client simulators, but I can't imagine what they would be.

I have something sketched out (below), but it's too much for this round of PRs. This is defining configs at the test-bed level, but they could also be encapsulated in the client sims if we wanted them to define their own controls and defaults. I think leaving these buttons is fine for now, serves as a TODO for me.

Something like this in the configs in future:

```
    TEST_LIVEMODE_API: {
        id: "TEST_LIVEMODE_API",
        name: "Point Simulation LIVE",
        description: "Real-time point-based simulation",
        ControlComponent: PointSimLiveControls, // react component with typed sendUpdate calls
        createSimulator: (config, controller) => {
            const sim = new PointSimulatorLive(config.count, config.spacing);
            controller.goToLiveMode();
            return sim;
        },
        defaultConfig: {
            count: 4,
            spacing: 4,
        },
    },

toloudis: This is exactly why the packet of data sent to the live mode simulator is not typed: we weren't ready to start designing a generic simulation api just yet.    So we figured you could just send name-value json-like data and the native simulation backend would have to figure it out by looking for names it recognizes.  We didn't want the viewer to have to have type information for each possible simulator, and instead preferred an abstraction.

toloudis (approved):
1. For large numbers of agent types, does this push the rest of the UI down off the page?  I might consider putting things that are dynamic and might have to scroll, down lower so they don't hide the more static ui elements
2. Does the viewport just stretch to fill the page?
3. it looks like you need a newline before "tick length"

examples/src/Viewer.tsx
interim17: Right, I think thats great. I'm not proposing a generic simulation API, the sketch above is more about conditional rendering controls so that we can develop client sims and the UI doesn't fill up with buttons that fire off json packets that may or may not trigger things in the sim you are running.

interim17:
> 1. For large numbers of agent types, does this push the rest of the UI down off the page?  I might consider putting things that are dynamic and might have to scroll, down lower so they don't hide the more static ui elements
> 2. Does the viewport just stretch to fill the page?
> 3. it looks like you need a newline before "tick length"

1) Ok good idea, I'll bump it down and make sure it scrolls after a certain height.
2) Yes the viewport stretches, `this.resize` updates the height and width props.
3) thank you

PR #455, opened 2025-02-21
bug fixes

interim17:
Time Estimate or Size
=======
_tiny_

Problem / Solution
=======
fixes two bugs
- test bed color picker wasn't working for agents without display states
- decrementing read index in `VisDataParse` was grabbing data off by one index when parsing fiber agent sub points

### DIFF
--- a/examples/src/Components/ColorPicker.tsx
+++ b/examples/src/Components/ColorPicker.tsx
@@ -30,7 +30,7 @@ const ColorPicker = ({
         if (!agent) {
             throw new Error("No agent found");
         }
-        if (!agent.displayStates.length) {
+        if (agent.displayStates.length === 0) {
             setSubAgents([{ name: "<unmodified>", id: "<unmodified>" }]);
             return;
         }
--- a/examples/src/Components/ColorPicker.tsx
+++ b/examples/src/Components/ColorPicker.tsx
@@ -30,6 +30,10 @@ const ColorPicker = ({
         if (!agent) {
             throw new Error("No agent found");
         }
+        if (!agent.displayStates.length) {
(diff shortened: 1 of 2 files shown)


src/simularium/VisDataParse.ts
```diff
@@ -57,7 +57,7 @@ function parseVisDataMessage(visDataMsg: VisDataMessage): CachedFrame {
         writeIndex += FRAME_DATA_SIZE;
 
         // Validate data integrity
-        if (--readIndex + nSubPoints > frame.data.length) {
```
meganrm, line 60: was this right at some point? or has this always been a bug? (just curious why it was like this)

interim17: It's a bug I introduced in a recent caching PR, it grabs the subpoints for fibers on JSON data one index early so you repeat the subpoint count as the first subpoint position, and then miss the last one.

examples/src/Components/ColorPicker.tsx
```diff
@@ -30,6 +30,10 @@ const ColorPicker = ({
         if (!agent) {
             throw new Error("No agent found");
         }
+        if (!agent.displayStates.length) {
```
frasercl, line 33: might be a bit clearer to say `=== 0` here

frasercl (approved):
I encountered subpoints off-by-one errors working on volumes as well. Good to see this fix go in at the parsing level.

interim17:
> I encountered subpoints off-by-one errors working on volumes as well. Good to see this fix go in at the parsing level.

Sorry about that!

meganrm (approved):

PR #456, opened 2025-02-21
build fix

interim17:
Time Estimate or Size
=======
_tiny_

Problem / Solution
=======
Test bed couldn't resolve import of main app stylesheet, I think because output dir of css build step wasn't right.

After deleting `es` both lockfiles and both `node_modules` then installing and building from scratch, this test-bed worked for me, and this build worked in a local website instance, and both showed the advanced controls when hitting ctrl-option-1 (which is what that stylsheet is for).

Add `dev` script for installing from scratch. 


toloudis (approved):

ShrimpCryptid (approved):

hughes036 (approved):

hughes036:
This does fix it for me (the test bed loads), though some controls appear to not be working.

interim17:
> This does fix it for me (the test bed loads), though some controls appear to not be working.

Let's chat at standup about what's not working, test-bed is in a somewhat transitional moment

PR #457, opened 2025-02-25
get first frame during initialize and start paused

interim17:
Time Estimate or Size
=======
_medium_

Problem
=======
@meganrm found some bugs resulting from recent prefetching branch merge

Solution
========

These changes:
- use `isPreFetching` (along with `isPlaying`) as state variable to control prefetching behavior (previously: `streaming`)
- clarify logic related to frame selection, streaming, and cache overflow handling
- update default `totalSteps` in client simulators to be `Infinity` because frame clamping in `goToFrame` was conflicting with `totalSteps` that were lower than the number of frames actually emitted by the binding sim
- update simulator `initialize` methods to always request frame 0 during initialization
- remove some unneeded streaming change callback from viewport callbacks
- pare down cache and streaming log display in the test-bed

This is intended as a prompt fix for bugs, and implies a few ongoing pieces of work:
- #458 
- #459 
- clarifying the controls/methods available to live client sims and handling frame requests from client sims
- using promises when sending sequences of websocket requests to ensure proper order


src/simularium/RemoteSimulator.ts
```diff
@@ -247,6 +247,9 @@ export class RemoteSimulator implements ISimulator {
                     "Start Trajectory File Playback"
                 );
             })
+            .then(() => {
+                this.requestFrame(0);
+            })
```
meganrm, line 252: Since the websocket request is not a promise my understanding is there is no guarantee  that the request has completed before this next block is executed. Could that cause any problems?

src/test/DummyRemoteSimulator.ts
```diff
@@ -119,11 +119,15 @@ export class DummyRemoteSimulator extends RemoteSimulator {
...
-        });
+        return this.connectToRemoteServer()
+            .then(() => {
+                this.fileName = fileName;
+                this.isStreamingData = true;
+                this.lastRequestedFile = fileName;
+            })
+            .then(() => {
+                this.requestFrame(0);
+            });
```
meganrm, line 130: I feel like I asked about this before, but can’t this.requestFrame(0) just be in the previous block?

src/controller/index.ts
```diff
@@ -433,7 +412,7 @@ export default class SimulariumController {
         // calls simulator.abort()
         this.stop();
 
-        this.visData.WaitForFrame(0);
+        this.visData.waitForFrame(0);
```
meganrm, line 415: What’s the difference between waitForFrame and requestFrame?

src/simularium/RemoteSimulator.ts
interim17: In practice, they seem to be landing in the proper order without issue, but you're right that there is no guarantee without a promise, which theoretically makes a race condition here.

The most precise fix here would be to change `sendWebSocketRequest` to return a promise. Hackier fix would be to resolve a Promise here with a tiny arbitrary delay, or  just leave as is for the moment because things seem to be landing in the correct order.

src/test/DummyRemoteSimulator.ts
interim17: Yes

src/simularium/RemoteSimulator.ts
meganrm: could you move the this.requestFrame within the webSocketRequest switch bracket?

meganrm:
can you explain why you needed to "update simulator initialize methods to always request frame 0 during initialization"? In the past when a trajectory has loaded it's always started paused and displayed first frame, so how was that happening before and why did you need to add requestFrame(0) to every simulator?

PR #467, opened 2025-03-17
controller params (controller maintenance 1/n)

interim17:
Time Estimate or Size
=======
_small_

points at `feature/controller-updates`. To be followed by bigger changes set to refined simulator connection params.

All code removal, so review should be quick IF this is a good idea. Big if, context needed on whether these methods are safe to remove.

Have a provisional thumbs up from @toloudis on removing controller params. Hoping @meganrm can weigh in on `simulariumController.connect`...

Problem
=======
### Controller params
Old code paths allowed us to initialize the controller with params, but we don't do this in production anymore. We only use controller params to load the DummyRemoteSimulator for testing. We can initialize a dummy simulator by different means which will allow to streamline the controller a bit, by removing this params pathway.

### `simulariumController.connect`
Then a method, `simulariumController.connect` could be called to configure the websocket client, and call `postConnect` callback when successful.
A truthy value for the viewer prop `loadInitialData` would prompt the viewport to set `initializeTrajectoryFile` as the `postConnect` callback.
None of our production apps seem to use any part of this flow.

As we refine the controller and simulator params and add testing to the controller, I'd like to remove these methods. They seem to be part of an older paradigm for loading trajectories into the viewer.

Is anyone aware of users or apps that still call `controller.connect()` or `controller.initializeTrajectoryFile()`? If so we can retain this code and fold it into the new updates.

Solution
========

We are removing public methods from the controller and props from the viewer, so this will be a breaking change that requires updates to front end apps.


src/viewport/index.tsx
```diff
@@ -243,7 +241,6 @@ class Viewport extends React.Component<
         const {
             backgroundColor,
             simulariumController,
-            loadInitialData,
```
meganrm, line 246: does this code path come from having a trajectory in the url? does that still work on this branch?

interim17: Our urls work, third party does not, but it's not because of this code path and I have a branch to fix that.

src/test/SimulariumController.test.ts
```diff
@@ -1,26 +1,12 @@
...
-                    expect(controller.time()).toEqual(2);
-                    done();
-                }, 500);
-            }));
+        test("Go to time in cache", () => {
+            // todo new approach to testing dummy simulator comming in
+            // changes to simulator params
+            const controller = new SimulariumController();
+            expect(controller).toBeDefined();
+        });
```
interim17, line 10: The other test file was an exact duplicate of this one.

meganrm (approved):

PR #468, opened 2025-03-17
discussion draft: simulator connection params

interim17:
Time Estimate or Size
=======
_large_

Problem
=======
`CreateSimulatorConnection` used to take several optional arguments that applied to conflicting cases, and a long standing todo called for "refining" these params.

In addition to adding typing and testing to simulator configuration, these changes allow us to simplify what the front end is responsible for when running auto-conversion or querying the BioSimulators API, and reduce the number of times we reconfigure the websocket connection during file changing.

This is a large change set but 696+ of these lines new lines are new tests or new dummy classes for testing purposes. I know it's still a big read, but the direct changes to the controller are more targeted than the line count implies.

I could make a separate PR for the testing if that would be less of a burden for reviewers.

These refinements are good to do on their own, and make third party API integrations easier for the front it. They are also a good set up for work on prefetching where lack of tests and bugs during file changing hid some problems in the previous attempt on that feature.

Solution
========

### Front end changes
Thanks to the `octopusClient` now checking for conversion completion, the front end can dispense with `receiveConvertedFile` and tracking the `conversionFileName`. Tracking `isAwaitingFileConversion` is not functionally necessary, but nice if a conversion is taking a while.

### `SimulatorParams`
A union type of `RemoteSimulatorParams`, `LocalProceduralSimulatorParams`, and `LocalFileSimulatorParams`.
This is now the type that `changeFile` and `createSimulatorConnection` accept.
Type guards are defined in the utils to allow for explicit checks of which kind of params we are dealing with when trying to build a simulator connection.

### `buildSimulator`
Handles the conditional logic for different simulator types and returns a value, which makes that value easily testable, along with being a public method.

### `createSimulatorConnection`
More readable and concise now, this sets the returned value of `buildSimulator` on the class. It also remains private, `buildSimulator` can be tested directly but `createSimulatorConnection` cannot be called from outside the class.

### `configureNetwork` / `needsNewNetworkConfig`
Old version of configure network wiped/configured the simulator. Now that we have the `octopusClient` we can configure the network independent of a simulator. Separating those concerns along with checking that our network config hasn't changed, means we keep our `octopusClient` and `remoteWebsocketClient` when the config hasn't changed.

### `start`
`start` is now used by three functions: `changeFile`, `startSmoldynSim` and `convertTrajectory`
To keep things dry I bumped the request of frame 0 inside of start, to make it part of the default initialization flow.

### `onConversionComplete` / `setUpConversion`
The octopus client now has a callback it can run when a converted files info arrives. We pass `controller.start` as the callback when we initialize conversion, and remove it when we cancel conversion. This is the change the frees the front end from having to track the conversion file name. All the front end needs to do now is call `convertTrajectory` or `startSmoldynSim`

### `changeFile` / `initNewFile`
`initNewFile` was primarily responsible for checking if we needed to retain our old remote connection, and then requesting frame zero.
Those responsibilities are handled by other `start` and `createSimulatorConnection` now so I don't think it's really needed anymore. A pared down `changeFile` handles everything.


src/controller/index.ts
```diff
@@ -78,90 +73,136 @@ export default class SimulariumController {
...
+        return !!this.simulator;
+    }
+
+    public get ensureOctopusClient(): OctopusServicesClient {
+        if (!this.octopusClient || !this.octopusClient.socketIsValid()) {
+            throw new Error(
+                "Remote Octopus client is not configured or socket is invalid."
+            );
+        }
+        return this.octopusClient;
```
interim17, line 170: This getter allows us to call `octopusClient` methods like this:
`ensureOctopusClient.cancelConversion()`
rather than 
`if (this.octopusClient) { octopusClient.cancelConversion }`
It's a DRYing convenience as I found myself writing that a lot, not essential to keep if anyone takes issue with that pattern.

src/test/BinaryFile.test.ts
```diff
@@ -11,7 +11,10 @@ function pad(buf: ArrayBuffer): ArrayBuffer {
         return newbuf;
     } else return buf;
 }
-function makeBinary(blocks: ArrayBuffer[], blockTypes: number[]): ArrayBuffer {
+export function makeBinary(
+    blocks: ArrayBuffer[],
+    blockTypes: number[]
+): ArrayBuffer {
```
interim17, line 17: exported these to use the controller test file for making local file params

src/test/DummyOctopusClient.ts
```diff
@@ -0,0 +1,58 @@
...
+    constructor(opts: NetConnectionParams) {
+        const webSocketClient = new WebsocketClient(opts);
+        super(webSocketClient);
+        this.webSocketClient = webSocketClient;
+        this.lastRequestedFile = "";
+        this.conversionCompleteDelay = 2000;
+        this.onConversionComplete = () => {
+            /* do nothing */
+        };
+    }
```
interim17, line 23: Similar to `DummyRemoteSimulator` a small testing version of the OctopusClient so we can write tests for the functions that rely on it without having an actual network connection.

src/simularium/OctopusClient.ts
```diff
@@ -21,6 +29,21 @@ export class OctopusServicesClient {
...
+                if (
+                    this.onConversionComplete &&
+                    msg.fileName === this.lastRequestedFile
+                ) {
+                    this.onConversionComplete(msg.fileName);
+                }
+            }
+        );
+    }
+
```
interim17, line 46: This is the critical change that allows the viewer to handle checking when conversion is complete.

src/test/TestClientSimulatorImpl.ts
```diff
@@ -0,0 +1,61 @@
...
+            bundleData: [
+                {
+                    data: [1, 0, 0, 0, 0, 0, 0, 0, 0, 0.1, 0], // One single point at origin
+                    frameNumber: this.frame,
+                    time: this.frame,
+                },
+            ],
+            fileName: "test",
+        };
+    }
```
interim17, line 29: A minimal client simulator for testing.

src/simularium/types.ts
```diff
@@ -209,3 +212,27 @@ export interface CacheNode {
...
     prev: CacheNode | null;
 }
+
+export interface BaseSimulatorParams {
+    fileName: string;
+}
+
+export interface RemoteSimulatorParams extends BaseSimulatorParams {
+    netConnectionSettings: NetConnectionParams;
+    requestJson?: boolean;
```
meganrm, line 222: what does this do?

src/simularium/types.ts
```diff
@@ -209,3 +212,27 @@ export interface CacheNode {
...
+    prefetchFrames?: boolean;
+}
+
+export interface LocalProceduralSimulatorParams extends BaseSimulatorParams {
+    clientSimulatorImpl: IClientSimulatorImpl;
+}
```
meganrm, line 228: I would keep the names consistent with the class names for now, ie this should be `ClientSimulatorParams`. 
I think `ClientSimulator` makes sense as a class name given `RemoteSimulator`, I think the `clientSimulatorImpl` should be renamed. Some suggestions: `simulator` or `generator`

src/simularium/types.ts
```diff
@@ -209,3 +212,27 @@ export interface CacheNode {
...
+
+export interface LocalFileSimulatorParams extends BaseSimulatorParams {
+    simulariumFile: ISimulariumFile;
+    geoAssets?: { [key: string]: string };
+}
```
meganrm, line 233: maybe this could be called something like `StoredFileSimulatorParams` in the future. since the file can actually be hosted somewhere, but the difference between this and the RemoteSimulator is that the data is completely loaded into memory.

src/util.ts
```diff
@@ -133,3 +137,29 @@ export const getNextAgentOffset = (
     const nSubPoints = view[currentOffset + AGENT_OBJECT_KEYS.length - 1];
     return currentOffset + AGENT_OBJECT_KEYS.length + nSubPoints;
 };
+
```
meganrm, line 140: I know this is still a draft, but I would want to see how you're planning on using these

src/controller/index.ts
meganrm: why not call this just `get octopusClient` and make this.octopusClient `this._octopusClient`? I would never guess that `ensureOctopusClient` would just return the client

src/controller/index.ts
```diff
@@ -266,103 +258,116 @@ export default class SimulariumController {
...
 
+    ///// File Changing /////
+
     public clearFile(): void {
         this.isFileChanging = false;
+        this.clearFileResources();
+    }
+
+    private cancelCurrentFile(): void {
+        this.isFileChanging = true;
```
meganrm, line 269: why is this true?

src/simularium/ClientSimulator.ts
```diff
@@ -12,7 +12,7 @@ import { ISimulator } from "./ISimulator.js";
 // a ClientSimulator is a ISimulator that is expected to run purely in procedural javascript in the browser client,
 // with the procedural implementation in a IClientSimulatorImpl
 export class ClientSimulator implements ISimulator {
-    private localSimulator: IClientSimulatorImpl;
+    public localSimulator: IClientSimulatorImpl;
```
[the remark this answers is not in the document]
interim17: Sure, I still prefer `plugin` or `generator` or anything that doesn't overlap with the naming of `ISimulator` but taking local out is fine.

src/simularium/types.ts
interim17: Request json? It's a flag alli and I added for dealing with the BrownianMotion simulator (or any other remote case) that sends json data

src/simularium/types.ts
interim17: Definitely, after we chatted I realized I don't think the test bed can currently handle drop box links etc., I'm working on fixing that.

src/simularium/types.ts
meganrm: as opposed to binary?

src/controller/index.ts
interim17: Pretty sure that the primary purpose of this flag is to alert the animate loop that we are changing files, as that is the place that the flag `markFileChangeAsHandled` gets called. I'd love to understand that architectural choice better, all I've found is that in practice if i call `clearFile` and don't set that flag to true, the data doesn't render.

The `cancelCurrentFile` function was added recently by Alli, my change is just to make both `cancelCurrentFile` and `clearFile` call `clearFileResources`.

Having the two similar but slightly different code paths seemed like a way that our code could start to diverge over time, the only real difference was whether or not they set that flag to true or false, so I consolidated the shared functionality in one place.

src/simularium/types.ts
interim17: Yes, was following Alli's lead on what octopus needed to get the right data out of that simulation

src/simularium/RemoteClient.ts
```diff
@@ -0,0 +1,90 @@
...
+        this.handleError =
+            errorHandler ||
+            (() => {
+                /* do nothing */
+            });
+    }
+
+    public setErrorHandler(handler: (msg: Error) => void): void {
+        this.handleError = handler;
+    }
```
interim17, line 34: This base class manages the websocket for classes that need a remote connection.

src/simularium/Simulator/RemoteSimulator.ts
```diff
@@ -1,51 +1,54 @@
...
 // and plays back a trajectory specified in the NetConnectionParams
-export class RemoteSimulator implements ISimulator {
-    public webSocketClient: WebsocketClient;
-    protected logger: ILogger;
+export class RemoteSimulator extends BaseRemoteClient implements ISimulator {
     public onTrajectoryFileInfoArrive: (NetMessage) => void;
     public onTrajectoryDataArrive: (NetMessage) => void;
-    public lastRequestedFile: string;
-    public handleError: (error: FrontEndError) => void | (() => void);
+    protected logger: ILogger;
```
interim17, line 16: Extending `BaseRemoteClient` while implementing `ISimulator` helps keep things dry, and keeps websocket code out of local simulators.

src/controller/index.ts
```diff
@@ -77,91 +68,77 @@ export default class SimulariumController {
...
-            );
+            message = "An unknown error occurred.";
         }
 
+        if (this.onError) {
+            this.onError(new FrontEndError(message));
+        } else throw new Error(message);
+    }
+
+    private createSimulatorConnection(params: SimulatorParams): void {
```
interim17, line 93: I think this is now concise enough to be one function again (removed `initSimulator`)

@meganrm and @toloudis as we discussed the params are weakly typed so this relies on runtime checks in the constructors of the simulators.

src/controller/index.ts
```diff
@@ -266,105 +193,166 @@ export default class SimulariumController {
...
+    }
+
+    private shouldReuseSimulator(params: SimulatorParams): boolean {
+        return (
+            "netConnectionSettings" in params &&
+            this.simulator instanceof RemoteSimulator &&
+            this.simulator.isConnectedToRemoteServer() &&
+            this.lastNetConnectionConfig === params.netConnectionSettings
+        );
+    }
```
interim17, line 210: As discussed we will reuse remote simulators if the net connection settings have not changed.

src/controller/index.ts
```diff
@@ -373,43 +361,24 @@ export default class SimulariumController {
...
-        netConnectionConfig: NetConnectionParams
-    ): void {
-        if (!this.isRemoteOctopusClientConfigured()) {
-            this.configureNetwork(netConnectionConfig);
-        }
-        if (this.octopusClient) {
-            this.octopusClient.setHealthCheckHandler(handler);
-            this.octopusClient.checkServerHealth();
-        }
-    }
```
interim17, line 387: These changes makes these server health checks an unnecessary burden on the front end. We configure the connection as needed, we can rely on `isConnectedToRemoteServer`.

Issue #423, opened 2024-10-23
Cache warming


interim17:
## Use Case

To satisfy [this UX ticket for a buffering indicator on the website time slider](https://github.com/simularium/simularium-website/issues/338), and potentially to improve performance, and potentially to facilitate playback speed feature, cache warming (pre-streaming of frames prior to playback) is needed.

Open questions:
- What should the default behavior be?
- How much to warm? A ratio of total frames? Whole trajectories? 
- What degree of control should the front end implementation have over this behavior?
- Do we want to make changes to the back end? streaming behaviors defined as "resume" or "pause" seem to presume that we stream only during playback, warming can probably be built without changing these behaviors or terms, but they may be less accurate/precise afterwards
- Control of playback speed is a requested front end feature, but currently playback speed is very much determined by network latency, at least on first playthrough, total/partial warming makes this a more reasonable thing to build?
- Sort out interactions of warming when cache size is limited.

Joe's rough draft idea 10/23:
- leave existing controller methods in place, no breaking changes
- don't change octopus/websocket message types (only update these after the fact if more naming accuracy is desired), use intervals and other checks internal to the viewer to manage warming
- expose props/controller methods to allow front end to initiate warming
- provided a default ratio of frames for warming, allow front end to update that amount down, or up to 100%
- allow warming to be initiated pre-playback/time jumps, and then allow playback to proceed with old behavior, or with ongoing buffering in advance of playback position

## Acceptance Criteria

- viewer can request and cache some some of frames without initiating playback
- decided upon default behavior happens
- decided upon implementation options/controls are exposed and function as expected

## Details

_Please provide any helpful specifications_

PR #469, opened 2025-03-18
add cache logging callback

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
Advances #423 

Solution
========
The logging component from the previous prefetching branch, pulled out to keep work encapsulated. This didn't have any issues previously besides feedback that it logged too much data, I've pared down the log to just current cache size and array of the frame numbers in the cache. Added getters for maxSize and cacheEnabled as those don't change rapidly and don't need to be in the log.

Also unbundled the test data in `VisData.test.ts` to match how our data is sent now.

Screenshot 2025-03-18 at 4 44 30 PM

### DIFF
--- a/src/simularium/VisDataCache.ts
+++ b/src/simularium/VisDataCache.ts
@@ -138,10 +138,7 @@ class VisDataCache {
     }
 
     public getFirstFrameNumber(): number {
-        if (this.head !== null) {
-            return this.head.data.frameNumber;
-        }
-        return -1;
+        return this.head?.data.frameNumber ?? -1;
     }
 
     public getFirstFrameTime(): number {
@@ -153,10 +150,7 @@ class VisDataCache {
     }
 
     public getLastFrameNumber(): number {
-        if (this.tail !== null) {
-            return this.tail.data.frameNumber;
-        }
-        return -1;
+        return this.tail?.data.frameNumber ?? -1;
     }
 
     public getLastFrameTime(): number {
--- a/src/simularium/VisDataCache.ts
+++ b/src/simularium/VisDataCache.ts
@@ -1,9 +1,10 @@
 import { compareTimes } from "../util.js";
-import { CachedFrame, CacheNode } from "./types.js";
+import { CachedFrame, CacheLog, CacheNode } from "./types.js";
 
 interface VisDataCacheSettings {
     maxSize: number;
     cacheEnabled: boolean;
+    onCacheUpdate?: (log: CacheLog) => void;
 }
 
 class VisDataCache {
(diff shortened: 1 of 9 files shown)


src/simularium/types.ts
```diff
@@ -209,3 +209,8 @@ export interface CacheNode {
     next: CacheNode | null;
     prev: CacheNode | null;
 }
+
+export interface CacheLog {
+    size: number;
+    framesInCache: number[];
+}
```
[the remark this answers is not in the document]
interim17: This is the current size of the cache, the max size is accessed by a method. Does that answer the question?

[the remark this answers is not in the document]
ShrimpCryptid: Ah, I meant, should the CacheLog include the maximum size of the cache that is emitting the log?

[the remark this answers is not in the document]
interim17: It used to be included, but Megan asked that I remove everything that wasn't changing on frame by frame updates and directly needed by the front end buffering indicator feature so I moved it to the method. Since this calls after frames get added, but maxSize doesn't change unless you call changeSettings I pulled it.

src/simularium/VisDataCache.ts
```diff
@@ -36,14 +39,38 @@ class VisDataCache {
...
+        this.onCacheUpdate();
+    }
+
+    public onCacheUpdate(): void {
+        if (this.logCacheUpdate) {
+            this.logCacheUpdate({
+                size: this.size,
+                framesInCache: this.getListOfCachedFrameNumbers(),
+            });
+        }
```
frasercl, line 63: I believe TS lets you shrink this to

```suggestion
        this.logCacheUpdate?.({
            size: this.size,
            framesInCache: this.getListOfCachedFrameNumbers(),
        });
```

...if you want to

src/simularium/VisDataCache.ts
```diff
@@ -36,14 +39,38 @@ class VisDataCache {
...
+    }
+
+    private getListOfCachedFrameNumbers(): number[] {
```
frasercl, line 66: nit/optional: might shorten this to `getCachedFrameNumbers`; the plural kind of implies a list already

src/simularium/VisDataCache.ts
```diff
@@ -111,7 +138,10 @@ class VisDataCache {
     }
 
     public getFirstFrameNumber(): number {
-        return this.head?.data.frameNumber || -1;
+        if (this.head !== null) {
+            return this.head.data.frameNumber;
+        }
+        return -1;
```
frasercl, line 144: If the issue here was that `0` was getting ignored, the solution might be the `??` operator?

src/simularium/VisDataCache.ts
```diff
@@ -123,7 +153,10 @@ class VisDataCache {
     }
 
     public getLastFrameNumber(): number {
-        return this.tail?.data.frameNumber || -1;
+        if (this.tail !== null) {
+            return this.tail.data.frameNumber;
+        }
+        return -1;
```
frasercl, line 159: same as above

frasercl (approved):
Looks good! Some nitpicky comments but no blockers.

PR #478, opened 2025-03-21
move plots and metrics API into ISimulator (controller maintenance 2/n)

### LINKED ISSUE #477: move plot data api into `ISimulator`
## Use Case

Discussion of who should be responsible for web sockets, data requests, etc. we arrived at the decision to move requesting plots and metrics into `ISimulator`

This is not fully implemented in our production apps.

## Acceptance Criteria

Current test bed functionality is retained when calling `requestAvailableMetrics` and `requestPlotData` on the simulators

interim17:
Time Estimate or Size
=======
_medium_
no changes to intended functionality, should work as before, code shuffling mostly

Problem
=======
Closes #477 

Solution
========
The remote metrics and plot API is not fully implemented in production, but in the test bed we can log results.

We are not building functionality, but moving the API into the `ISimulator` interface.
Client simulators: not implemented.
Local file simulators: received a remote metrics calculator from the controller to request with.
Remote simulators: handles these message types independently on its own websocket connection.

Controller: does not hold remote metrics calculator instance, build configures one for local simulators when necessary.

Forthcoming work will move web socket connection out of controller entirely, so better that we start here by taking the metrics calculator off the class.


src/controller/index.ts
```diff
@@ -392,59 +391,39 @@ export default class SimulariumController {
         }
     }
 
-    private setupMetricsCalculator(
+    private async setupMetricsCalculator(
```
toloudis, line 394: shouldn't all the implementation of this move inside RemoteSimulator? who calls this method?

src/controller/index.ts
```diff
@@ -392,59 +391,39 @@ export default class SimulariumController {
...
     public async getMetrics(config: NetConnectionParams): Promise<void> {
-        if (
-            !this.metricsCalculator ||
-            !this.metricsCalculator.socketIsValid()
-        ) {
-            this.metricsCalculator = this.setupMetricsCalculator(config);
-            await this.metricsCalculator.connectToRemoteServer();
+        if (this.simulator instanceof LocalFileSimulator) {
+            const calculator = await this.setupMetricsCalculator(config);
+            await this.simulator.setupMetricsCalculator(calculator);
```
toloudis, line 413: is this usage scenario real?  when does this actually happen?  it would be great to eliminate this case if poss

src/simularium/ISimulator.ts
```diff
@@ -22,6 +22,16 @@ export interface ISimulator {
...
 
+    /** todo implement callback pattern for plots and metrics */
+    /** a callback to receive available metrics */
+    // setAvailableMetricsHandler(
+    //     handler: (msg: NetMessage) => void
+    // ): void;
+    // /** a callback to receive plot data */
+    // setPlotDataHandler(
+    //     handler: (msg: NetMessage) => void
+    // ): void;
```
toloudis, line 33: for completeness, would it be good to close the loop and actually add this stuff right now in this pr?

toloudis (commented):
comments added...

src/controller/index.ts
interim17: It is real. It's not used much or fully implemented but I talked to Alli this morning and we can send a local file to octopus and simulariumio will generate metrics for that file.

src/controller/index.ts
interim17: Local file simulators can request metrics from octopus. I have a PR coming that moves this inside of local file simulator since its the only one who needs it.

src/simularium/ISimulator.ts
interim17: Up to you, but it feels out of scope since we don't have a full implementation in place.

In production the website gets its plots a different way, bypasses this API. It seems to be "in development". I'm into the idea of moving it forward but imo it would make more sense to do after the simulator params updates are done, since it will involve updating the simulator creation process and adding props to the viewport.

I'm about to open another PR that kind of finishes what I would call the first round of this work. Alternately, we could make this whole feature fully dormant and just remove it entirely for now, since no production apps use it, but that makes me worry that we won't ever pick it

interim17: https://github.com/simularium/simularium-viewer/pull/481/files

That PR is the branch where I removed the controller `setUpMetricsCalculator` but did finally capitulate and add back one `simulator instanceof LocalFileSimulator` check since it just felt wrong to have to have `setupMetricsCalculator` on `ClientSimulator` or `RemoteSimulator`.

That's my current best thought on this, unless we want to just nix getting metrics from local files.

src/controller/index.ts
toloudis: ok but who does this and why?

interim17: I don't have a good enough sense of what our users actually do to answer that, maybe we can ask Blair at dev meeting. Left a scenario below but its very speculative.

For simplicty's sake I would love to remove it, simplifies things a lot. It would be easier honestly to just let remote simulators take consume a local file and run the service if we want it back in the future.

src/simularium/ISimulator.ts
```diff
@@ -46,4 +56,11 @@ export interface ISimulator {
     requestFrameByTime(time: number): void;
     /** request trajectory metadata */
     requestTrajectoryFileInfo(fileName: string): void;
+    /** request available metrics */
+    requestAvailableMetrics(): void;
+    // /** request available plots */
+    requestPlotData(
+        data: Record<string, unknown>,
+        plots: Array<PlotConfig>
+    ): void;
```
ShrimpCryptid, line 65: These docstrings seem a little simple since they're basically the name of the function... Consider either adding more detail or removing?

src/simularium/LocalFileSimulator.ts
```diff
@@ -133,4 +140,45 @@ export class LocalFileSimulator implements ISimulator {
     public getSimulariumFile(): ISimulariumFile {
         return this.simulariumFile;
     }
+
+    public async setupMetricsCalculator(
```
ShrimpCryptid, line 144: Nit: Consider `setMetricsCalculator`? I know this will eventually do setup steps, but it makes it clear that it's setting a local field in addition to whatever other setup it'll be doing. Also `setup` implies to me that it would be returning the modified `RemoteMetricsCalculator`

ShrimpCryptid (approved):
Left some minor comments!

src/controller/index.ts
toloudis: I guess if this PR is "move plots and metrics API into ISimulator" then this should be gone

src/controller/index.ts
toloudis: yeah I think we need to know if it's actually being used for real.  my instinct is to shove it into the LocalSimulator as a constructor option (pass in connection params in case the localsimulator needs to talk to a metrics server) or else just get rid of it until there's some reason to mix local and remote

interim17:
@toloudis for now after various chats I've just removed the ability to request remote metrics for local files entirely.

it exists on remote simulator, we have `noop` plot and metrics requests on local and client simulators.

code is simpler, less abstractions floating about, and no functionality removed that in use by anyone afawk.

if/when we want to reimplement there are lots of ways to handle it, seems like remote simulator an `ISimulatorimFile` handle everything current users need just fine.

src/controller/index.ts
```diff
@@ -131,6 +134,12 @@ export default class SimulariumController {
                 this.handleTrajectoryInfo(trajFileInfo);
             }
         );
+        this.simulator.setMetricsHandler((metrics: Record<string, unknown>) =>
```
toloudis, line 137: can this be typed better? I saw at least you have a Metrics type.  even if that type is a generic Record<string,unknown> it would still be better to use `Metrics` here and above on line 65

src/controller/index.ts
```diff
@@ -392,59 +401,16 @@ export default class SimulariumController {
...
-        if (
-            !this.metricsCalculator ||
-            !this.metricsCalculator.socketIsValid()
-        ) {
-            this.metricsCalculator = this.setupMetricsCalculator(config);
-            await this.metricsCalculator.connectToRemoteServer();
-        }
-        this.metricsCalculator.getAvailableMetrics();
+    // todo handle config in subsequent work on "last known net settings"
+    public async getMetrics(_config: NetConnectionParams): Promise<void> {
```
toloudis, line 405: remove the config argument if it's not used

src/controller/index.ts
```diff
@@ -392,59 +401,16 @@ export default class SimulariumController {
...
-            );
-        } else if (this.simulator instanceof RemoteSimulator) {
-            // we don't have the simularium file, so we'll just send an empty data object
-            this.metricsCalculator.getPlotData(
-                {},
-                requestedPlots,
-                this.simulator.getLastRequestedFile()
-            );
-        }
+        this.simulator?.requestPlotData({}, requestedPlots);
```
toloudis, line 413: remove the config argument

src/controller/index.ts
```diff
@@ -460,6 +426,20 @@ export default class SimulariumController {
         }
     }
 
+    public set onMetricsCallback(callback: (msg: unknown) => void) {
+        this.handleMetrics = callback;
+        if (this.simulator) {
+            this.simulator.setMetricsHandler(callback);
+        }
```
toloudis, line 433: why do you store the callback both here and in the simulator?

src/simularium/ClientSimulator.ts
```diff
@@ -48,6 +61,14 @@ export class ClientSimulator implements ISimulator {
     ): void {
         this.onTrajectoryDataArrive = handler;
     }
+    public setMetricsHandler(
+        handler: (msg: Record<string, unknown>) => void
```
toloudis, line 65: use a Metrics type (even if it is an alias for Record<string, unknown>

src/simularium/ISimulator.ts
```diff
@@ -22,6 +27,13 @@ export interface ISimulator {
         handler: (msg: VisDataMessage | ArrayBuffer) => void
     ): void;
 
+    /** todo implement callback pattern for plots and metrics */
+    /** a callback to receive available metrics */
+    setMetricsHandler(handler: (msg: Record<string, unknown>) => void): void;
```
toloudis, line 32: type

src/simularium/ISimulator.ts
```diff
@@ -46,4 +58,11 @@ export interface ISimulator {
     requestFrameByTime(time: number): void;
     /** request trajectory metadata */
     requestTrajectoryFileInfo(fileName: string): void;
+    /** request available metrics */
+    requestAvailableMetrics(): void;
+    /** request available plots */
+    requestPlotData(
+        data: Record<string, unknown>,
+        plots: Array<PlotConfig>
```
toloudis, line 66: can you describe what these args are?  it's hard to tell what `data` means in this context, better typing and/or naming would make it a bit clearer I hope

toloudis (commented):
this is looking so much cleaner.  I just have a few more cleanup comments but the structure looks good

src/simularium/ISimulator.ts
```diff
@@ -22,6 +28,13 @@ export interface ISimulator {
         handler: (msg: VisDataMessage | ArrayBuffer) => void
     ): void;
 
+    /** todo implement callback pattern for plots and metrics */
+    /** a callback to receive available metrics */
+    setMetricsHandler(handler: (msg: Metrics) => void): void;
+
+    // /** a callback to receive plot data */
```
toloudis, line 35:
```suggestion
    /** a callback to receive plot data */
```

toloudis (approved):
2 tiny cleanup suggestions

PR #479, opened 2025-03-21
remove server health checks (controller maintenance 3/n)

interim17:
Time Estimate or Size
=======
_tiny_

Problem
=======
Server health checks were instituted for auto-conversion in anticipation of a more complex Octopus architecture and high loads on the server.

For now they are redundant and socket validity checks are sufficient for confirming that auto-conversion service is available

Solution
========
Health check message types remain and functionality can be added back in the future if the server behavior seems to make it necessary.


meganrm (approved):

toloudis (approved):
Changes LGTM although I am cautious about why we should be removing something called a "health check"!

PR #481, opened 2025-03-21
move WebsocketClient into remote abstractions // onConversionComplete handler (controller maintenance 4/n)

### LINKED ISSUE #472: handle receiving converted file for the front end
## What needs to happen?

Currently the front end needs to track the outgoing conversion file name and then call changeFile with a particular flag when the metadata lands, we can intercept the fileName in the viewer and simplify the process for the caller.

## Why should we do this?

Make the tool easier to use. Will scale to other back end services and API integrations.

### LINKED ISSUE #480: move web socket connection down into the remote connected classes and off the controller
## Use Case

Remote connected classes should have their own web socket connection, web socket connection should not live on the controller anymore.

## Acceptance Criteria

_Please describe how you know this is done_

## Details

interim17:
Time Estimate or Size
=======
_medium large_

Problem
=======
Closes #480
Closes #472 

Solution
========

Sorry for the code churn on this branch. Wanted to go with an inheritance pattern on web socket client for conciseness and @frasercl talked me out of it. Favoring composition over inheritance here.

### Inheritance of web socket client 
Remote simulator and conversion compose/own their own instance of websocket client.

### Controller
No longer holds web socket client on behalf of service clients.

### `ConversionClient`
Has a callback to watch for conversion completion. This removes the need for the front end to track and manage receiving converted files, or for change/init file request to track whether we need to keep our current remote connection.

###


examples/src/Viewer.tsx
```diff
@@ -460,39 +445,17 @@ class Viewer extends React.Component<InputParams, ViewerState> {
...
-                netConnectionSettings: this.netConnectionSettings,
-            })
-            .then(() => {
-                simulariumController.gotoTime(0);
-            })
-            .catch((e) => {
-                console.warn(e);
-            });
-    }
-
```
interim17, line 479: Viewer handles this internally, no need to watch for flag.

src/controller/index.ts
```diff
@@ -190,67 +168,94 @@ export default class SimulariumController {
...
+    private async configureConversionClient(
+        config: NetConnectionParams
+    ): Promise<void> {
+        this._conversionClient = new ConversionClient(
+            config,
+            this.playBackFile,
+            this.onError
+        );
+
+        await this.conversionClient.initialize(this.playBackFile);
```
meganrm, line 189: this also seems like you're sending the file in twice

src/simularium/OctopusClient.ts
```diff
@@ -1,23 +1,74 @@
...
+    }
+
+    public async initialize(fileName: string): Promise<void> {
+        this.lastRequestedFile = fileName;
+        try {
+            await this.webSocketClient.connectToRemoteServer();
+        } catch (error) {
+            this.handleError(error as FrontEndError);
+        }
     }
```
meganrm, line 47: do you really need a separate `initialize` here? If the Conversion client only exists to convert a file on command, and you're already sending in the file to the construction why not just connect in the constructor as well?

meganrm, line 47: do you really need a separate `initialize` here? If the Conversion client only exists to convert a file on command, and you're already sending in the file to the construction why not just connect in the constructor as well?

src/simularium/OctopusClient.ts
meganrm, line 1: rename to ConversionClient

src/simularium/RemoteSimulator.ts
```diff
@@ -30,12 +34,16 @@ export class RemoteSimulator implements ISimulator {
...
         errorHandler?: (error: FrontEndError) => void,
         jsonResponse = false
     ) {
-        this.webSocketClient = webSocketClient;
-        this.lastRequestedFile = "";
+        this.webSocketClient = new WebsocketClient(
+            netConnectionSettings,
+            errorHandler
+        );
+        this.lastRequestedFile = fileName;
```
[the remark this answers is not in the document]
interim17: We discussed always having the file name in the params so it seems odd to pass it in but then just ignore it.

If the class instance can be reused by calling initialize a second time we need to take in the new file name, which is what we wanted for remote simulator.

But this seems true about the about conversion client is true since that is a single use instance, we can just set it when we call the conversion methods!

src/controller/index.ts
```diff
@@ -182,6 +159,7 @@ export default class SimulariumController {
         if (this.simulator) {
             this.simulator.abort();
         }
+        this.simulator = undefined;
```
meganrm, line 162: this seems like it will prevent us from keeping the RemoteSimulator live and just switching the file name?

src/controller/index.ts
interim17: Will remove and just pass in when we call the methods

src/simularium/RemoteSimulator.ts
```diff
@@ -82,8 +88,36 @@ export class RemoteSimulator implements ISimulator {
...
+     */
+    public async initialize(fileName: string): Promise<void> {
+        this.lastRequestedFile = fileName;
+        try {
+            await this.webSocketClient.connectToRemoteServer();
+            this.registerBinaryMessageHandlers();
+            this.registerJsonMessageHandlers();
+        } catch (error) {
+            this.handleError(error as FrontEndError);
+        }
```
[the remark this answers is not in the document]
interim17: If you are already connected the web socket client will return true on this call, but will try to connect if disconnected, so it's a fallback in case the connect was interrupted. We can't properly await the connection process in the constructor.

src/simularium/OctopusClient.ts
interim17: You can't use `await` in the constructor.

src/simularium/OctopusClient.ts
```diff
@@ -1,23 +1,74 @@
...
+            }
+        );
     }
 
     public async convertTrajectory(
         dataToConvert: Record<string, unknown>,
         fileType: TrajectoryType,
         fileName: string
     ): Promise<void> {
+        await this.webSocketClient.connectToRemoteServer();
```
meganrm, line 71: this seems to be doing the same thing as initialize

src/controller/index.ts
interim17: I was planning to handle keeping remote simulator instances in the next PR where we do simulator params and updates to `createSimulatorConnection`.

src/simularium/OctopusClient.ts
interim17: Good point, I'll remove initialize here and also `configureConversionClient`, that will make this more concise, i had initialize in when this class was extending web socket client, but you're right its not needed anymore.

src/controller/index.ts
```diff
@@ -190,32 +168,69 @@ export default class SimulariumController {
...
+        );
+
+        this.conversionClient.setOnConversionCompleteHandler(() => {
+            this.changeFile(
+                {
+                    netConnectionSettings: netConnectionConfig,
+                },
+                this.playBackFile
+            );
+            this.closeConversionConnection();
```
toloudis, line 203: Nit: do you maybe want to do this first? does the order matter here?  (close old connection before opening new one?)

interim17:
@toloudis In regards to comment above, I would be down to just do everything in one PR! It will be a big change set but if y'all are game to review it I would love to get this all done in one coordinated move...

src/controller/index.ts
```diff
@@ -190,32 +168,69 @@ export default class SimulariumController {
...
+        fileName: string
+    ): Promise<void> {
+        this.cancelCurrentFile(fileName);
+
+        this._conversionClient = new ConversionClient(
+            netConnectionConfig,
+            this.onError
+        );
+
+        this.conversionClient.setOnConversionCompleteHandler(() => {
```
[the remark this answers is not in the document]
interim17: Could do. I wanted it separate for the future case where instead of running start we might pass in something else, like download, for a case we want to convert something and immediately output a trajectory file instead of playback?

src/controller/index.ts
interim17: The idea is that we always close on completion of conversion, so there shouldn't be a connection open when we are starting? But we could check just in case first too...

src/controller/index.ts
```diff
@@ -316,36 +322,15 @@ export default class SimulariumController {
     }
 
     public initNewFile(
```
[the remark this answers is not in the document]
interim17: I consolidated this inside `changeFile` without otherwise expanding the scope of the PR for now.

src/simularium/RemoteSimulator.ts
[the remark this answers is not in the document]
interim17: removing and setting file name only in initialize

examples/src/Viewer.tsx
toloudis: does this imply a necessary change in simularium-website to keep up?

src/controller/index.ts
```diff
@@ -40,8 +39,7 @@ interface SimulatorConnectionParams {
 
 export default class SimulariumController {
     public simulator?: ISimulator;
-    public remoteWebsocketClient?: WebsocketClient;
-    public octopusClient?: OctopusServicesClient;
+    public _conversionClient?: ConversionClient;
```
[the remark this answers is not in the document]
interim17: Whoops yeah supposed to be private!

interim17:
@toloudis yes and no. the default behavior then and now was more like "convert and load" rather than just "convert"

the front end could still handle its own logic by checking the incoming metadata but it's not designed for that anymore, this is definitely written to make the call of `start`automatic and internal.

the new callback makes this easily configurable if we want to adjust behavior going forward, something like the snippet below, or just new fixed methods like `convertAndDownload` `convert` `convertAndLoad`

```
   interface ConversionParams {
        netConnectionSettings: NetConnectionParams;
        fileName: string;
        loadOnComplete: boolean;
        onComplete?: (arg: unknown) => void;
    }


    private async setupConversion(
        params: ConversionParams
    ): Promise<void> {
        const { netConnectionSettings, fileName, loadOnComplete, onComplete } = params;
        this.cancelCurrentFile(fileName);

        this._conversionClient = new ConversionClient(
            netConnectionSettings,
            this.onError
        );

        this.conversionClient.setOnConversionCompleteHandler(() => {
            if (onComplete) {
                onComplete(fileName);
            }
            if (loadOnComplete) {
                this.changeFile(
                    {
                        netConnectionSettings,
                    },
                    fileName
                );
            }
            this.closeConversionConnection();
        });
    }

meganrm (approved):

PR #489, opened 2025-04-21
Feature/new build simulators

interim17:
Time Estimate or Size
=======
_medium-large_

Problem
=======
Final round of controller refactoring of controller and simulator params.

@meganrm  @toloudis  The approach to most of these changes was discussed in detail at our meeting a few weeks back. I officially futzed with this for too long and then went on vacation. Simularium is paused, would be good to get this done but I know it's not high priority.

More than half of the lines of changes are new tests or moved utils, so the line count isn't as bad as it seems.

After this I will finalize prefetching and controller speed with these new changes and that will get me close to putting simularium work to bed for now.

The utilities that are shared by the `BinaryFIleReader` and `SimulariumController` tests I moved to a util file.

Solution
========

We now have typed interfaces for the argument passed to `changeFile`, there is no more code path to initialize the controller with params. I've updated the readme to show that `changeFile` needs to be called imperatively or in a lifecycle method in order to load a trajectory.

I added quite a bit of testing to the controller, where previously there wasn't much. There could be more, but the goal was to add tests relevant to this change set, and establish a base of coverage so that in the future when we add features, adding tests doesn't require as big of alift.

## Type of change


src/controller/index.ts
```diff
@@ -264,76 +254,58 @@ export default class SimulariumController {
...
-        this.playBackFile = newFileName;
-
-        // calls simulator.abort()
-        this.stop();
-
-        this.visData.WaitForFrame(0);
-        this.visData.clearForNewTrajectory();
+    public clearFile(): void {
+        this.isFileChanging = false;
+        this.clearFileResources();
```
interim17, line 275: `clearFile` as a part of the public API is not functionally changed, but the business logic has been moved into `clearFileResources` to allow some internal flexibility when we want to switch between remote files and reuse the simulator.

src/controller/index.ts
```diff
@@ -79,82 +69,78 @@ export default class SimulariumController {
...
-                this.visData.parseAgentsFromNetData.bind(this.visData)
-            );
+    }
+
+    /**
+     * @param error a string or an Error object, which can be of type
+     * "unknown" if passed by a catch block
+     */
+    private handleError(error: unknown | string): void {
+        let message: string;
```
interim17, line 79: Small update to `handleError` in the controller to make sure we are capturing all possible errors we send to it (particularly try/catch blocks were not getting handled correctly in some cases before.

src/Simulator/ISimulator.ts
```diff
@@ -4,8 +4,42 @@ import {
...
+    } else if ("simulariumFile" in params) {
+        return {
+            simulatorClass: LocalFileSimulator,
+            typedParams: params as LocalFileSimulatorParams,
+        };
+    } else {
+        return { simulatorClass: null, typedParams: null };
+    }
+};
+
```
interim17, line 42: `getSimulatorClassFromParams` handles generating the right type of simulator for the incoming params, the simulators have runtime checks to make sure they have received the right params as this can be run in JS on the client side. @meganrm outlined this function in comments on previous PR

src/test/SimulariumController.test.ts
```diff
@@ -1,12 +1,407 @@
...
+const mockConversionClient = {
+    setOnConversionCompleteHandler: vi.fn().mockImplementation(() => {
+        console.log("mocked handler");
+    }),
+    cancelConversion: vi.fn(),
+    disconnect: vi.fn(),
+    convertTrajectory: vi.fn().mockResolvedValue(undefined),
+    sendSmoldynData: vi.fn().mockResolvedValue(undefined),
+    lastRequestedFile: "",
+};
```
interim17, line 74: This doesn't mock the internal functionality of the conversion client or the messages we expect to receive, but rather the presence of the functions and if they get called at the right time. Internal testing out of scope imo.

src/test/SimulariumController.test.ts
```diff
@@ -1,12 +1,407 @@
...
+                return {
+                    simulatorClass: LocalFileSimulator,
+                    typedParams: params as LocalFileSimulatorParams,
+                };
+            } else {
+                return { simulatorClass: null, typedParams: null };
+            }
+        },
+    };
+});
```
interim17, line 110: Previously we got a dummy remote simulator but using controller params.

This mock is unchanged from the original except that it provides a dummy instead of a remote simulator.

This is admittedly a maintenance burden as we need to make sure that if we make changes to the real function they are reflected here in the mock overwrite. I'm open to a better approach, but happy that this puts the burden of generating the dummy simulator on the testing file, and out of the production code where we maintaining a code path solely for the purpose of testing.

src/test/SimulariumController.test.ts
```diff
@@ -1,12 +1,393 @@
...
+        ...actual,
+        getSimulatorClassFromParams: (params?: SimulatorParams) => {
+            const result = actual.getSimulatorClassFromParams(params);
+            if (result.simulatorClass === RemoteSimulator) {
+                return { ...result, simulatorClass: DummyRemoteSimulator };
+            }
+            return result;
+        },
+    };
+});
```
interim17, line 96: This may seem a bit convoluted and I'm open to suggestions, but what we are achieving here is getting the dummy remote simulator in as the return value of `getSimulatorClassFromParams` but not mocking the whole function, so that if the real function ever changes, the tests reflect that.

A reason to do this is that by removing the old controller params, we lost the code path to let the client provide the simulator they want to use, and so we need a way to get the dummy in play for testing. This is bit of a hack, but keeps the mess here in the testing and out of the production code.

src/Simulator/ISimulator.ts
```diff
@@ -4,8 +4,42 @@ import {
...
+        };
+    } else if ("simulariumFile" in params) {
+        return {
+            simulatorClass: LocalFileSimulator,
+            typedParams: params as LocalFileSimulatorParams,
+        };
+    } else {
+        return { simulatorClass: null, typedParams: null };
+    }
+};
```
toloudis, line 41: this seems to be the wrong place to put this code.  ISimulator should not depend on the files that implement ISimulator.  

Also do you think this function could probably be more like a factory method that actually returns a ISimulator instance (call `new ClientSimulator`, or `new RemoteSimulator` etc, inside the if statements)?

src/controller/index.ts
```diff
@@ -13,32 +12,22 @@ import {
...
     PlotConfig,
-    Metrics,
 } from "../simularium/types.js";
 
-import { ClientSimulator } from "../simularium/ClientSimulator.js";
-import { IClientSimulatorImpl } from "../simularium/localSimulators/IClientSimulatorImpl.js";
-import { ISimulator } from "../simularium/ISimulator.js";
-import { LocalFileSimulator } from "../simularium/LocalFileSimulator.js";
+import {
+    getSimulatorClassFromParams,
```
toloudis, line 18: prob could just define getSimulatorClassFromParams in here, or make a new ts module called SimulatorFactory.ts

src/controller/index.ts
```diff
@@ -264,76 +254,58 @@ export default class SimulariumController {
...
-            connectionParams.clientSimulator,
-            connectionParams.simulariumFile,
-            connectionParams.geoAssets,
-            connectionParams.requestJson
+    private shouldReuseSimulator(params: SimulatorParams): boolean {
+        return (
+            "netConnectionSettings" in params &&
+            this.simulator instanceof RemoteSimulator &&
+            this.simulator.isConnectedToRemoteServer() &&
+            this.lastNetConnectionConfig === params.netConnectionSettings
```
toloudis, line 283: I like this function, but does this need to be a deeper equality check to compare the netConnectionSettings?

src/Simulator/ISimulator.ts
interim17: @meganrm want to weigh in on this? I left this where and as you wrote it for the most part.

src/controller/index.ts
interim17: Changed to `==`

toloudis: that (changing from `===` to `==`) would be shallower...  deeper check would be to compare the equality of the actual properties of the object.

interim17: Ha, I misunderstood your comment

src/Simulator/types.ts
```diff
@@ -0,0 +1,26 @@
...
+    fileName: string;
+}
+
+export interface RemoteSimulatorParams extends BaseSimulatorParams {
+    netConnectionSettings?: NetConnectionParams;
+    requestJson?: boolean;
+    prefetchFrames?: boolean;
+}
+
+export interface ClientSimulatorParams extends BaseSimulatorParams {
```
toloudis, line 14: just curious, why do these all extend BaseSimulatorParams?  it feels like BaseSimulatorParams is never going to grow beyond fileName and ClientSimulatorParams doesn't even have a fileName.. 
Is BaseSimulatorParams used anywhere outside of this file?

src/controller/index.ts
interim17: Pushed a more robust check

examples/src/Viewer.tsx
```diff
@@ -369,17 +366,17 @@ class Viewer extends React.Component<InputParams, ViewerState> {
             });
     }
 
-    public clearPendingFile() {
-        this.setState({ filePending: null });
-    }
-
```
toloudis, line 375: quick sanity check: this looks like something important, where the code wanted to know that there was a file in-flight?  is it really ok to remove it and within scope of this PR?  or is this dead code that was never used?

toloudis (approved):
I think this is looking good.  Only concern now, is there more work after merging this to update simularium-website? 😢 And mucho testing with existing production simulations from the server!

meganrm:
I have read through this and looks good, but I still want to test it with the education app. 

Could be of interest: One thing I've noticed using the main branch is the "wait for frame" is never really used because the cache is cleared so often the waited for frame gets reset before the frames arrive

src/controller/index.ts
```diff
@@ -264,76 +252,73 @@ export default class SimulariumController {
...
 
-    public cancelCurrentFile(newFileName: string): void {
+    public async changeFile(params: SimulatorParams): Promise<FileReturn> {
+        if (
+            "simulariumFile" in params &&
+            !params.fileName.includes(".simularium")
+        ) {
+            this.handleError("File must be a .simularium file");
+            return Promise.reject({ status: FILE_STATUS_FAIL });
+        }
```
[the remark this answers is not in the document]
meganrm: I removed these lines and the file loaded. For context, this is how I call changeFile: 
```
 const response = await fetch(url);
        if (response.ok) {
            const blob = await response.blob();
            const simulariumFile = await loadSimulariumFile(blob);
            await simulariumController.changeFile({
                fileName: "example",
                simulariumFile: simulariumFile,
            });
            const plotData = simulariumFile.getPlotData();
            setTrajectoryPlotData(plotData[0].data as ScatterTrace[]); // we're not using histograms
        } else {
            throw new Error(`Failed to fetch - ${response.status}`);
        }
```

meganrm:
if you haven't already, I think you should make a branch on the website repo that installs this version and make sure every way of loading a file works: url, local, drag and drop local, examples etc

toloudis:
> if you haven't already, I think you should make a branch on the website repo that installs this version and make sure every way of loading a file works: url, local, drag and drop local, examples etc

YES.  test all the things.

PR #496, opened 2025-06-10
add export field to package.json

interim17:
Time Estimate or Size
=======
_small_

Problem
=======
Installing current viewer version is breaking the website build

It seems the way we are exporting our stylesheet from the viewer, the website bundler can't grab it correctly.
https://webpack.js.org/guides/package-exports/

I was able to inspect `/es` in the node_modules of the website and see the file we were trying to import, but the bundled website was not including it without the explicit export in package.json of the viewer.

Solution
========
Adding "exports" field to package.json tells webpack to bundle the stylesheet during website build.

Locally I can test the fix both a dev build via a directory and a `.tgz`, but I get a little dizzy making changes to both repos, rebuilding, repacking, and reinstalling so I would appreciate a double check, we can also try by just patching the published version.

The unfortunate thing, which I also don't understand, is that this seemed to trigger some breakages in the typing on our styled components in the website.

I have no idea why, it does make sense because we had a mismatch in our v6 `styled-components` and `@types/styled-components` version, but I don't know why the viewer build seemed to trigger it.

Either way, I have a draft PR here for the website that fixes the typing on the styled components.
When we publish the viewer patch, I will install it on that branch, and then hopefully we will see a successful website build.

Final confirmation will be a successful website build with the ability to alt-ctrl-1 to see the advanced controls.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,10 @@
         "registry": "https://registry.npmjs.org/",
         "access": "public"
     },
+    "exports": {
+        ".": "./es/index.js",
+        "./style/style.css": "./style/style.css"
+    },
     "repository": {
         "type": "git",
         "url": "https://github.com/simularium/simularium-viewer.git"


ascibisz (approved):
this looks good to me! thanks for the quick turnaround on the fix

package.json
```diff
@@ -37,6 +37,10 @@
         "registry": "https://registry.npmjs.org/",
         "access": "public"
     },
+    "exports": {
+        ".": "./es/index.js",
+        "./style/style.css": "./style/style.css"
```
toloudis, line 42: this is odd to me because of the `"files"` section above, that has both `"es"` and `"style"` in it

interim17: Oh you're right, so this branch is currently putting it at the top level and in /es. The /es one has always been there but webpack wasn't using it. It works...

But I found another way to fix it which is to just import `import "./style/style.css";` in `src/index.ts`...

Feels too simple, so I'm going to keep rebuilding and repacking and re installing until I'm sure but it seems like the better way to do this.

interim17:
Importing the stylesheet into index.ts: locally the builds, installs and tests are working, but CI tests fail to resolve import.

It's notable to me that Vite building the test bed resolves the import just fine where webpack for the website doesn't. I'm down to look at the simularium website moving to Vite, or otherwise tweaking that build, but I won't be able to get that done today.

Current patch option that makes simularium website build happy is the "exports" field with styles at both levels. or leave this open for a minute

interim17:
@meganrm @toloudis 

Here is my best-thought/understanding about this issue and how to resolve it.

Issue 1: When we switched to ES modules, https://github.com/simularium/simularium-viewer/pull/431, we broke imports that were not in `es/index.js`, because we were not including them specifically in "exports" field. Including the files is not enough to tell bundlers that those imports are allowed when using ES modules.

Issue 2: We thought we fixed that by putting stylesheet in `/es` during `processCss` script https://github.com/simularium/simularium-viewer/pull/456
That only fixed the test bed because of local alias of of `/es` in vite config, and resulted in duplicating the stylesheet in production build when we released 3.10.0. Testing local builds for website seemed to work because we structured the output directory differently than in CI/CD.

My proposed fix in commit [ab61007](https://github.com/simularium/simularium-viewer/pull/496/commits/ab6100712d7405db5bf1f9ffda6004cdfed23aa6):
- Remove top level style directory from "files" in `package.json` and from `copy-build` script
- Put the stylesheet into `es/style/style.css` during `build` and then copy the whole directory when running `copy-build` (used in publishing/CI/CD)
- Alias in exports field so that imports are backwards compatible with docs and previous published packages:
```
    "exports": {
        ".": "./es/index.js",
        "./style/style.css": "./es/style/style.css"
    },
```
allows this import to resolve: import `@aics/simularium-viewer/style/style.css`

Pros: There is no duplication of stylesheet, uses exports field which should have been part of our ES migration, installing local build folder resolves the same way published packages do.
I have tested it with test-bed installing `file:../` aka local build, as well as with test-bed and website installing a packed `.tgz`
Cons: Someone inspecting their installed `node_modules` might be confused to see that they can import `@aics/simularium-viewer/style/style.css` but not see that directory at that level, and they would not understand why that resolves unless they looked in more detail at `package.json` exports field.

Final proof will be to release a patch and install it in the website to make sure that CI/CD pipeline is also happy.

package.json
```diff
@@ -41,6 +40,10 @@
         "type": "git",
         "url": "https://github.com/simularium/simularium-viewer.git"
     },
+    "exports": {
+        ".": "./es/index.js",
+        "./style/style.css": "./es/style/style.css"
```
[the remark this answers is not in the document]
toloudis: okay, then in that case I think I like your fix as is, assuming it works in all the cases we care about! Or if @meganrm has any better ideas

toloudis (approved):
I'd love an elegant thing that doesn't really require special export aliasing config, but I'm happy this makes things work

interim17:
> I'd love an elegant thing that doesn't really require special export aliasing config, but I'm happy this makes things work

Just in time for you to approve this version, I may have found a more elegant thing.

So the reason we couldn't just import style/style.css in `index.ts` was because it was being created by the build step and style was in gitignore. The stylesheet was never actually named `src/style/style.css` it was `src/viewport/index.css`

But `viewport.tsx` and `index.ts' never import it, so it was only making it to the build via the `processCss` step. That made imports/tests fail if you tried to import it locally.

Is there a reason for doing it that way that I'm missing? Like, why we want index.css nested in viewport and not in file path where we expect users to import it?

If we just move the stylesheet up to `src/style/style.css` then tests pass (because it's there), we tell the linter to ignore css so it's happy, and it makes it into the build without the need for an "exports" field!

branch i was testing on: https://github.com/simularium/simularium-viewer/pull/499

